PR #1, opened 2017-11-08
Set up lint and test

SimenB:
Not sure what we want here, so I just threw together something that works.

- Do we want flow?
- Do we want babel?

Other:
- Do we want pre-commit hooks for formatting?
- Do we want Travis or Circle?
- Do we want to support node 4?


orta:
As it's not an externally used lib flow isn't too important to me, and without flow then it doesn't need babel 👍 

> Do we want pre-commit hooks for formatting?

pre-commit with Prettier is a big 👍 for me

> Travis vs Circle

Travis is used on some of the others?

> Node 4 

I don't have strong opinions - it's still actively supported by node though

SimenB:
Ok, made those changes.

Tests are failing on node 4 with a really useless error

.travis.yml
```diff
@@ -0,0 +1,11 @@
...
+node_js:
+  - node
+  - 8
+  - 6
+  - 4
+cache:
+  yarn: true
+branches:
+   only:
+   - master
```
thymikee: geez, I love Travis config being so simple compared to Circle 2

package.json
```diff
@@ -2,23 +2,41 @@
...
+    "eslint-config-prettier": "^2.7.0",
+    "eslint-plugin-prettier": "^2.3.1",
+    "husky": "^0.14.3",
+    "jest": "^21.2.1",
+    "lint-staged": "^4.3.0",
+    "prettier": "^1.8.1"
+  },
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "all"
```
thymikee: add `"bracketSpacing": false` maybe?

SimenB: do you prefer it? Seems like a fb-ism. Happy to add, of course, I don't really care as long as it's formatted for me :D

thymikee: I have no preference, just that the diff would be smaller :D. We can scratch that

SimenB:
#yolo

PR #6, opened 2017-11-09
Fix to-have-length when there are no argument properties

SimenB:
@Ionut-Milas

### DIFF
--- a/rules/prefer_to_have_length.js
+++ b/rules/prefer_to_have_length.js
@@ -17,7 +17,11 @@ module.exports = context => {
         const argumentObject = node.arguments[0].object;
         const argumentProperty = node.arguments[0].property;
 
-        if (propertyName === 'toBe' && argumentProperty.name === 'length') {
+        if (
+          argumentProperty &&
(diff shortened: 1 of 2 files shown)


rules/prefer_to_have_length.js
```diff
@@ -17,7 +17,11 @@ module.exports = context => {
         const argumentObject = node.arguments[0].object;
         const argumentProperty = node.arguments[0].property;
 
-        if (propertyName === 'toBe' && argumentProperty.name === 'length') {
+        if (
+          argumentProperty &&
+          propertyName === 'toBe' &&
```
thymikee: just a nit, can we switch these two lines?

SimenB: what do you mean?

SimenB: Oh, gotcha

thymikee: Sorry for being indirect :D

SimenB:
The rule now runs successfully on the jest codebase as well

PR #8, opened 2017-11-10
Add prefer-to-be-defined rule

xfumihiro:


docs/rules/prefer-to-be-defined.md
```diff
@@ -0,0 +1,28 @@
+# Suggest using  `toBeDefined()` (prefer-to-be-defined)
+
+In order to have a better failure message, `toBeDefined()` should be used upon asserting expections on defined value.
+
+## Rule details
+
+This rule triggers a warning if `toBe()` is used to assert a undefined value.
```
SimenB: `not.toBe()`

docs/rules/prefer-to-be-defined.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+This rule is enabled by default.
+
+### Default configuration
+
+The following patterns are considered warning:
+
+```js
+expect(true).not.toBe(undefined);
```
SimenB: I don't think this should warn, that should be caught by the other rule.

As eslint fixes are multipass, you'd get the same result in the end when autofixing, and then you don't have 2 rules warning about the same thing at once 🙂

rules/prefer_to_be_defined.js
```diff
@@ -0,0 +1,53 @@
...
+        node.parent.type === 'MemberExpression' &&
+        node.parent.parent &&
+        node.parent.parent.type === 'MemberExpression'
+      ) {
+        const parentProperty = node.parent.property;
+        const propertyName = parentProperty.name;
+        const parentProperty2 = node.parent.parent.property;
+        const propertyName2 = parentProperty2.name;
+        const argument = node.parent.parent.parent.arguments[0];
+        const propertyDot = context
```
SimenB: put this inside of the `fixer` so we don't do it every time

rules/prefer_to_be_defined.js
```diff
@@ -0,0 +1,53 @@
...
+          .getSourceCode()
+          .getFirstTokenBetween(
+            parentProperty,
+            parentProperty2,
+            token => token.value === '.'
+          );
+        if (
+          (propertyName === 'not' &&
+            propertyName2 === 'toBe' &&
```
SimenB: and `toEqual`

rules/__tests__/prefer_to_be_defined.test.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const RuleTester = require('eslint').RuleTester;
+const { rules } = require('../../');
```
SimenB: Can't use destructuring

rules/__tests__/prefer_to_be_defined.test.js
```diff
@@ -0,0 +1,46 @@
...
+const rules = require('../../').rules;
+
+const ruleTester = new RuleTester();
+
+ruleTester.run('prefer_to_be_defined', rules['prefer-to-be-defined'], {
+  valid: ['expect(true).toBeDefined();'],
+
+  invalid: [
+    {
+      code: 'expect(true).not.toBe(undefined);',
```
SimenB: See https://github.com/jest-community/eslint-plugin-jest/pull/8#discussion_r150161191, I don't think this rule should warn on `toBeUndefined()`, the other rule can warn on that

SimenB:
@xfumihiro This needs a rebase.

Can you also update the changelog? 🙂

rules/__tests__/prefer_to_be_defined.test.js
```diff
@@ -0,0 +1,35 @@
...
+const rules = require('../../').rules;
+
+const ruleTester = new RuleTester();
+
+ruleTester.run('prefer_to_be_defined', rules['prefer-to-be-defined'], {
+  valid: ['expect(true).toBeDefined();'],
+
+  invalid: [
+    {
+      code: 'expect(true).not.toEqual(undefined);',
```
SimenB: Should have a test for `toBe` as well

rules/__tests__/prefer_to_be_defined.test.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const RuleTester = require('eslint').RuleTester;
+const rules = require('../../').rules;
+
+const ruleTester = new RuleTester();
+
+ruleTester.run('prefer_to_be_defined', rules['prefer-to-be-defined'], {
+  valid: ['expect(true).toBeDefined();'],
```
SimenB: Can you add some more examples here? Mostly to be sure we avoid stuff like #6 and #20

SimenB:
I like the direction! Definitely something we can build upon.

It's currently failing on Node 4, as that does not support destructuring.

xfumihiro:
Isn't Node 4 support [being dropped](https://github.com/facebook/jest/pull/4769)?

thymikee:
Yep, in the next release.

SimenB:
It's being dropped for jest core in the next release, not necessarily here 🙂 See #1 where it was discussed

SimenB:
We could add babel, but it is nice to not have to do that

xfumihiro:
I agree it makes more sense that this should be merged with **prefer-to-be-undefined** rule, but what name should it have?
prefer-to-be-defined-or-undefined?
prefer-to-be-defined-undefined?
prefer-to-be-un-defined?

SimenB:
What about just `to-be-defined`? I'm not sure if the `un` part needs to be explicit.

SimenB:
Needs a rebase 🙂

SimenB:
needs a rebase again, sorry.

SimenB:
Can you mark the commit with `BREAKING CHANGE` in the body when you rebase?

G-Rath:
@xfumihiro Are you still interested in championing this PR, as it would be cool to get this across the line :)

Since it was done before the typescript migration, it's probably best to create a new PR rather than rebase.

Otherwise if you're not interested (or don't have the time), I'm happy to pick up this PR.

xfumihiro:
@G-Rath go ahead plz :)

G-Rath:
I'm going to close this off for now - overall I'm not against the rule, but this PR has been around for a while, and you should at least be able to archive a similar effect with `no-restricted-matchers` (specifically, you should be able to restrict using `not.toBeUndefined` in favor of `toBeDefined`).

PR #10, opened 2017-11-10
Add prefer-to-be-null rule

xfumihiro:


rules/prefer_to_be_null.js
```diff
@@ -0,0 +1,34 @@
...
+        node.arguments.length == 1 &&
+        node.parent &&
+        node.parent.type === 'MemberExpression' &&
+        node.parent.parent
+      ) {
+        const parentProperty = node.parent.property;
+        const propertyName = parentProperty.name;
+        const argument = node.parent.parent.arguments[0];
+
+        if (propertyName === 'toBe' && argument.value === null) {
```
SimenB: check `toEqual` as well

rules/__tests__/prefer_to_be_null.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+const RuleTester = require('eslint').RuleTester;
+const rules = require('../../').rules;
+
+const ruleTester = new RuleTester();
+
+ruleTester.run('prefer_to_be_null', rules['prefer-to-be-null'], {
+  valid: ['expect(null).toBeNull();'],
+
+  invalid: [
```
SimenB: Could you add a test for `toBeEqual` here (and the others)?

README.md
```diff
@@ -32,6 +32,7 @@ Then configure the rules you want to use under the rules section.
     "jest/no-disabled-tests": "warn",
     "jest/no-focused-tests": "error",
     "jest/no-identical-title": "error",
+    "jest/prefer-to-be-null": "warn",
```
SimenB: IMO this should be off by default, otherwise it's a breaking change

SimenB:
I never got an email that you pushed, sorry about being slow here!

xfumihiro:
No worries!

README.md
xfumihiro: Why so?
Same for all others?

SimenB: If I use `toBe(null)` somewhere in my code, and have activated the recommended rules, my build is potentially broken in a feature release (or we can do a major, but I don't think that's necessary).

other open PRs yes

thymikee: It's a "warn" level, so by default it should not fail the lint command.

xfumihiro: Just removed all warnings. 😓 
Should redo?

SimenB: At work, we have `--no-warnings`

xfumihiro: How about moving forward on this at the moment?

SimenB: Oh, sorry! yeah, let's just roll with it

PR #20, opened 2017-11-24
Add failing test for prefer-to-be-undefined

SimenB:
This throws. We should create a util which analyses a `CallExpression` and gives you useful properties out (like value, matcher, if it's `not`, `resolves` or `rejects`, argument in matcher, etc)

/cc @xfumihiro


SimenB:
@xfumihiro do you think you could fix this separately so we can release? 🙂

Issue #21, opened 2017-11-27
New rule proposal: disallow large snapshots (no-large-snapshots)


anescobar1991:
Hey all,

I've written a rule that I think would benefit the greater Jest community.

`no-large-snapshots` looks at all Jest snapshot files (files with `.snap` extension) and validates that each stored snapshot within those files does not exceed a certain number of lines (by default 35, though this is configurable and I am more than willing to change the default as is seen fit).

The reasoning for the rule is that as a matter of best practice snapshots should be limited in size in order to be more manageable and reviewable. A stored snapshot is only as good as its review and as such keeping it short, sweet, and readable is very important in my mind.

I already wrote the rule (and tests and docs) and it is being used internally by my team so if I was to open a PR to add it to `eslint-plugin-jest` would you consider it?

thymikee:
I'm good with adding that, seems like a nice-to-have. Although I'd switch it off by default or increase the default value as it would be too painful 😄

anescobar1991:
Great will open PR!

SimenB:
Fixed in #24, released in 21.4.0. Thanks!

PR #24, opened 2017-11-30
Adds no-large-snapshots rule

anescobar1991:
Implements `no-large-snapshots` rule as discussed in issue #21. This rule serves to discourage having large snapshots which are difficult to review and can lead to bugs being snuck in.

I added it to the `recommended` config as a `warn` and kept the default max of 50 lines threshold. I can modify this as needed though if you all would like.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const globals = require('./index').environments.globals.globals;
+
 module.exports = {
   extends: ['eslint:recommended', 'prettier'],
   plugins: ['prettier'],
@@ -13,4 +15,10 @@ module.exports = {
     strict: 'error',
     'prettier/prettier': 'error',
   },
+  overrides: [
+    {
+      files: ['*.test.js'],
+      globals,
+    },
+  ],
 };
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ Then configure the rules you want to use under the rules section.
     "jest/no-disabled-tests": "warn",
     "jest/no-focused-tests": "error",
     "jest/no-identical-title": "error",
-    "jest/no-large-snapshots": "warn",
     "jest/prefer-to-have-length": "warn",
     "jest/valid-expect": "error"
   }
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Then configure the rules you want to use under the rules section.
     "jest/no-disabled-tests": "warn",
     "jest/no-focused-tests": "error",
     "jest/no-identical-title": "error",
+    "jest/no-large-snapshots": "warn",
     "jest/prefer-to-have-length": "warn",
     "jest/valid-expect": "error"
   }
@@ -72,8 +73,8 @@ config file:
 }
 ```
 
-See [ESLint
-documentation](http://eslint.org/docs/user-guide/configuring#extending-configuration-files)
+See
+[ESLint documentation](http://eslint.org/docs/user-guide/configuring#extending-configuration-files)
 for more information about extending configuration files.
 
 ## Rules
@@ -83,6 +84,7 @@ for more information about extending configuration files.
 | [no-disabled-tests](docs/rules/no-disabled-tests.md)           | Disallow disabled tests       | ![recommended](https://img.shields.io/badge/-recommended-lightgrey.svg) |                                                             |
 | [no-focused-tests](docs/rules/no-focused-tests.md)             | Disallow focused tests        | ![recommended](https://img.shields.io/badge/-recommended-lightgrey.svg) |                                                             |
 | [no-identical-title](docs/rules/no-unhandled-errors.md)        | Disallow identical titles     | ![recommended](https://img.shields.io/badge/-recommended-lightgrey.svg) |                                                             |
+| [no-large-snapshots](docs/rules/no-large-snapshots.md)         | Disallow large snapshots      |                                                                         |                                                             |
 | [prefer-to-have-length](docs/rules/prefer-to-have-length.md)   | Suggest using toHaveLength()  | ![recommended](https://img.shields.io/badge/-recommended-lightgrey.svg) | ![fixable](https://img.shields.io/badge/-fixable-green.svg) |
 | [prefer-to-be-null](docs/rules/prefer-to-be-null.md)           | Suggest using toBeNull()      |                                                                         | ![fixable](https://img.shields.io/badge/-fixable-green.svg) |
 | [prefer-to-be-undefined](docs/rules/prefer-to-be-undefined.md) | Suggest using toBeUndefined() |                                                                         | ![fixable](https://img.shields.io/badge/-fixable-green.svg) |
--- a/docs/rules/no-large-snapshots.md
+++ b/docs/rules/no-large-snapshots.md
@@ -97,14 +97,15 @@ line 4
 
 ## Options
 
-This rule has option for modifying the threshold number of lines:
+This rule has option for modifying the max number of lines allowed for a
+snapshot:
 
 In an `eslintrc` file:
 
 ```json
 ...
   "rules": {
-    "jest/no-large-snapshots": ["warn", { "sizeThreshold": 12 }]
+    "jest/no-large-snapshots": ["warn", { "maxSize": 12 }]
   }
 ...
 ```
--- a/docs/rules/no-large-snapshots.md
+++ b/docs/rules/no-large-snapshots.md
@@ -1,0 +1,110 @@
+# disallow large snapshots (no-large-snapshots)
+
+When using Jest's snapshot capability one should be mindful of the size of
+created snapshots. As a best practice snapshots should be limited in size in
+order to be more manageable and reviewable. A stored snapshot is only as good as
+its review and as such keeping it short, sweet, and readable is important to
+allow for thorough reviews.
+
+## Usage
+
+Because Jest snapshots are written with back-ticks (\` \`) which are only valid
+with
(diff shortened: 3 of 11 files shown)


ntwb:
Neat rule, but I don't think it should be added to the `recommended` config.

stylelint currently has 5 snapshot tests but each is thousand/s of lines of code.

How stylelint uses snapshots at the moment is to take a large real world CSS file, e.g. BootStrap and lint against it with a specific set of stylelint rules, if a change is made to stylelint that breaks one of these snapshots then we know we have introduced a regression that is more than likely not covered by a unit test https://github.com/stylelint/stylelint/tree/master/system-tests

index.js
```diff
@@ -3,18 +3,25 @@
...
   configs: {
     recommended: {
+      parserOptions: {
+        ecmaVersion: 2015,
+      },
       rules: {
         'jest/no-disabled-tests': 'warn',
         'jest/no-focused-tests': 'error',
         'jest/no-identical-title': 'error',
+        'jest/no-large-snapshots': 'warn',
```
SimenB: I agree it should not be in the default config

index.js
```diff
@@ -3,18 +3,25 @@
...
 const preferToBeUndefined = require('./rules/prefer_to_be_undefined');
 const preferToHaveLength = require('./rules/prefer_to_have_length');
 const validExpect = require('./rules/valid_expect');
 
+const snapshotProcessor = require('./processors/snapshot-processor');
+
 module.exports = {
   configs: {
     recommended: {
+      parserOptions: {
```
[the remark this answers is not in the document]
SimenB: But we shouldn't lint snapshots, so this isn't needed

package.json
```diff
@@ -23,6 +23,7 @@
     "eslint": "^4.10.0",
     "eslint-config-prettier": "^2.7.0",
     "eslint-plugin-prettier": "^2.3.1",
+    "eslint-plugin-self": "^1.0.1",
```
SimenB: should update the lockfile?

.eslintrc.js
```diff
@@ -13,4 +13,12 @@ module.exports = {
     strict: 'error',
     'prettier/prettier': 'error',
   },
+  overrides: [
+    {
+      files: ['*.test.js'],
+      env: {
```
SimenB: can you import the list maintained in the plugin?

rules/no_large_snapshots.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+module.exports = context => {
+  if (context.getFilename().endsWith('.snap')) {
+    const lineLimit = context.options[0] || 50;
```
SimenB: I think it should take an object as options, just to be easier to expand in the future.

rules/__tests__/no_large_snapshots.test.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+
+const noLargeSnapshots = require('../no_large_snapshots');
+
+// was not able to use https://eslint.org/docs/developer-guide/nodejs-api#ruletester for these because there is no way to configure RuleTester to run non .js files
```
SimenB: Is there an issue for that?

If not, should we create one?

docs/rules/no-large-snapshots.md
```diff
@@ -0,0 +1,96 @@
...
+
+When using Jest's snapshot capability one should be mindful of the size of
+created snapshots. As a best practice snapshots should be limited in size in
+order to be more manageable and reviewable. A stored snapshot is only as good as
+its review and as such keeping it short, sweet, and readable is important to
+allow for thorough reviews.
+
+## Rule Details
+
+This rule looks at all Jest snapshot files (files with `.snap` extension) and
```
SimenB: is it possible to ignore any of them? without using `overrides` in the config

rules/no_large_snapshots.js
anescobar1991: Yeah I was debating that. Will do.

package.json
anescobar1991: Yes! I was using `npm` so didn't see that.

docs/rules/no-large-snapshots.md
anescobar1991: Yes it does work. But on snapshot updates it would be wiped.

rules/__tests__/no_large_snapshots.test.js
anescobar1991: I'm looking more into this. I know it wasn't working properly for me when I tried using `ruleTester` to test this rule and I thought I had read in their docs that it does not support non `.js` extensions but I can't find where I read that anymore.

docs/rules/no-large-snapshots.md
SimenB: Should we include some sort of config then, to be able to ignore specific snapshots?

anescobar1991: Specific snapshots or specific snapshot files? If it is the latter then I am not sure how much value that would add, we would just be duplicating eslint functionality. If you mean ignoring specific snapshots then yes I believe that can be done pretty easily, I can work on that today.

SimenB: I meant snapshot files, not specific snapshots within them 🙂

anescobar1991: Thinking about it more I am not sure that being able to ignore a specific snapshot through a config would be a good user experience. They would have to provide the snapshot names in an array which could get unwieldy. It may be better to ignore whole files using built in `eslint` functionality.

How are you envisioning it?

SimenB: Yeah, `.eslintignore` is probably enough :)

anescobar1991:
@SimenB Just pushed changes that takes care of all the feedback. Few things to note: `eslint-plugin-self` is no longer needed and I removed `parserOptions` from the recommended config since this rule is no longer being exported with it.

README.md
```diff
@@ -40,6 +40,7 @@ Then configure the rules you want to use under the rules section.
     "jest/no-disabled-tests": "warn",
     "jest/no-focused-tests": "error",
     "jest/no-identical-title": "error",
+    "jest/no-large-snapshots": "warn",
```
SimenB: no?

anescobar1991: I thought this was a list of all the rules exported that the user can configure independently of the `recommended` config. Why wouldn't it be in there?

SimenB: well, they now match the recommended config. Maybe that's weird? It doesn't list all rules

anescobar1991: Yeah you're right that is weird. As a user I took that as an example of how you could configure the rules exported within this plugin. To me the table near the bottom of the `README` is what lets me know whether a rule is in the `recommended` config or not.

Either way let me know how you want me to proceed.

SimenB: I don't really care, but having 2 lists of all rules doesn't make sense.

Take a look at e.g. https://github.com/yannickcr/eslint-plugin-react (just a couple examples, then a huge list). Or maybe https://github.com/benmosher/eslint-plugin-import, which has 5, then "etc.". followed by a list

anescobar1991: Makes sense. I was just trying to follow what I thought was the pattern already established. But I see that it wasn't a complete list to begin with so my bad! I will remove.

SimenB:
GTG from my side, would just like another set of eyes on it 🙂

docs/rules/no-large-snapshots.md
```diff
@@ -0,0 +1,110 @@
...
+## Options
+
+This rule has option for modifying the threshold number of lines:
+
+In an `eslintrc` file:
+
+```json
+...
+  "rules": {
+    "jest/no-large-snapshots": ["warn", { "sizeThreshold": 12 }]
```
thymikee: Can’t we just name it “size”?

rules/__tests__/__snapshots__/no_large_snapshots.test.js.snap
```diff
@@ -0,0 +1,45 @@
...
+`;
+
+exports[`no-large-snapshots ExpressionStatement function should report if node has more than 50 lines of code and no sizeThreshold option is passed 1`] = `
+Array [
+  Object {
+    "data": Object {
+      "lineCount": 52,
+      "lineLimit": 50,
+    },
+    "message": "Expected Jest snapshot to be smaller than {{ lineLimit }} lines but was {{ lineCount }} lines long",
```
thymikee: Why are the values not replaced here?

thymikee: I’m fine if eslint just picks this up from corresponding data obj, which it probably does (sorry, I’m not an eslint plug-in expert 😅)

SimenB: >I’m fine if eslint just picks this up from corresponding data obj, which it probably does 

It does

rules/__tests__/no_large_snapshots.test.js
SimenB: News on this?

docs/rules/no-large-snapshots.md
SimenB: or `maxSize`?

anescobar1991: Done in https://github.com/jest-community/eslint-plugin-jest/pull/24/commits/8267cfa33139ce1c5f70d1d576c57e6e4394ea6d

anescobar1991:
Anything pending or can this be merged?

SimenB:
I would like to figure out https://github.com/jest-community/eslint-plugin-jest/pull/24/files#r154083575

SimenB:
Release blocked on #20 (#8), unfortunately.

SimenB:
Published as 21.4.0

PR #26, opened 2017-12-04
Refactor prefer rules

xfumihiro:
fixes #20

### DIFF
--- a/rules/prefer_to_have_length.js
+++ b/rules/prefer_to_have_length.js
@@ -1,46 +1,42 @@
 'use strict';
+const expectCase = require('./util').expectCase;
+const expectNotCase = require('./util').expectNotCase;
+const expectResolveCase = require('./util').expectResolveCase;
+const expectRejectCase = require('./util').expectRejectCase;
+const method = require('./util').method;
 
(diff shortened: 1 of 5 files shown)


xfumihiro:
btw, do we need this? [Commit Message Guidelines](https://gist.github.com/abravalheri/34aeb7b18d61392251a2)

SimenB:
I'm all for adding conventional commits, and validating them

rules/prefer_to_have_length.js
```diff
@@ -1,46 +1,44 @@
 'use strict';
+const {
```
SimenB: Destructure doesn't work in node 4. We can revisit dropping node 4 later, but for now can you make this work there?

xfumihiro:
Should we use [this](https://github.com/conventional-changelog/standard-version) then?

SimenB:
I'm thinking of setting up https://github.com/semantic-release/semantic-release/.

SimenB:
Thanks for this! I'll try to set up automatic releases later today or tomorrow.

For now, released as 21.4.0

PR #27, opened 2017-12-04
ci: attempt semantic release

SimenB:
This is an attempt to set up semantic release. Currently setup with my own user, but we can create a service user if we want.

This goes all out, and has linting both on as commit hook and CI on the commit message.


SimenB:
Ultimately I want to get to know semantic release better, and hopefully bring it back into Jest core. For that to happen, we probably need to help out with https://github.com/semantic-release/semantic-release/issues/193, but if we could get some push in Jest, maybe yarn and lerna peeps there might help us out 🙂

SimenB:
We hit issues with node 4 again...

Just drop it?

.travis.yml
```diff
@@ -4,9 +4,17 @@ node_js:
...
+- curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.3.2
+- export PATH="$HOME/.yarn/bin:$PATH"
+- yarn global add greenkeeper-lockfile@1
 before_script: greenkeeper-lockfile-update
+script:
+- yarn commitlint --from="$TRAVIS_BRANCH" --to="$TRAVIS_COMMIT"
+- yarn commitlint --from=$TRAVIS_COMMIT
+- yarn test
 after_script: greenkeeper-lockfile-upload
+after_success: yarn semantic-release
```
thymikee: Can we skip node 4 somehow? If not, let’s drop it and release next version as a major

.travis.yml
```diff
@@ -15,4 +23,6 @@ branches:
...
+  - secure: MBK8f97y4VozvQtJAPkFdoeyKyEfoX9mln/HE4AaAm2XB+xmCVQxFayTcfJoqhL4ZYh4xKrRIndxnnVHexc9aLuJKooC9azBU7jyQXIEKiGxlvo6Y2BOmsWLDZl41KIwfjWRoyCV1Of6Bf2RFyZg1Yut2RGtjYHTD6P4jRKwBNEpE9MyBJLQYmJw0KEwYQiY4Jt3fZtouSgyZS2jBeiBQEyS9mShbISdtjqUN021H/zpdfqvD0UUAehu/PhztiD2x0d23GkcLASMeBUlnrf9Vy5JmRFmrFgw9i4BDg6i1pDooV6xXShV5VoMmtPmSAuCkDvrGI8ETFqUB9799mHmyd1qcjHNWN+xojKfDVE+5G0AnyHWzHwI7PixDhWdNREKhuk8un5ACG+r3EZKyLkfN43RLoid+OHIaZk8Gfo+12Z4b9Z2v5ixjOvD/K8CIC3EKxee4iYRE4QTfeHf5NMoC5lfzOFVfcb84w5I12RK8dHiLRtQa5x6R665w5wGtQQWpRhaXMb4qOcA4jD3WE/g+ugnYJU18uDS6Awlg3MKKhRo+l+T7neusAZTLlqlAp0Y4rvZ9BlJRjIyeIxgC9yTI4Gz+daiskVtNuNh9yAv38MA3q1HP3rfRU1T21SmLBNMRe9uaDARk5jPDjUDQEYfMoGVt78xn+atwz1YdKTG1oA=
```
thymikee: I believe these can go into environment variables in Travis, no?

.travis.yml
SimenB: Could add it as a dependency in the build stages.

.travis.yml
SimenB: It's less work to run `travis encrypt` locally :P.

What's the advantage of having them in repo settings?

SimenB:
That means we lose lockfile guarantees about semantic release (`optionalDevDependencies` isn't a thing, unfortunately), maybe worth it? Can re-add it whenever we drop node 4 I guess

thymikee:
I'm fine with that 👍

SimenB:
Let's go

PR #32, opened 2017-12-05
Next

SimenB:
Opening PR for green CI


SimenB:
@xfumihiro Thanks for being quick to fix the issues! I'm wondering if we should have a huge list of `expect`s we can feed through all the tests's `valid` blocks to ensure the rule doesn't yell (or crash) for stuff it shouldn't.

PR #33, opened 2017-12-05
fix: support identifiers in to-be-undefined

SimenB:
I've ran the plugin with all rules enabled over the jest code base. Found a false positive, but with this change, it looks pretty great 🙂 

`expect(something).to{Be,Equal}(somethingElse)` were the false positives. The other assertions were added as precautions as I broke all of them while fixing the first one.

/cc @xfumihiro

### DIFF
--- a/rules/__tests__/prefer_to_be_undefined.test.js
+++ b/rules/__tests__/prefer_to_be_undefined.test.js
@@ -11,6 +11,10 @@ ruleTester.run('prefer_to_be_undefined', rules['prefer-to-be-undefined'], {
     'expect(true).not.toBeUndefined();',
     'expect({}).toEqual({});',
     'expect(null).toEqual(null);',
+    'expect(something).toBe(somethingElse)',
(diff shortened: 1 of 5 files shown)


rules/__tests__/prefer_to_be_undefined.test.js
```diff
@@ -11,6 +11,10 @@ ruleTester.run('prefer_to_be_undefined', rules['prefer-to-be-undefined'], {
     'expect(true).not.toBeUndefined();',
     'expect({}).toEqual({});',
     'expect(null).toEqual(null);',
+    'expect(something).toBe(somethingElse)',
```
SimenB: these 2 were false positives

SimenB:
There will be a PR incoming adding tons of assertions to `valid` in the different rules to ensure that we don't break them. There have been way too many issues here...

SimenB:
Merging as it's a pretty ugly bug

PR #34, opened 2017-12-06
refactor: rename util to guards & add tests

xfumihiro:
Moved false positive cases into guard tests.
We should put general valid tests in guards from now on.

/cc @SimenB


xfumihiro:
@SimenB can you review this?

xfumihiro:
Can anyone review this?

SimenB:
This is cleaned up as part of the TS migration. Thanks for taking a look at it, and sorry about never reviewing it!

PR #36, opened 2017-12-14
Fix rule link - no-identical-title

chentsulin:


thymikee:
@SimenB I've updated the commit message to `fix:` type, because README is updated in npm registry upon publish.

SimenB:
cool!

PR #38, opened 2017-12-15
feat(recommended): Add plugin and globals env

scriptdaemon:
Add the 'jest' plugin and 'jest/globals' environment to the recommended
config so the user doesn't have to include these manually.


SimenB:
`commitlint` wants lowercase "Add" in the title of the commit

SimenB:
Which I can fix on merge, of course. Thanks!

Issue #41, opened 2017-12-30
Lint rule to indicate missing return staement when testing inside promise then or catch block.


tushardhole:
**[PORTING THIS ISSUE [FROM](https://github.com/facebook/jest/issues/4843) JEST REPO]**

**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
It is mentioned in the [jest docs](https://facebook.github.io/jest/docs/en/tutorial-async.html) to return a promise when you want to test inside its then or catch block. In case, if a developer misses to add a return statement, then that expectation will not get executed and may result in test which is false positive.

**What is the expected behavior?**
A lint rule

**Please provide your exact Jest configuration and mention your Jest, node, yarn/npm version and operating system.**
jest: 21.2.1
node: 8.0.0
npm: 5.5.1
OS: Sierra (10.12.6)

tushardhole:
ported this [PR](https://github.com/jest-community/eslint-plugin-jest/pull/42) for this issue from jest repo.

SimenB:
Fixed in #42

PR #42, opened 2017-12-30
Introduce a lint rule to report error when testing promises. If a exp…

tushardhole:
…ectation is added inside a then or catch block of promise then as per jest docs, it is mandatory to return that promise for succesfull execution of that expectation. This rull will report all such expectations where promise is not returned.

**Summary**
Implements  #41

As jest user I work on a project that heavily uses promises and jest is our tool to write tests. We were aware that as per [jest docs](https://facebook.github.io/jest/docs/en/tutorial-async.html), it is required to return a promise, if you want to test some thing in its fulfillment or rejection block.

If you do not return that promise and write some exceptions in its then or catch block. Then that test will always be green, as the expectation in then or catch block is never executed. If you return that promise then the expectation will be executed and tests will be red or green depending on the actual result of that expectation.

In this situation it becomes very important to make sure that,
a) You make that test red before making it green for the final push. Which will assure that the exception is  getting executed.

As human, we may forget to do that and hence this lint rule will be helpful.

**Test plan**
1. Added a unit test case with many combinations.
2. Tested on an existing project which uses jest & promises.
3. Ran all tests using 'yarn test' command

**Limitations**
1. If the expectation inside then or catch is a expect statement then this rule will work correctly. In case, there is call expression to some function, from then or catch block and that function has the expect statement then this rule will not be able to detect that. I will work on that issue but I feel that as 'Good to Have' and should be next step to improve this rule and current PR changes can be very well considered as 'Step 1' which is very useful in its current state.

### DIFF
--- a/rules/__tests__/valid_expect_in_promise.test.js
+++ b/rules/__tests__/valid_expect_in_promise.test.js
@@ -105,6 +105,19 @@ ruleTester.run('valid-expect-in-promise', rules['valid-expect-in-promise'], {
         },
       ],
     },
+    {
+      code:
+        "it('test function', " +
+        '\n() => {Builder.getPromiseBuilder()..get().build()' +
+        "\n.then((data) => {expect(data).toEqual('Hi');});});",
+      errors: [
+        {
+          column: 8,
+          endColumn: 48,
+          message: expectedMsg,
+        },
+      ],
+    },
   ],
 
   valid: [
@@ -180,5 +193,9 @@ ruleTester.run('valid-expect-in-promise', rules['valid-expect-in-promise'], {
       'const promise = nodeCrawl({}).' +
       "\nthen(data => {expect(childProcess.spawn).lastCalledWith('find');});" +
       '\nreturn promise;});',
+
+    "it('test function', " +
+      '\n() => {return Builder.getPromiseBuilder().get().build()' +
+      "\n.then((data) => {expect(data).toEqual('Hi');});});",
   ],
 });
--- a/rules/__tests__/valid_expect_in_promise.test.js
+++ b/rules/__tests__/valid_expect_in_promise.test.js
@@ -165,5 +165,7 @@ ruleTester.run('valid-expect-in-promise', rules['valid-expect-in-promise'], {
          }).toThrow();
        }));
     `,
+
+    "it('promise test', () => {const somePromise = getThatPromise();\nsomePromise.then((data) => {expect(data).toEqual('foo');});\nexpect(somePromise).toBeDefined();\nreturn somePromise;});",
   ],
 });
--- a/rules/__tests__/valid_expect_in_promise.test.js
+++ b/rules/__tests__/valid_expect_in_promise.test.js
@@ -218,5 +218,7 @@ ruleTester.run('valid-expect-in-promise', rules['valid-expect-in-promise'], {
       "\n.then((data) => {expect(data).toEqual('Hi');});});",
 
     'it("it1", () => somePromise.then(() => {expect(someThing).toEqual(true)}))',
+
+    'it("it1", () => somePromise.then(() => expect(someThing).toEqual(true)))',
   ],
 });
--- a/rules/__tests__/valid_expect_in_promise.test.js
+++ b/rules/__tests__/valid_expect_in_promise.test.js
@@ -118,6 +118,21 @@ ruleTester.run('valid-expect-in-promise', rules['valid-expect-in-promise'], {
         },
       ],
     },
+    {
+      code:
+        'it("it1", () => { somePromise.' +
+        '\nthen(() => {\n' +
+        '\ndoSomeOperation();' +
+        '\nexpect(someThing).toEqual(true);' +
+        '\n})})',
+      errors: [
+        {
+          column: 19,
+          endColumn: 3,
+          message: expectedMsg,
+        },
+      ],
+    },
   ],
 
   valid: [
--- a/rules/__tests__/valid_expect_in_promise.test.js
+++ b/rules/__tests__/valid_expect_in_promise.test.js
@@ -220,5 +220,15 @@ ruleTester.run('valid-expect-in-promise', rules['valid-expect-in-promise'], {
     'it("it1", () => somePromise.then(() => {expect(someThing).toEqual(true)}))',
 
     'it("it1", () => somePromise.then(() => expect(someThing).toEqual(true)))',
+
+    `it('promise test with done', (done) => {
+      const promise = getPromise();
+      promise.then(() => expect(someThing).toEqual(true));
+    });`,
+
+    `it('name of done param does not matter', (nameDoesNotMatter) => {
+      const promise = getPromise();
+      promise.then(() => expect(someThing).toEqual(true));
+    });`,
   ],
 });
--- a/rules/__tests__/valid_expect_in_promise.test.js
+++ b/rules/__tests__/valid_expect_in_promise.test.js
@@ -148,5 +148,22 @@ ruleTester.run('valid-expect-in-promise', rules['valid-expect-in-promise'], {
     "it('it1', () => {return somePromise." +
       'then(() => {expect(someThing).toEqual(true);})' +
       '.catch(() => {expect(someThing).toEqual(false);})});',
+    `
+     test('later return', () => {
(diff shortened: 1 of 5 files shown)


SimenB:
CI is failing because of non-semantic commit message. See http://karma-runner.github.io/2.0/dev/git-commit-msg.html

rules/__tests__/valid_expect_in_promise.test.js
```diff
@@ -0,0 +1,135 @@
...
+
+    'it("it1", function() { return Promise.resolve().then(' +
+      'function() { /*fulfillment*/ expect(someThing).toEqual(true)}, ' +
+      'function() { /*rejection*/ expect(someThing).toEqual(true)})})',
+
+    'it("it1", function() { return Promise.resolve().then(' +
+      'function() { /*fulfillment*/}, ' +
+      'function() { /*rejection*/ expect(someThing).toEqual(true)})})',
+
+    'it("it1", function() { return somePromise.then()})',
```
SimenB: can you add assertions that `await` is ok without `return`?

tushardhole: I will work on that.

tushardhole:
I will work on fixing  non-semantic commit message.

tushardhole:
@SimenB I have done the semantic commit message fix and a new commit to allow await expressions without return statement

tushardhole:
@SimenB doc changes done! Please review the same.

Also did some sanity on a project which uses promises and jest. Found one bug with nested then or catch, added unit test & fixed  😊

tushardhole:
@SimenB thanks for that feedback! I will work on that.

tushardhole:
@SimenB When i tried to run the rule against jest, the rule was not running at all, I believe following was the cause,

In line no 21 of .eslintrc.js in jest code base, it says to check only .md files, I.e. `files: ['*.md']`,
when I changed the line to below, the rule started reporting,
`files: ['*.js', '*.md'],` 

To summarize I did following steps to test this rule agains jest code base,
1. Copied `index.js` and `rules dir` from `eslint-plugin-jest` to `node_modules/eslint-plugin-jest/` of jest code base

2. Changed the block on line 20 in .eslintrc.js to this,
```
    {
      files: ['*.js', '*.md'],
      rules: {
        'consistent-return': 0,
        'flowtype/require-valid-file-annotation': 0,
        'import/no-extraneous-dependencies': 0,
        'import/no-unresolved': 0,
        "jest/valid-expect-in-promise": 2,
        'jest/no-focused-tests': 0,
        'jest/no-identical-title': 0,
        'no-undef': 0,
        'no-unused-vars': 0,
        'prettier/prettier': 0,
        'react/jsx-no-undef': 0,
        'react/react-in-jsx-scope': 0,
        'sort-keys': 0,
        'unicorn/filename-case': 0,
      },
    }
```

Here is the output,
```
/Users/tushard/OpenProjects/jest/packages/jest-haste-map/src/crawlers/__tests__/node.test.js
  103:11  error  Promise should be returned to test its fulfillment or rejection  jest/valid-expect-in-promise

/Users/tushard/OpenProjects/jest/packages/jest-runtime/src/__tests__/runtime_module_directories.test.js
  12:48  error  Promise should be returned to test its fulfillment or rejection  jest/valid-expect-in-promise

/Users/tushard/OpenProjects/jest/packages/jest-runtime/src/__tests__/runtime_require_mock.test.js
  133:58  error  Promise should be returned to test its fulfillment or rejection  jest/valid-expect-in-promise

/Users/tushard/OpenProjects/jest/packages/jest-runtime/src/__tests__/runtime_require_module.test.js
  134:56  error  Promise should be returned to test its fulfillment or rejection  jest/valid-expect-in-promise
  159:54  error  Promise should be returned to test its fulfillment or rejection  jest/valid-expect-in-promise
```

Let me know your thoughts, we may need to add `.js` in `.eslintrc.js` of jest

SimenB:
To test the rule in the Jest repo, you have to add the rule here: https://github.com/facebook/jest/blob/c799a0251d84a7e0c04d98e7d921a5a0e707c72f/packages/eslint-config-fb-strict/index.js#L50-L52

tushardhole:
@SimenB Thanks,  I will test the rule that way. 

On other hand , I don't find the previousaly mentioned false positives in the output posted in my previous comment. Do you find that as the expectated output is or something is missing in that.

SimenB:
>On other hand , I don't find the previousaly mentioned false positives in the output posted in my previous comment.

They are in your paste, in particular
```
/Users/tushard/OpenProjects/jest/packages/jest-haste-map/src/crawlers/__tests__/node.test.js
  103:11  error  Promise should be returned to test its fulfillment or rejection  jest/valid-expect-in-promise
```

and 
```
/Users/tushard/OpenProjects/jest/packages/jest-runtime/src/__tests__/runtime_module_directories.test.js
  57:48  error  Promise should be returned to test its fulfillment or rejection  jest/valid-expect-in-promise

```

SimenB:
I'm having issues creating a test for the case of arrow shorthand, but I pushed up a test for one of the failing cases at least

rules/__tests__/valid_expect_in_promise.test.js
```diff
@@ -0,0 +1,169 @@
...
+         expect(value).toBe('red');
+       });
+
+       return promise;
+     });
+    `,
+    `
+     it('shorthand arrow', () =>
+       something().then(value => {
+         expect(() => {
```
SimenB: this passes if I add `return` before `expect`

rules/__tests__/valid_expect_in_promise.test.js
```diff
@@ -0,0 +1,169 @@
...
+          message: expectedMsg,
+        },
+      ],
+    },
+    {
+      code:
+        'it("it1", function() { Promise.resolve().then(' +
+        'function() { /*fulfillment*/ expect(someThing).toEqual(true)}, ' +
+        'function() { /*rejection*/ expect(someThing).toEqual(true)})})',
+      errors: [
```
SimenB: why is this 2 errors? It should just be one, there is only a single promise here

SimenB:
Another false positive is if I use the `done` callback.

```js
test('callback in handler', done => {
 something().then(value => {
   expect(value).toBe('red');
   
   done();
  });
});
```

This shouldn't warn, but it's also weird usage of promises, so I think we should just not warn here. (if it's easy to solve, please do so)

tushardhole:
> This shouldn't warn, but it's also weird usage of promises, so I think we should just not warn here. (if it's easy to solve, please do so)

Yes I will do that. I think, this rule should just ignore processing if the test function have done parameter. Also i verified, the name of parameter can be anything not necessary to be named as done only.

Whether that done function is actually called or not, should not be verified by this rule. We can add a different rule for that like `prefer-calling-done`, thoughts ?

rules/__tests__/valid_expect_in_promise.test.js
tushardhole: I will work on this.

rules/__tests__/valid_expect_in_promise.test.js
tushardhole: I will fix that.

rules/__tests__/valid_expect_in_promise.test.js
tushardhole: There are two more similar cases,
1. Where expect is inside a block,
```
test('invalid return', () => {
       const promise = something().then(value => {
         { expect(value).toBe('red'); }
    });
 });
```
2. Where expect is inside another method,
```
test('invalid return', () => {
       const promise = something().then(value => {
         anotherMethod(); //this method contains the expect stament
    });
 });
```
I have already added this limitation(#>2) in the PR description. I think these are rare scenarios, the first one may not occur at all. As mentioned earlier in the PR description, I think they are `good to have` can be done in next enhancement to this rule.
If this scenarios are `must to have` , then I am happy to implement in this PR itself.

tushardhole:
@SimenB below feedback's are fixed,
1. short hand arrow function ✔️
2. later promise return ✔️
3. done async callback ✔️
4. duplicate errors for same promise ✔️

SimenB:
Ran it on the jest code base, and it actually found an error! https://github.com/facebook/jest/pull/5336

Great job!

PR #43, opened 2018-01-02
Missing expect assertions call

tushardhole:
Adding new lint rule which validates,

**Every test to have expect.assertions({number of assertions}) as first line**
_**OR**_
**Every test to have expect.hasAssertions() as first line**


tushardhole:
Looking for someone to help me understand how to solve below CI issues,
⧗   input: Enhance the missing_expect_assertions rule to,
✖   message may not be empty [subject-empty]
✖   type may not be empty [type-empty]
⚠   body must have leading blank line [body-leading-blank]
✖   found 2 problems, 1 warnings
⧗   input: Enhance missing expect assertions call rule to report error when expect.assertions() call is present but the argument provided is not a number.
✖   header must not be longer than 72 characters [header-max-length]
✖   message may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 3 problems, 0 warnings

I am not able to figure out what is subject, header,body etc.

cc: @SimenB

SimenB:
It's yelling at you for not having semantic commits. See http://karma-runner.github.io/2.0/dev/git-commit-msg.html

tushardhole:
@SimenB Thanks for the  help   :thumbsup:

I will follow those conventions and verify it locally before pushing.

I am thinking to amend the commits on my branch and then do force push. Will that work or I am required to create a new PR.

giodamelio:
Not sure about the name, seems like `prefer-expect-assertions` would be more consistent with existing rules.

rules/__tests__/missing_expect_assertions_call_test.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+const RuleTester = require('eslint').RuleTester;
+const rules = require('../..').rules;
+
+const ruleTester = new RuleTester();
+const expectedMsg =
+  'Every test should have expect.assertions({number of assertions}) OR expect.hasAssertions() as first expression';
```
giodamelio: I think this might read better as

    Every test should have either `expect.assertions(<number of assertions>)` or `expect.hasAssertions()` as it's first expression

rules/missing_expect_assertions_call.js
```diff
@@ -0,0 +1,82 @@
...
+  });
+};
+
+module.exports = context => {
+  return {
+    CallExpression(node) {
+      if (isTestOrItFunction(node)) {
+        if (!isFirstLineExprStmt(node)) {
+          reportMsg(context, node);
+          return;
```
giodamelio: I think this return statement is unnecessary?

tushardhole:
@giodamelio Thank for your feedback. I have renamed rule to `prefer-expect-assertions` and updated rule message.

@SimenB All CI checks are green. Thank you for your help! 🙂

tushardhole:
@SimenB added the missing coverage.

SimenB:
Thanks! Released as 21.6.0

SimenB:
@tushardhole we forgot to update the README 😱 Mind sending a PR for that? 🙂

tushardhole:
@SimenB I would create an PR for README

SimenB:
Docs published in 21.6.1

PR #45, opened 2018-01-05
chore: update jest to version 22

SimenB:
Closes #39 

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_install:
 - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.3.2
 - export PATH="$HOME/.yarn/bin:$PATH"
 - yarn global add greenkeeper-lockfile@1
+install: yarn --ignore-engines
 before_script: greenkeeper-lockfile-update
 script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -8,6 +8,7 @@ before_install:
 - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.3.2
 - export PATH="$HOME/.yarn/bin:$PATH"
 - yarn global add greenkeeper-lockfile@1
+install: yarn --ignore-engines
```
thymikee: Why is this needed now?

SimenB: https://github.com/facebook/jest/blob/9cbc26bab656fa9e9952242b2d733a86c6a61d43/packages/jest/package.json#L13

ntwb: What about dropping Node.js 4.x support here, if Jest requires > Node.js 6.x?

SimenB: Mainly because we don't really need to, and this plugin isn't necessarily bound to a single version of Jest

thymikee: @ntwb Also, Jest supports Node 4 with custom environment targeting jsdom@9

PR #47, opened 2018-01-09
chore(package): add semantic-release as a dev dependency

SimenB:
Since we use `--ignore-engines` when installing, this is now ok

Might even speed up the build a tiny bit.


SimenB:
Such green, nice

PR #48, opened 2018-01-09
fix: remove `prefer-to-have-length` from recommended rules

SimenB:
I don't think this is a rule that makes sense as a recommended rule (even though I personally would enable it).

As it's so tiny, we can hold off on merging (or just not merge it at all if people think it should remain in `recommended`)


thymikee:
Actually I'd keep it. If there's a better matcher to do the job, why not recommend it? Especially as a warning.

ntwb:
FYI: It caught me out the other day, I had 21.0.x installed, and a CI reported back the error because it doesn't honour lock files!. That said I happily switched 😄

SimenB:
Ah, adding it in the first place was actually a breaking change

PR #51, opened 2018-01-11
Also warn on .only (skipped 'all but this')

jez:
`it.only()` is like the inverse of `it.skip()`. Since `it.only()` skips all
tests **but** the current test, we should lint against this one as well.


jez:
Ah, nevermind! I see there's also a 'disallow focused tests' feature.

PR #53, opened 2018-01-13
fix(valid-expect): error on missing async matchers

SimenB:
We got an error previously as well, but with a misleading message (`"resolves" was not called.`).

### DIFF
--- a/rules/__tests__/valid_expect.test.js
+++ b/rules/__tests__/valid_expect.test.js
@@ -107,7 +107,7 @@ ruleTester.run('valid-expect', rules['valid-expect'], {
         {
           endColumn: 22,
           column: 14,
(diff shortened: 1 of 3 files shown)


rules/__tests__/valid_expect.test.js
```diff
@@ -101,5 +101,35 @@ ruleTester.run('valid-expect', rules['valid-expect'], {
         },
       ],
     },
+    {
+      code: 'expect(true).resolves;',
+      errors: [
+        {
+          endColumn: 22,
+          column: 14,
+          message: '"resolves" needs a matcher.',
```
thymikee: maybe "needs to call a matcher"?

SimenB: done

Issue #46, opened 2018-01-09
Enforce it block to begin with lower case test decription


ismail-syed:
Some projects prefer that descriptions in `it` blocks beginning with lowercase letters. Is this something that can be enforced as a rule in `eslint-plugin-jest`

eg:

The following patterns are considered warnings: 
```
it('Adds 1 + 2 to equal 3', () => {
  expect(sum(1, 2)).toBe(3);
});
```

These patterns would not be considered warnings:
```
it('adds 1 + 2 to equal 3', () => {
  expect(sum(1, 2)).toBe(3);
});
```

SimenB:
I see no reason not to have a rule for this. @thymikee thoughts?

thymikee:
Makes sense, just don't make it recommended/default 😄

PR #56, opened 2018-01-14
Add lowercase-description rule

ismail-syed:
Fixes https://github.com/jest-community/eslint-plugin-jest/issues/46


rules/__tests__/lowercase_description.test.js
```diff
@@ -0,0 +1,40 @@
...
+    "it('foo', function () {})",
+    'it("foo", function () {})',
+    'it(`foo`, function () {})',
+    'it("<Foo/>", function () {})',
+    'it("123 foo", function () {})',
+    'describe("foo", function () {})',
+    'test("foo", function () {})',
+  ],
+
+  invalid: [
```
SimenB: I think it should warn for `describe` and `it` as well

macklinu: I agree, I think that'd be nice!

SimenB:
@ismail-syed ping 🙂

ismail-syed:
Sorry for the delay, thanks for the pin @SimenB. 
Comments have been addressed. Not sure why CI is complain, any suggestions?

rules/lowercase_description.js
```diff
@@ -0,0 +1,61 @@
...
+  const foo =
+    node.arguments &&
+    node.arguments[0] &&
+    (node.arguments[0].type === 'Literal' ||
+      node.arguments[0].type === 'TemplateLiteral');
+  return foo;
+};
+
+const testDescription = node => {
+  const { type } = node.arguments[0];
```
macklinu: Object destructuring is not supported on Node 4, which is causing the Node 4 failure on Travis.

Related: https://github.com/jest-community/eslint-plugin-jest/pull/64#discussion_r167413883

rules/lowercase_description.js
```diff
@@ -0,0 +1,61 @@
...
+    node.type === 'CallExpression' &&
+    node.callee &&
+    (node.callee.name === 'it' ||
+      node.callee.name === 'test' ||
+      node.callee.name === 'describe')
+  );
+};
+
+const isItDescription = node => {
+  const foo =
```
macklinu: Minor suggestion: return this conditional statement instead of assigning to `foo`, since the variable doesn't explain what's going on in this function.

rules/lowercase_description.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+const ruleMsg = 'it() description should begin with lowercase';
+
+const isItOrTestorDescribeFunction = node => {
```
macklinu: Minor function renaming suggestion: `isItTestOrDescribeFunction`

rules/lowercase_description.js
```diff
@@ -0,0 +1,61 @@
...
+    return description[0] === description[0].toLowerCase();
+  }
+  return true;
+};
+
+module.exports = {
+  meta: {
+    docs: {
+      url:
+        'https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/lowercase_description.md',
```
macklinu: `lowercase_description` should be `lowercase-description` to match the `lowercase-description.md` filename.

I think if you have master merged in, this is autofixable with ESLint (introduced in #60).

rules/lowercase_description.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+  if (type === 'TemplateLiteral') {
+    return node.arguments[0].quasis[0].value.raw;
+  }
+};
+
+const descriptionBeginsWithLowerCase = node => {
+  if (isItOrTestorDescribeFunction(node) && isItDescription(node)) {
+    const description = testDescription(node);
+    return description[0] === description[0].toLowerCase();
```
macklinu: Looking at the definition of `testDescription()`, it can technically return undefined if the first argument's type is not `Literal` or `TemplateLiteral`, or if the `Literal` provided is not a string.

For example, the following code:

```js
it(42, () => {})
```

will result in `TypeError: Cannot read property 'toLowerCase' of undefined` since the first argument of `it()` is a `Literal`, but it's not a string.

While passing a number as the first argument of `it()` isn't really valid, I think this rule should protect against a possibility like this.

rules/lowercase_description.js
ismail-syed: Ops, forgot to remove this when I poking around. Nice catch.

rules/lowercase_description.js
ismail-syed: I did have the merged changes. The issue was that the file name was `rules/lowercase_description.js`, instead of `rules/lowercase-description.js`.

rules/lowercase_description.js
ismail-syed: Good catch, updated and added some tests to check this as well.

rules/lowercase-description.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+const ruleMsg = 'it() description should begin with lowercase';
```
SimenB: `it` part should reflect `test` or `describe` as well

PR #57, opened 2018-02-03
feat(rule): add valid-describe rule

### LINKED ISSUE #18: Prevent async `describe`
Thanks for this great plugin!

A mistake I've seen some people make is using an async function for `describe`, which then makes jest fail silently for a case like the following:

```js
describe('sample case', () => {
  it('works', () => {
    expect(true).toEqual(true);
  });
  describe('async', async () => {
    await new Promise(setImmediate);
    it('breaks', () => {
      throw new Error('Fail');
    });
  });
});
```

It would be nice to have a rule that prevents this. Same would go for returning a promise (or any other value) inside `describe`. 

```js
describe('sample case', () => {
  it('works', () => {
    expect(true).toEqual(true);
  });
  describe('async', () => {
    return Promise.resolve().then(() => {
      it('breaks', () => {
        throw new Error('Fail');
      });
    });
  });
});
```

What do you guys think?

macklinu:
Resolves #18 

This rule validates that the second parameter of a `describe()` function is a callback function. This callback function:

* should not be [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)
* should not contain any parameters
* should not contain any `return` statements

File naming is kebab-case, per the discussion in https://github.com/jest-community/eslint-plugin-jest/pull/60.


index.js
```diff
@@ -1,5 +1,4 @@
-'use strict';
```
macklinu: I'm not sure why lint-staged is removing this as part of the `eslint --fix` step in my local environment.

SimenB:
Fitting, my first eslint rule ever was handling a callback in `describe` for jasmine! 🙂 See https://github.com/tlvince/eslint-plugin-jasmine/pull/17, can probably copy that implementation

index.js
SimenB: you probably have an `eslintrc` file in `$HOME` or something. Try adding `root: true` in the `eslintrc.js` in this repo

macklinu: Thanks for the tip! I had an unexpected `.eslintrc` file one level up from this repo on my machine. Removed that and was able to see my local linting match up with what Travis CI is reporting. 👍

macklinu:
I'll work towards including the two other use cases in this PR - thanks for the suggestions on other implementations to look at. We can figure out a proper name for this rule before merging, but I won't worry much about that right now.

macklinu:
I'm still working on use case number 2, but is there ever a reason to return within a `describe()` callback?

```js
// invalid
describe('foo', () => {
  return Promise.resolve().then(() => {
    it('breaks', () => {
      throw new Error('Fail')
    })
  })
})
```

Looking at this code, `return Promise.resolve()` is something we want this rule to detect and report. 886ce50 is simply checking if the `describe()` callback body contains any `return` statements, not necessarily that it's returning a Promise.

I can update to check that the returned value is a Promise (by looking at #42, as suggested), but I am wondering if there's a code example where returning anything within the body of a `describe()` callback is a good practice.

macklinu:
I'm still unsure about the previous comment. Also, is there ever a reason to call `await` within a `describe()` block?

```js
// we will warn on the `async` describe callback function
describe('foo', async () => {
  // should we also warn that we are calling `await` here
  // or is the `async` warning above enough?
  await something()
  it('does something', () => {
    // ...
  })
})
```

SimenB:
You have to call `it` synchronously, so no, there's never a reason to make `describe` async

docs/rules/valid-describe.md
```diff
@@ -0,0 +1,53 @@
+# Enforce valid `describe()` callback (valid-describe)
```
macklinu: I settled on calling this `valid-describe` since this addresses a few different cases. This follows the convention of the `valid-expect` rule. Open to other naming suggestions too!

rules/valid-describe.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const describeAliases = Object.assign(Object.create(null), {
```
macklinu: This was taken from `no-identical-title`:

https://github.com/jest-community/eslint-plugin-jest/blob/5e5ceba70f13ed2a909744adde82a6ed15fe4768/rules/no_identical_title.js#L3-L9

I can extract an `isDescribe(node)` function, move it to `util.js` and import it into `no-identical-title` and this rule, if that refactor would be helpful.

rules/valid-describe.js
```diff
@@ -0,0 +1,66 @@
...
+  meta: {
+    docs: {
+      url:
+        'https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-describe.md',
+    },
+  },
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (node && isDescribe(node)) {
```
macklinu: Can `node` ever be falsy here? I'm not sure I need to include this `node &&`.

rules/valid-describe.js
```diff
@@ -0,0 +1,66 @@
...
+            if (isAsync(callbackFunction)) {
+              context.report({
+                message: 'No async describe callback',
+                node: callbackFunction,
+              });
+            }
+            if (hasParams(callbackFunction)) {
+              context.report({
+                message: 'Unexpected argument in describe callback',
+                node: callbackFunction.params[0],
```
macklinu: This only warns for the first param - perhaps this should warn for all passed in params?

rules/valid-describe.js
SimenB: yeah, that sounds good

rules/valid-describe.js
SimenB: yeah, that makes sense.

rules/valid-describe.js
SimenB: not sure. Seems weird if it can be?

macklinu: Does seem weird, will remove that check

rules/valid-describe.js
macklinu: Will update with a test case for multiple params in a `describe()` callback

SimenB:
needs a rebase

macklinu:
Rebased and force pushed up.

rules/__tests__/valid-describe.test.js
```diff
@@ -0,0 +1,195 @@
...
+    describe('foo', () => {
+      it('bar', async () => {
+        expect(await Promise.resolve(42)).toBe(42)
+      })
+    })
+    `,
+  ],
+  invalid: [
+    {
+      code: 'describe("foo", async () => {})',
```
SimenB: as a follow up, mind adding a rule that `describe(() => {})` is invalid as well?

macklinu: For sure, I'll open an issue to track that suggestion. 👍

PR #58, opened 2018-02-06
Add guard to arguments

tinovyatkin:
I'm getting an error `cannot read property [0] of undefined` point to `arguments` access here while using ESLint in VSCode, it seems like ESLint while running on incomplete code can call this with `arguments` undefined. Adding guard should fix that.


SimenB:
@thymikee Not sure I follow. `commitlint` runs separately already, so it just fails the whole build, it doesn't stop the next script from running. We could (and maybe should) separate `lint` and `test` though, so one can see test output even if lint fails 🙂

SimenB:
https://github.com/jest-community/eslint-plugin-jest/commit/911bba1d2578e20edf987a16033b99230d5009b1

SimenB:
@tinovyatkin ping 🙂Just adding a code snippet which shows the error you got would be great!

steelbrain:
`eslint-plugin-jest` is crashing for me at `expect(undefined).toBe` on v22.0.1
I think this PR will fix it

The stack trace of the failure is

```
TypeError: Cannot read property '0' of undefined
at argument (node_modules/eslint-plugin-jest/rules/util.js:76:54)
at expectToBeCase (node_modules/eslint-plugin-jest/rules/util.js:34:3)
...
```

PR #59, opened 2018-02-07
chore: add eslint-plugin-eslint-plugin

macklinu:
This PR adds [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) and its recommended configuration to this project's ESLint configuration. It has some useful rules for ESLint plugin codebases.

When adding the recommended configuration, there were three reports of this codebase violating the [require-meta-fixable](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-fixable.md) rule, which I have addressed in this PR.

In the future, I would like to include and fix the [require-meta-docs-url](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-url.md) rule as well, but I wanted to see if you were open to this addition to the codebase first.

It may be easier to append `?w=1` to the URL in your browser when reviewing this PR (like [this](https://github.com/jest-community/eslint-plugin-jest/pull/59/files?w=1)), which will show the diff with whitespace ignored.

### DIFF
--- a/rules/prefer_to_be_null.js
+++ b/rules/prefer_to_be_null.js
@@ -8,31 +8,36 @@ const expectNotToBeCase = require('./util').expectNotToBeCase;
 const method = require('./util').method;
 const method2 = require('./util').method2;
 
-module.exports = context => {
-  return {
-    CallExpression(node) {
-      const is = expectToBeCase(node, null) || expectToEqualCase(node, null);
-      const isNot =
-        expectNotToEqualCase(node, null) || expectNotToBeCase(node, null);
+module.exports = {
+  meta: {
+    fixable: 'code',
+  },
+  create(context) {
+    return {
+      CallExpression(node) {
+        const is = expectToBeCase(node, null) || expectToEqualCase(node, null);
+        const isNot =
+          expectNotToEqualCase(node, null) || expectNotToBeCase(node, null);
 
-      if (is || isNot) {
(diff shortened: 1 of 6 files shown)


rules/prefer_to_be_null.js
```diff
@@ -8,31 +8,36 @@ const expectNotToBeCase = require('./util').expectNotToBeCase;
...
 
-module.exports = context => {
-  return {
-    CallExpression(node) {
-      const is = expectToBeCase(node, null) || expectToEqualCase(node, null);
-      const isNot =
-        expectNotToEqualCase(node, null) || expectNotToBeCase(node, null);
+module.exports = {
+  meta: {
+    fixable: 'code',
```
SimenB: Wee, awesome! This is a change I wanted to make, nice that this yells at us

SimenB:
Feel free to send PR for `require-meta-docs-url` as well 🙂

rules/prefer_to_be_null.js
SimenB: should probably make all of the rules be an object with a `create` function, even if they don't have any other properties

macklinu:
Thanks, I'll open a PR for `require-meta-docs-url` later today 🙂

PR #60, opened 2018-02-08
fix: include meta.docs.url for all rules

macklinu:
This enables and fixes the [`eslint-plugin/require-meta-docs-url`](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-url.md) rule. The diff is large since most of the rules were converted from `function(context)` to `{ meta, create }` signatures (it's much more manageable to look at when appending `?w=1` to the pull request URL).

One way this PR could be improved:

The `require-meta-docs-url` rule is fixable, given the following option:

```json
{
  "eslint-plugin/require-meta-docs-url": ["error", {
    "pattern": "https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/{{name}}.md"
  }]
}
```

The rule will validate all `meta.docs.url` properties to match the provided pattern, replacing `{{name}}` with the filename.

The current issue is that the rule filenames are snake-cased (`valid_expect`) and the doc filenames are kebab-cased (`valid-expect`), so the `require-meta-docs-url` rule will generate links that 404. Would it be worthwhile to rename the rule and test files to use kebab-casing, or perhaps rename the documentation files to use snake-casing? I don't know what the preferred convention is in this codebase, and I didn't want to introduce a change like that without pointing it out and discussing first.

(meta: I wasn't sure if the commit prefix should be chore, refactor, or feat, so please feel free to comment with what I should change it to or change it when squashing - thanks 😄 )


SimenB:
>Would it be worthwhile to rename the rule and test files to use kebab-casing

Yes please, it is the way it is because that's the way it is in the core Jest repo. Feel free to normalize to kebab-case.

>I wasn't sure if the commit prefix should be chore, refactor, or feat

depends on if we want it to be released or not. Will this make any difference too tooling? If yes, I guess `fix` is fine, otherwise `chore`

macklinu:
Thanks, will push up some new commits to address this, @SimenB 👍

macklinu:
This is ready for another look.

PR #62, opened 2018-02-09
style(lint): enable eqeqeq rule

macklinu:
https://eslint.org/docs/rules/eqeqeq

Thoughts on enabling this rule? I think it's a good practice, since `==` can lead to some unexpected behavior (some examples in the link above).


SimenB:
Sure, makes sense!

Issue #12, opened 2017-11-11
suggestion: prefer `test` or `it`


dylang:
# Suggest using `test` or `it` (always-test-or-it)

For consistency in code, always use `test` or always use `it`.

## Rule details

When `test` is preferred, this rule triggers a warning for any instances of `it`.
When `it` is preferred, this rule triggers a warning for any instances of `test`.

This rule is not set by default.

### Default configuration

The following pattern is considered warning:

```js
it('example 1', () => {
   expect(files).toHaveLength(1);
});

test('example 2', () => {
   expect(files).toHaveLength(1);
});
```

The following pattern is not warning when `test` is set:

```js
test('example 1', () => {
   expect(files).toHaveLength(1);
});

test('example 2', () => {
   expect(files).toHaveLength(1);
});
```

The following pattern is not warning when `it` is set:

```js
it('example 1', () => {
   expect(files).toHaveLength(1);
});

it('example 2', () => {
   expect(files).toHaveLength(1);
});
```

SimenB:
Good idea! Should also be autofixable. Would you send a PR for it? 🙂

Couple of q's:
- What should the default be?
- Should we allow config to differentiate between an `it` within a `describe` and by itself at the top level?

macklinu:
> Should we allow config to differentiate between an it within a describe and by itself at the top level?

I think that could be nice.

When using `describe()`, I prefer using only `it()`.

```js
describe('myFunction()', () => {
  it('does this thing')
  it('does that thing')
  // etc.
})
```

When not using `describe()`, I prefer using only `test()`.

```js
test('does this thing')
test('does that thing')
```

I've worked in codebases that have done one or the other, so being able to configure this could help ensure consistency in those codebases.

macklinu:
> What should the default be?

I'm not sure what the default should be here. :thinking: Does a rule need a default, or could we force an option to be provided? (Perhaps that's not a best practice of ESLint rules?)
 
The enum option I imagine with an example `.eslintrc` usage:

* `test`: all tests must use the `test()` function
* `it`: all tests must use the `it()` function
* `describe/it`: all tests must use the `it()` function

```json
{
  "rules": {
    "jest/always-test-or-it": ["error", "test"]
  }
}
```

I would like to contribute a PR for this unless anyone else would like to.

SimenB:
I would like to force top level to be `test` and for all nested within `describe` to be `it`.

Not sure how to best formulate that.

[...]

PR #64, opened 2018-02-10
Consistent-test-it rule

ranyitz:
Following the discussion in #12 

## Summary
This rule gives you control over the usage of these keywords in your codebase.
You can control which style you prefer, within describe or outside.

- [x] `{"fn": "it"}` enforce `it` calls and fix `test` to be `it`.
- [x] `{"fn": "test"}` enfore `test` calls and fix `it` to be `test`.
- [x] `{"fn": "test", "withinDescribe":"it"}` enforce `test` on top level tests and enforce `it` when nested inside of `describe`, fix `it` to `test` when using in top level, fix `test` to `it` when using inside of `describe`.
- [x] add documentation.


README.md
```diff
@@ -80,6 +80,7 @@ for more information about extending configuration files.
 
 | Rule                                                               | Description                                                     | Recommended                                                             | Fixable                                                     |
 | ------------------------------------------------------------------ | --------------------------------------------------------------- | ----------------------------------------------------------------------- | ----------------------------------------------------------- |
+| [consistent-test-it](docs/rules/consistent-test-it.md)             | Enforce consistent test or it keyword                           | ![recommended](https://img.shields.io/badge/-recommended-lightgrey.svg) | ![fixable](https://img.shields.io/badge/-fixable-green.svg) |
```
SimenB: Not recommended

rules/consistent-test-it.js
```diff
@@ -0,0 +1,124 @@
...
+        if (
+          isTestCase(node) &&
+          describeNestingLevel === 0 &&
+          nodeName.indexOf(testKeyword) === -1
+        ) {
+          const opositeTestKeyword = getOpositeTestKeyword(testKeyword);
+
+          context.report({
+            message:
+              "Prefer using '{{ testKeyword }}' on '{{ opositeTestKeyword }}'",
```
SimenB: Switch "on" with "instead of" perhaps?

ranyitz: I agree, "instead of" is better.

README.md
ranyitz: badge removed

rules/consistent-test-it.js
```diff
@@ -0,0 +1,124 @@
+'use strict';
+
+const { getNodeName, isTestCase, isDescribe } = require('./util');
```
SimenB: cannot use destructuring due to node 4 support - see failing CI

macklinu: May be worth adding the [`es5/no-destructuring`](https://github.com/nkt/eslint-plugin-es5#list-of-supported-rules) rule to help prevent/communicate this.

rules/consistent-test-it.js
```diff
@@ -0,0 +1,124 @@
...
+          withinDescribe: {
+            enum: ['it', 'test'],
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+  },
+  create(context) {
+    const configObj = context.options[0];
```
macklinu: I think you could declare this like:

```js
const configObj = context.options[0] || {}
```

And then you're always dealing with a truthy object, so that on the following lines, you do not need to call `configObj && configObj.propertyName`.

rules/consistent-test-it.js
```diff
@@ -0,0 +1,124 @@
...
+        if (isDescribe(node)) {
+          describeNestingLevel++;
+        }
+
+        if (
+          isTestCase(node) &&
+          describeNestingLevel === 0 &&
+          nodeName.indexOf(testKeyword) === -1
+        ) {
+          const opositeTestKeyword = getOpositeTestKeyword(testKeyword);
```
macklinu: `oposite` is misspelled - should be `opposite`.

rules/consistent-test-it.js
```diff
@@ -0,0 +1,124 @@
...
+      'CallExpression:exit'(node) {
+        if (isDescribe(node)) {
+          describeNestingLevel--;
+        }
+      },
+    };
+  },
+};
+
+function getPreferedNodeName(nodeName, preferedTestKeyword) {
```
macklinu: `prefered` is mispelled - should be `preferred`.

rules/consistent-test-it.js
macklinu: Opened https://github.com/jest-community/eslint-plugin-jest/pull/66 to address this.

ranyitz: Yes, I see, Will be fixed

rules/consistent-test-it.js
ranyitz: I agree about that. Thanks

rules/consistent-test-it.js
ranyitz: 🙃

ranyitz:
@macklinu @SimenB Thanks for the review! 😃 
I made the fixes according to comments.

rules/consistent-test-it.js
```diff
@@ -0,0 +1,124 @@
...
+        additionalProperties: false,
+      },
+    ],
+  },
+  create(context) {
+    const configObj = context.options[0] || {};
+    const testKeyword = configObj.fn || 'test';
+    const testKeywordWithinDescribe =
+      configObj.withinDescribe || configObj.fn || 'it';
+
```
SimenB: This top part is missing some coverage, mind adding the necessary tests to cover all cases?

image

rules/__tests__/consistent-test-it.test.js
```diff
@@ -0,0 +1,352 @@
...
+);
+
+ruleTester.run(
+  'consistent-test-it default behavior',
+  rules['consistent-test-it'],
+  {
+    valid: [
+      {
+        code: 'test("foo")',
+        options: [{ fn: 'test', withinDescribe: 'it' }],
```
SimenB: I think removing `options` here is enough to get to 100% coverage 🙂

ranyitz: Yes, you are right :) will push fix soon

ranyitz:
@SimenB Thanks for the comment.
Now We've got 100% coverage here.
I also suggested to [add codecov integration](https://github.com/jest-community/eslint-plugin-jest/issues/69) which could probably help.

PR #67, opened 2018-02-11
fix(coverage): get close to 100% test coverage

SimenB:
It would be great if we could get to 100% coverage, especially as we're so close!

Missing a couple of branches in utils.js, but I've also added a couple of ignores I would like to drop. Help appreciated!

Missing coverage:
image

### DIFF
--- a/rules/no-disabled-tests.js
+++ b/rules/no-disabled-tests.js
@@ -69,6 +69,11 @@ module.exports = {
                 message: 'Call to pending() within test suite',
                 node,
               });
+            } else {
+              context.report({
+                message: 'Call to pending()',
+                node,
+              });
             }
             break;
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ script:
 - yarn commitlint --from="$TRAVIS_BRANCH" --to="$TRAVIS_COMMIT"
 - yarn commitlint --from=$TRAVIS_COMMIT
 - yarn prettylint
-- yarn test
+- yarn test --coverage
 after_script: greenkeeper-lockfile-upload
 jobs:
   include:
--- a/package.json
+++ b/package.json
@@ -51,6 +51,14 @@
     "*.{md,json}": ["prettier --write", "git add"]
   },
   "jest": {
+    "coverageThreshold": {
+      "global": {
+        "branches": 96,
+        "functions": 97,
+        "lines": 99,
+        "statements": 99
+      }
+    },
     "projects": [
       {
         "displayName": "test",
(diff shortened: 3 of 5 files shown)


rules/no-disabled-tests.js
```diff
@@ -1,19 +1,16 @@
 'use strict';
 
 function getName(node) {
-  function joinNames(a, b) {
```
SimenB: @borilla mind sharing why you added the check? None of the tests hit this path, so I just inlined it

rules/no-disabled-tests.js
```diff
@@ -1,19 +1,16 @@
...
       return node.name;
     case 'Literal':
       return node.value;
     case 'MemberExpression':
-      return joinNames(getName(node.object), getName(node.property));
+      return `${getName(node.object)}.${getName(node.property)}`;
   }
 
+  /* istanbul ignore next */
   return null;
```
SimenB: Is this just to be future proof?

rules/no-focused-tests.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
   create: context => ({
     CallExpression(node) {
       const callee = node.callee;
+      /* istanbul ignore if */
       if (!callee) {
```
SimenB: @jkimbo this is an old one, can you provide details on the cases where a `CallExpression` has no `callee`?

rules/valid-expect-in-promise.js
```diff
@@ -17,6 +17,7 @@ const isExpectCallPresentInFunction = body => {
     return body.body.find(line => {
       if (line.type === 'ExpressionStatement')
         return isExpectCall(line.expression);
+      /* istanbul ignore else */
       if (line.type === 'ReturnStatement') return isExpectCall(line.argument);
```
SimenB: @tushardhole can you provide some sample code where this condition is `false`?

SimenB:
@xfumihiro Do you think we could add some tests covering the uncovered parts of `utils.js`?

rules/no-disabled-tests.js
borilla: Just defensive coding, but I can come up with a theoretical example using an anonymous object:
```
({ f: function () {} }).f()
```

rules/no-disabled-tests.js
borilla: This is for any unmatched node-types (ie those we can't find names for), eg
```
(a || b).f(); // callee.object here is LogicalExpression
```

SimenB: Great, thanks! Added those 2 examples, and coverage hit 100% 🙂

borilla: :+1:

SimenB:
I'm gonna merge as is, and open up a separate issue for 100%

PR #74, opened 2018-02-12
feat(rule): add no-hooks rule

### LINKED ISSUE #72: New rule request: no-hooks
[eslint-plugin-mocha](https://github.com/lo1tuma/eslint-plugin-mocha) has a rule called [`no-hooks`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-hooks.md) that I think could be a good addition to this ESLint plugin. I think the [`mocha/no-hooks` docs](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-hooks.md) as well as [this tweet](https://twitter.com/thejameskyle/status/954906997169664000) help explain how using nested `describe()` and `beforeEach()` hook functions can result in harder to follow code / shared state between tests.

I think when combined with [`jest/consistent-test-it`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/consistent-test-it.md), you could enforce tests that look more like:

```js
function setupFoo(options) { /* ... */ }
function setupBar(options) { /* ... */ }
function setupBaz(options) { /* ... */ }

test('foo does this', () => {
  const foo = setupFoo({ /* options */ })
  expect(foo).toEqual(/* expected */)
})

test('bar does this', () => {
  const bar = setupBar({ /* options */ })
  expect(bar).toEqual(/* expected */)
})

test('baz does this', () => {
  const baz = setupBaz({ /* options */ })
  expect(baz).toEqual(/* expected */)
})
```

instead of looking like this:

```js
let foo
let bar
let baz

beforeEach(() => {
  foo = setupFoo()
  bar = setupBar()
  baz = setupBaz()
})

afterEach(() => {
  foo = null
  bar = null
  baz = null
})

test('foo does this', () => {
  expect(foo).toEqual(/* expected */)
})

test('bar does this', () => {
  expect(bar).toEqual(/* expected */)
})

test('baz does this', () => {
  expect(baz).toEqual(/* expected */)
})
```

I think there are times when using setup and teardown hooks are necessary, but I find that they can often be avoided. I could see this rule providing an option that would whitelist allowed hook function names, or the user could disable with `// eslint-disable-next-line jest/no-hooks`.

Thoughts on this rule? Would you accept a PR for this feature?

macklinu:
Resolves #72


rules/no-hooks.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const isHook = node => testHookNames[node.callee.name];
+    const isWhitelisted = node => allowed[node.callee.name];
+
+    return {
+      CallExpression(node) {
+        if (isHook(node) && !isWhitelisted(node)) {
+          context.report({
+            node,
+            message: 'Unexpected setup/teardown hook',
```
macklinu: Should the message including the hook name (e.g. beforeEach), or is this sufficient?

macklinu: I'm going to include the hook name in the error message. 👍

macklinu:
This is ready for review.

PR #78, opened 2018-02-13
fix(valid-expect): validate async expect calls

### LINKED ISSUE #54: Assert that async `expect`s are awaited or returned
`valid-expect` should check that if there's a `rejects` or `resolves` on an `expect`, that it's either `await`ed or `return`ed.

Example where we should warn:

```js
test('some test', () => {
  expect(Promise.resolve('hello')).resolves.toBe('hello');
});
```

macklinu:
Resolves #54 

I first tried working my way up the tree - to find a parent node of an `expect()` call that was a `ReturnStatement` or `AwaitStatement` and then finding the function that contained the `expect()` call, but I had trouble thinking through how to do that with the current `valid-expect` rule code. I ended up checking for a test case `CallExpression` and working my way down the tree - into the body of the callback function:

```js
test("foo", () => {
  // validate each `expect().resolves` and `expect().rejects` in this block
})
```

One downside is that this results in some really deeply nested property accessing (like `node.expression.callee.object.object.arguments[0].type`).

Am I approaching this the right way? Feedback is certainly welcome on how to improve this.

### DIFF
--- a/rules/__tests__/valid-expect.test.js
+++ b/rules/__tests__/valid-expect.test.js
@@ -173,5 +173,19 @@ ruleTester.run('valid-expect', rule, {
         { message: "Cannot use 'rejects' with an awaited expect expression" },
       ],
     },
+    {
+      code:
+        'test("foo", async () => { expect(await Promise.resolve(undefined)).resolves.not.toBeDefined(); });',
+      errors: [
+        { message: "Cannot use 'resolves' with an awaited expect expression" },
+      ],
+    },
+    {
+      code:
+        'test("foo", async () => { expect(await Promise.reject(undefined)).rejects.not.toBeDefined(); });',
+      errors: [
+        { message: "Cannot use 'rejects' with an awaited expect expression" },
+      ],
+    },
   ],
 });
--- a/rules/__tests__/valid-expect.test.js
+++ b/rules/__tests__/valid-expect.test.js
@@ -3,7 +3,11 @@
 const RuleTester = require('eslint').RuleTester;
 const rule = require('../valid-expect');
 
-const ruleTester = new RuleTester();
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 8,
+  },
+});
 
 ruleTester.run('valid-expect', rule, {
   valid: [
@@ -11,8 +15,12 @@ ruleTester.run('valid-expect', rule, {
     'expect(true).toBeDefined();',
     'expect([1, 2, 3]).toEqual([1, 2, 3]);',
     'expect(undefined).not.toBeDefined();',
-    'expect(Promise.resolve(2)).resolves.toBeDefined();',
-    'expect(Promise.reject(2)).rejects.toBeDefined();',
+    'test("foo", () => { return expect(Promise.resolve(2)).resolves.toBeDefined(); });',
+    'test("foo", async () => { await expect(Promise.reject(2)).rejects.toBeDefined(); });',
+    'test("foo", () => expect(Promise.resolve(2)).resolves.toBeDefined());',
+    'test("foo", async () => await expect(Promise.reject(2)).rejects.toBeDefined());',
+    'test("foo", async () => { expect(await Promise.resolve(2)).toBeDefined(); });',
+    'test("foo", async () => { expect(await Promise.reject(2)).toBeDefined(); });',
   ],
 
   invalid: [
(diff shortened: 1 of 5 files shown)


rules/valid-expect.js
```diff
@@ -5,8 +5,26 @@
...
+const util = require('./util');
+
 const expectProperties = ['not', 'resolves', 'rejects'];
 
+function isResolvesOrRejectsExpression(node) {
+  if (node.expression.type !== 'CallExpression') {
+    return false;
+  }
+  const callee = node.expression.callee;
+  return (
```
macklinu: This could be improved with something like [`_.get`](https://lodash.com/docs/4.17.5#get) or [`get-value`](https://www.npmjs.com/package/get-value).

rules/__tests__/valid-expect.test.js
```diff
@@ -131,5 +139,30 @@ ruleTester.run('valid-expect', rules['valid-expect'], {
...
+      errors: [{ message: "Must return or await 'expect.rejects' statement" }],
+    },
+    {
+      code:
+        'test("foo", async () => { expect(Promise.reject(2)).rejects.toBeDefined(); });',
+      errors: [{ message: "Must await 'expect.rejects' statement" }],
+    },
+    {
+      code:
+        'test("foo", async () => expect(Promise.reject(2)).rejects.toBeDefined());',
```
macklinu: This is technically valid, since this arrow function implicitly returns the `expect()` statement. However, the `test()` callback function is async, so I figured we could warn when you're using `async` without `await` in this case.

Thoughts on this?

SimenB: I think that should be fine as a separate rule.

https://github.com/avajs/eslint-plugin-ava/blob/master/docs/rules/no-async-fn-without-await.md

macklinu: Ok, I can remove that check from this rule then, and we can open an issue for another rule.

rules/__tests__/valid-expect.test.js
```diff
@@ -3,16 +3,24 @@
...
     'expect([1, 2, 3]).toEqual([1, 2, 3]);',
     'expect(undefined).not.toBeDefined();',
-    'expect(Promise.resolve(2)).resolves.toBeDefined();',
-    'expect(Promise.reject(2)).rejects.toBeDefined();',
+    'test("foo", () => { return expect(Promise.resolve(2)).resolves.toBeDefined(); });',
+    'test("foo", async () => { await expect(Promise.reject(2)).rejects.toBeDefined(); });',
+    'test("foo", () => expect(Promise.resolve(2)).resolves.toBeDefined());',
+    'test("foo", async () => await expect(Promise.reject(2)).rejects.toBeDefined());',
+    'test("foo", async () => { expect(await Promise.resolve(2)).resolves.toBeDefined(); });',
+    'test("foo", async () => { expect(await Promise.reject(2)).rejects.toBeDefined(); });',
```
macklinu: Based on the comments in https://github.com/jest-community/eslint-plugin-jest/issues/54#issuecomment-365531253, I might remove the logic surrounding these test cases, as it's not valid to await the Promise inside `expect()` and also use `resolves` or `rejects`.

SimenB: It'll _work_, it just doesn't make any sense. So I think we should warn in that case as you're doing currently

rules/valid-expect.js
SimenB: I'm down with that. Will make coverage a bit easier as well 🙂

macklinu: Do you have preferences over which dependency to add? `get-value` is smaller, although I don't know if that's as big a deal with ESLint plugins (for example, I would be more concerned about adding the entire Lodash dependency for a web app than a command line tool).

SimenB: no preference 🙂

SimenB:
conflicts again, sorry

rules/__tests__/valid-expect.test.js
SimenB: Actually, I was wrong, jest'll throw.

image

SimenB: so if anything it should be added to `invalid` 🙂

macklinu:
Thank you for the review, @SimenB - I'll fix the conflicts and make the necessary changes today.

rules/valid-expect.js
```diff
@@ -5,19 +5,72 @@
...
+            });
+          });
+      }
+    }
+
     return {
       CallExpression(node) {
+        if (util.isTestCase(node)) {
+          validateAsyncExpects(node);
+          return;
```
macklinu: If we are in a test case function, there is no need to proceed, since the `node.callee.name` will never be `expect`.

rules/__tests__/valid-expect.test.js
macklinu: Created #81 so this wouldn't get lost

rules/valid-expect.js
```diff
@@ -5,19 +5,72 @@
...
+            );
+            const propertyName = get(
+              node,
+              'expression.callee.object.property.name'
+            );
+
+            return (
+              node.expression.type === 'CallExpression' &&
+              objectName === 'expect' &&
+              (propertyName === 'resolves' || propertyName === 'rejects')
```
SimenB: I wonder if we can reuse the logic somehow with the other `resolves`/`rejects` checks

rules/__tests__/valid-expect.test.js
```diff
@@ -131,5 +139,39 @@ ruleTester.run('valid-expect', rules['valid-expect'], {
         },
       ],
     },
+    {
+      code:
+        'test("foo", () => { expect(Promise.resolve(2)).resolves.toBeDefined(); });',
+      errors: [{ message: "Must return or await 'expect.resolves' statement" }],
```
SimenB: can you assert on the locations as well?

macklinu: Yep, will make those changes 👍

rules/valid-expect.js
macklinu: Good idea. I'll look into that while resolving the failing test case you reported below.

SimenB:
I don't get emails when people push to PRs I'm reviewing for some reason 🙁 So please ping me when this is ready

macklinu:
Will do. I need to sit down this week and give this PR a solid look to see
how we can share logic between the current behavior and the new behavior
I’m introducing.

macklinu:
@SimenB I'm in-between jobs and moving right now and probably won't be available to contribute again until late March/early April. Thoughts on what to do with this PR? Just leave it open for now?

macklinu:
I'm not sure this is the right approach, and I haven't been able to take a look for quite some time. Closing for now.

PR #79, opened 2018-02-15
docs: Use link to tagged version for rule docs

sudo-suhas:
This adds a utility function and uses the version present in package.json for generating the URL to tagged version of the rule documentation. I think this is good because the documentation will match the version the user has installed. This way, even if a rule is changed or removed in the future, the user can find the documentation with ease.

Note: The rule `eslint-plugin/require-meta-docs-url` cannot be used with a function which would return the URL(see [not-an-aardvark/eslint-plugin-eslint-plugin#58](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/58)) so I have disabled it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,13 +18,6 @@ module.exports = {
   rules: {
     eqeqeq: ['error', 'smart'],
     strict: 'error',
-    'eslint-plugin/require-meta-docs-url': [
-      'error',
-      {
-        pattern:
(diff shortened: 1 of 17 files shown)


SimenB:
@sudo-suhas I added some comments to the merge commit instead of in this PR... Mind taking a look?

e0160e59e939e633282443f2fe5768ef09da2c48

PR #80, opened 2018-02-16
docs: Improve docs URL resolution

sudo-suhas:
- Use `__filename` to determine the rule name and build the docs link. This also fixes an inadvertent bug where the URL ended with `.md.md`(ex: `https://github.com/jest-community/eslint-plugin-jest/blob/v0.0.0-development/docs/rules/valid-expect.md.md` for `valid-expect`).
- Check the docs file exists and throw an error otherwise.


SimenB:
Thanks for following up!

PR #82, opened 2018-02-16
chore: add jest-runner-eslint

macklinu:
I've been learning more about Jest runners this week and am opening this PR to see if you are 👍 / 👎  on adding [jest-runner-eslint](https://github.com/jest-community/jest-runner-eslint) as a Jest runner for this project.

Sample output of `yarn test`:

image

Running `yarn test` takes slightly longer, but it comes with the added benefit of being able to see lint and test failures all from within the Jest UI.

Another thing I noticed is that our tests look like this:

```js
const RuleTester = require('eslint').RuleTester;
const rules = require('../..').rules;

const ruleTester = new RuleTester();

ruleTester.run('valid-expect', rules['valid-expect'], {
  valid: [/* ... */],
  invalid: [/* ... */],
});
```

Since we require `rules` from `index.js` and access the rule itself by calling `rule['rule-name']`, any time I change one rule, Jest thinks all the rules have changed and runs all the rule test files. Refactoring to just require the rule under test would make Jest's watch mode more what I expect - I change `valid-expect.js` and Jest will only test/lint the `valid-expect.test.js` file.

This is sort of like an issue-meets-PR. 😄 Let me know what you think about these things, if you'd like to address any of these things here or in future issues, and I'd be happy to help make those things happen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ before_script: greenkeeper-lockfile-update
 script:
 - yarn commitlint --from="$TRAVIS_BRANCH" --to="$TRAVIS_COMMIT"
 - yarn commitlint --from=$TRAVIS_COMMIT
-- yarn lint
 - yarn prettylint
 - yarn test
 after_script: greenkeeper-lockfile-upload
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "eslint-plugin-prettier": "^2.3.1",
     "husky": "^0.14.3",
     "jest": "^22.0.4",
+    "jest-runner-eslint": "^0.4.0",
     "lint-staged": "^6.0.0",
     "prettier": "^1.10.2",
     "prettylint": "^1.0.0",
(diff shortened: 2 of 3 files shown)


macklinu:
I'm not sure what is causing the Node 4 test failures. 😅

ranyitz:
@macklinu, it seems that the error comes from `jest-environment-jsdom`. I think that you should configure the environment for the "test" project.

macklinu:
@ranyitz thanks for that - it helped me find the issue! I'll amend the commit and push up. 👍

ranyitz:
@macklinu Happy to help! ☺️

SimenB:
>Another thing I noticed is that our tests look like this:

PR welcome! Sounds like a good change.

.travis.yml
```diff
@@ -13,7 +13,6 @@ before_script: greenkeeper-lockfile-update
 script:
 - yarn commitlint --from="$TRAVIS_BRANCH" --to="$TRAVIS_COMMIT"
 - yarn commitlint --from=$TRAVIS_COMMIT
-- yarn lint
 - yarn prettylint
```
SimenB: A jest runner for prettylint would be awesome 🙂

Issue #86, opened 2018-02-17
Get to 100% test coverage


SimenB:
In #67 thresholds were added, but I was unable to set them at 100 as we don't have coverage. Help reaching it would be greatly appreciated!

SimenB:
With #88 we're getting really close! Only a couple of branches left

tushardhole:
@SimenB I will look into missing branch coverage for valid-expect-in-promise.js

tushardhole:
@SimenB I will also check remaining files for 100% coverage.

tushardhole:
Created PR [here](https://github.com/jest-community/eslint-plugin-jest/pull/89)

SimenB:
Woo! ❤️

PR #88, opened 2018-02-21
Get full test coverage

ranyitz:
Regarding #86, this should get us to 100% coverage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,10 +53,10 @@
   "jest": {
     "coverageThreshold": {
       "global": {
-        "branches": 96,
-        "functions": 97,
-        "lines": 99,
-        "statements": 99
+        "branches": 98,
+        "functions": 100,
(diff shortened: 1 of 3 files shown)


ranyitz:
image

The untested branches are from [this function](https://github.com/jest-community/eslint-plugin-jest/blob/master/rules/util.js#L24) and [this function](https://github.com/jest-community/eslint-plugin-jest/blob/master/rules/util.js#L29) which are not tested and I think should be `resolves` and `rejects` instead of `resolve` and `reject`.

The other two are `else path not taken` which is actually a case when we're not entering an if statement.

Should be fixed? Maybe in a different PR?

SimenB:
Thanks you! ❤️ Mind bumping the coverage thresholds in package.json?

ranyitz:
With pleasure :smile:

PR #89, opened 2018-02-22
Coverage

tushardhole:
* Added missing branch coverage for valid-expect-in-promise.js & utils.js
* Added a scenario in valid describe where second argument is not a function(there was missing coverage for same)


SimenB:
This is awesome, thank you so much!

Issue #76, opened 2018-02-13
Addition: lowercase-name only for tests


nevir:
It'd be helpful to allow uppercase names for describes (e.g. subjects - `ClassName`, etc), while still enforcing lowercase test titles

macklinu:
We could provide an option that allows for ignoring the check for `describe()`.

```json
{
  "jest/lowercase-name": ["error", {
    "ignore": ["describe"]
  }]
}
```

I'm also wondering if `lowercase-name` should check `describe()` at all, or if only checking `it()` and `test()` is more valuable. 🤔

nevir:
There's also advice like "don't start test descriptions with 'should'" and
similar things.

Idea: what if there's an option to provide a regex (or multiple) that can
be used to validate tests (and a similar option for describes)?

SimenB:
I like the idea of a regex (and an `ignore` array). PR welcome for both!

macklinu:
This code introduced in #74 could be a starting point for introducing an `ignore` option to the `lowercase-name` rule.

https://github.com/jest-community/eslint-plugin-jest/blob/47eb6c27c45f5b9207c37a18dd5d40d46d0943d8/rules/no-hooks.js#L30-L35

lionize:
I submitted a PR to add the `ignore` option for this rule. I started to do some work on regex but it doesn't really feel like the description message regex belongs in the `lowercase-name` rule. Thoughts?

nevir:
maybe rename the rule to `name`, `naming-style` or something along those lines?

SimenB:
Renaming it is a breaking change... We have a few breaking changes lined up anyways, so not a blocker. Worth it, though?

nevir:
Unsure!  could maybe duplicate it into the new name & deprecate old, if you want a smoother transition (but a bit more annoying to maintain)

SimenB:
This should be covered by the new option, agreed @nevir?

PR #92, opened 2018-02-27
feat(lowercase-name): add 'ignore' option (#76)

lionize:
This PR adds the `ignore` option discussed in #76.

### DIFF
--- a/rules/lowercase-name.js
+++ b/rules/lowercase-name.js
@@ -54,11 +54,20 @@ module.exports = {
     },
   },
   create(context) {
+    const ignore = (context.options[0] && context.options[0].ignore) || [];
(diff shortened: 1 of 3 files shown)


rules/lowercase-name.js
```diff
@@ -54,11 +54,21 @@ module.exports = {
     },
   },
   create(context) {
+    const ignoredFunctionNames = (
+      context.options[0] || { ignore: [] }
```
SimenB: what happens here if `context.options[0] = {}`?

lionize: @SimenB fixed

lionize:
Also added a check for empty argument list which could throw an error.

mchandler-cng:
@SimenB fixed

mchandler-cng:
@SimenB 👍

PR #95, opened 2018-03-02
feat: adds no jest import rule

### LINKED ISSUE #90: Add rule disallowing importing `jest`
See https://github.com/facebook/jest/issues/4613, this is apparently a footgun.

We might take it one step further an yell at all assignments to a `jest` object as it might mess up static analysis, thus negating hoisting of `jest.mock`?

But at least a rule for not importing `jest` (as it doesn't export anything) makes sense.

brianlmacdonald:
Closes #90.

Adds the eslint rule 'no-jest-import', disallowing any import from Jest.

Adds tests and documentation for rule.

### DIFF
--- a/docs/rules/no-jest-import.md
+++ b/docs/rules/no-jest-import.md
@@ -1,0 +1,20 @@
+# Disallow importing Jest(no-jest-import)
+
+The jest object is automatically in scope within every test file. The methods in
+the jest object help create mocks and let you control Jest's overall behavior.
+It is therefore completely unnecessary to import in Jest, as Jest doesn't export
+anything in the first place.
+
+### Rule details
+
+This rule reports on any importing of Jest.
+
+To name a few: *var jest = require('jest'); *const jest = require('jest');
+*import jest from 'jest'; *import {jest as test} from 'jest';
+
+There is no correct usage of this code, other than to not import Jest it in the
+first place.
+
+## Further Reading
+
+\*[The Jest Object](https://facebook.github.io/jest/docs/en/jest-object.html)
--- a/docs/rules/no-jest-import.md
+++ b/docs/rules/no-jest-import.md
@@ -1,18 +1,18 @@
 # Disallow importing Jest(no-jest-import)
 
-The jest object is automatically in scope within every test file. The methods in
-the jest object help create mocks and let you control Jest's overall behavior.
-It is therefore completely unnecessary to import in Jest, as Jest doesn't export
-anything in the first place.
+The `jest` object is automatically in scope within every test file. The methods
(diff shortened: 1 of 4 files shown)


docs/rules/no-jest-import.md
```diff
@@ -0,0 +1,20 @@
+# Disallow importing Jest(no-jest-import)
+
+The jest object is automatically in scope within every test file. The methods in
```
ranyitz: Use a code block for the `jest` object.

docs/rules/no-jest-import.md
```diff
@@ -0,0 +1,20 @@
...
+the jest object help create mocks and let you control Jest's overall behavior.
+It is therefore completely unnecessary to import in Jest, as Jest doesn't export
```
ranyitz: I would change Jest to `jest` here, as you refer to the object.

docs/rules/no-jest-import.md
```diff
@@ -0,0 +1,20 @@
...
+anything in the first place.
+
+### Rule details
+
+This rule reports on any importing of Jest.
+
+To name a few: *var jest = require('jest'); *const jest = require('jest');
```
ranyitz: Maybe use code blocks here?

* `var jest = require('jest');`
* `const jest = require('jest');`
* `import jest from 'jest';`
* `import {jest as test} from 'jest';`

docs/rules/no-jest-import.md
```diff
@@ -0,0 +1,20 @@
...
+*import jest from 'jest'; *import {jest as test} from 'jest';
+
+There is no correct usage of this code, other than to not import Jest it in the
```
ranyitz: Jest => `jest`

rules/no-jest-import.js
```diff
@@ -0,0 +1,39 @@
...
+      url: getDocsUrl(__filename),
+    },
+  },
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        if (node.source.value === 'jest') {
+          context.report({
+            node,
+            message: `Jest is automatically in scope. Do not import Jest, as it doesn't export anything.`,
```
ranyitz: The message can be moved into a variable so you won't write it twice.

index.js
```diff
@@ -29,6 +30,7 @@ module.exports = {
         'jest/no-disabled-tests': 'warn',
         'jest/no-focused-tests': 'error',
         'jest/no-identical-title': 'error',
+        'jest/no-jest-import': 'error',
```
SimenB: technically breaking to add it here, mind removing it?

SimenB:
@ranyitz thanks for reviewing! 😀

Issue #98, opened 2018-03-09
Missing `no-jest-import` from the Rules table in README.md


alexilyaev:
Maybe even make it recommended.

ranyitz:
@alexilyaev Yes, seems that it does missing, I've opened a PR that fixes that #99.

Regarding it being `recommended` and the error level, I think it should be `recommended` with error level `warn`, what do you think @SimenB ?

PR #99, opened 2018-03-10
No jest import fixes

ranyitz:
Following to #98, This PR:

* Add `no-jest-import` rule to the `README.md`.
* Make it `recommended`.
* Set error level to `warn`.


SimenB:
Technically it's a breaking change adding it to `recommended`, but if people _do_ import `jest` then it doesn't work, so it might be OK.

@thymikee thoughts?

thymikee:
I'd rather say it's a fix, being breaking at the same time. I'm OK with adding it to the next minor/patch

SimenB:
🤞

PR #106, opened 2018-04-18
fix(valid-describe): add check for concise body arrow functions

### LINKED ISSUE #105: Bug with valid-describe
I found this bug when converting some tests from another repo.

Sample test:
```js
describe('test name', () =>
  test('some test', () => {...})
);
```
Will cause the following error:
```
/eslint-plugin-jest/rules/valid-describe.js:79
          callbackFunction.body.body.forEach(node => {
                                    ^

TypeError: Cannot read property 'forEach' of undefined
    at CallExpression (/eslint-plugin-jest/rules/valid-describe.js:79:37)
    at listeners.(anonymous function).forEach.listener (/eslint/lib/util/safe-emitter.js:47:58)
    at Array.forEach (native)
    at Object.emit (/eslint/lib/util/safe-emitter.js:47:38)
    at NodeEventGenerator.applySelector (/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (/eslint/lib/code-path-analysis/code-path-analyzer.js:606:23)
    at Traverser.enter (/eslint/lib/linter.js:955:32)
    at Traverser.__execute (/estraverse/estraverse.js:397:31)
    at Traverser.traverse (/node_modules/estraverse/estraverse.js:501:28)
```

```
"eslint": "4.9.0",
"eslint-config-airbnb": "^16.1.0",
"eslint-import-resolver-babel-module": "^4.0.0",
"eslint-plugin-import": "^2.11.0",
"eslint-plugin-jest": "^21.15.0",
"eslint-plugin-jsx-a11y": "^6.0.3",
"eslint-plugin-promise": "^3.7.0",
"eslint-plugin-react": "^7.7.0"
```

Saberre:
Fixes #105.

I've added a new condition for an arrow function with concise body. I'm not good at writing English, so the report message may not be look good to you. If so, please let me know and I'll fix it!

### DIFF
--- a/rules/__tests__/valid-describe.test.js
+++ b/rules/__tests__/valid-describe.test.js
@@ -122,6 +122,20 @@ ruleTester.run('valid-describe', rule, {
     },
     {
       code: `
+      describe('foo', () =>
+        test('bar', () => {})
+      )
+      `,
+      errors: [
+        {
+          message: 'Arrow function describe callback must have a block body',
+          line: 2,
+          column: 23,
+        },
+      ],
+    },
+    {
+      code: `
       describe('foo', function () {
         return Promise.resolve().then(() => {
           it('breaks', () => {
--- a/rules/__tests__/valid-describe.test.js
+++ b/rules/__tests__/valid-describe.test.js
@@ -34,6 +34,11 @@ ruleTester.run('valid-describe', rule, {
       })
     })
     `,
+    `
+    describe('foo', () =>
+      test('bar', () => {})
+    )
+    `,
   ],
   invalid: [
     {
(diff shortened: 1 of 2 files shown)


SimenB:
I wonder if we should allow it? I don't think it hurts.

@thymikee thoughts?

(CI seems confused, I'm not sure what's up with it...)

macklinu:
I think we should allow this `ArrowFunctionExpression` syntax. I don't normally see tests written with a body-less arrow function, which is mentioned in #105, but I think that's an edge case we should check and allow. If the user wants to only include one `test()` or `it()` function per `describe()`, that's their choice. Also, it's technically valid to write tests this way in Jest (if I'm not mistaken), so I don't think the `valid-describe` rule should prevent something that is technically valid. That's my opinion, but definitely open to hearing others' thoughts!

Saberre:
I first thought that a concise body arrow function violates the existing lint rule because it implicitly returns the value of its body. But I also think that it doesn't harm the describe block by returning a value. Please let me know which one is better and I'll reflect it to the code!

SimenB:
Brace-less arrows should be supported even though returning in a `describe` is weird in the normal case

SimenB:
Woah, I never got an email that you fixed this @Saberre. So sorry! 🙏

Issue #107, opened 2018-04-19
`no-identical-title` maybe could be ignored for nested `describe` blocks...?


bunnymatic:
My scenario is that the linter is whining about `no-identical-title`s for nested describes where that nested title is actually what i want.

Imagine this:
```javascript
describe("my component", () => {
  describe("west facing", () = {
    describe("when the sun is up", () => {  
       it("sees the sun", () => {});
    })
    describe("when the sun is down", () => {  
       it("sees the moon", () => {});
    });
  });
  describe("east facing", () = {
    describe("when the sun is up", () => {  
       it("does not see the sun", () => {});
    })
    describe("when the sun is down", () => {  
       it("sees whatever", () => {});
    });
  });
});
```

In this case... the linter doesn't like `when the sun is up` and `when the sun is down` because they are the same, but as nested describes, they seem like a valid use case.

What do you think of that?  could there be a rule that looks at the full "nested" title to determine it's allowed "uniqueness"?

SimenB:
That definitely makes sense! We should try to match against full names (so `my component > west facing > when the sun is up > sees the sun`) and not against `describe`s and `it`s in isolation.

Not sure how hard that is, but PR welcome!

ranyitz:
> It will report when two test suites or two test cases **at the same level** of a test suite have the same title.

@bunnymatic @SimenB The linter would not raise an error on the provided code, because `when the sun is up` & `when the sun is down` are nested under different describe blocks.

The message can be confusing though,

https://github.com/jest-community/eslint-plugin-jest/blob/03a7cdab01fac29550345bb952e1fb5ca70f05e5/rules/no-identical-title.js#L30

I would change it to be `describe block title is used multiple times in the same describe block` so it'll be clearer.

bunnymatic:
You are totally right.  I had some deep nesting in my actual code which made me think this was the issue.  But as you point out, it's because we have 2 blocks in the same describe with the same sub title.  Sorry to bug you.

Definitely the messaging might be a little better to indicate describe titles vs test titles.  But it is working as expected.

Sorry to bother you.  And thanks for the quick reply.

SimenB:
Ah, ok. Cool!

@ranyitz PR improving the message very much welcome 🙂

ranyitz:
@SimenB gladly #108 😄

bunnymatic:
I'll see what i can do

SimenB:
[...]

PR #108, opened 2018-04-19
fix(rules): improve no-identical-title error message

ranyitz:
This is a PR to improve the error message of `no-identical-title` rule, after the issue #107


SimenB:
Thanks!

PR #122, opened 2018-05-28
docs: clean up rules table links

macklinu:
This PR converts the README's Rules table from inline- to reference-style links, which narrows the width of the table and (in my opinion) makes it easier to read and modify. Thoughts on this change?


SimenB:
Yeah, way better. Thanks!

PR #126, opened 2018-07-09
Prepare for update to ESLint v5

macklinu:
This PR targets the `next` branch as discussed in https://github.com/jest-community/eslint-plugin-jest/pull/125#issuecomment-403375120 and takes care of some initial work for the next major version of eslint-plugin-jest.

This PR:

* updates the dev and peer dependency to ESLint >=5
* updates the supported Node engine to v6 (in both package.json and .travis.yml)
* refactors to use object destructuring when requiring functions from './util.js' and 'eslint'

Each commit could be applied as-is to the `next` branch using GitHub's **Rebase and merge** feature, but I will let you determine what the proper merging strategy is for this PR into the `next` branch. 👍 

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ node_js:
 - 8
 - node
 - 6
-- 4
 before_install:
 - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.3.2
 - export PATH="$HOME/.yarn/bin:$PATH"
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "index.js"
   ],
   "engines": {
-    "node": ">= 4"
+    "node": ">=6"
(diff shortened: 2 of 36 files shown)


SimenB:
Once this lands we can land all the other greenkeeper PRs as well - they all drop node 4 as their only breaking change

SimenB:
@macklinu ping 🙂

macklinu:
@SimenB I've updated the commit messages. Let me know if I can change any wording to be clearer or more appropriate for the breaking changes.

SimenB:
Thanks! I just pushed a commit updating all dependencies as well now that we've dropped node 4.

Not sure what else we want to land here. We definitely want to use `messageId` and take a look at the recommended rules. Anything else we want to do?

PR #129, opened 2018-07-16
feat(rules): add prefer-inline-snapshots

tryggvigy:
## This PR will 

Implement the `prefer-inline-snapshots` rule. This rule serves to encourage interspersing snapshots inline in test file code over writing them to a separate `_snapshots_/*.snap` file.

## Why?
The believe is that the snapshot test is easier to comprehend when it can be seen as a whole. This is especially convenient during code-reviews.


tryggvigy:
I'll amend to last commit asap to fix the readme.

SimenB:
This will probably fail the publish seeing as npm has revoked all tokens and I haven't updated this... Will do later today

tryggvigy:
Awesome, thank you!

PR #132, opened 2018-07-18
fix: no-large-snapshots to allow maxSize of 0

paularmstrong:
**Problem:** Configuring the rule `jest/no-large-snapshots` with `[{ maxSize: 0 }]` sets the line limit to `50` instead of `0`.

**Solution:** Treat `0` as a requirement for no snapshot.

### DIFF
--- a/rules/no-large-snapshots.js
+++ b/rules/no-large-snapshots.js
@@ -11,7 +11,9 @@ module.exports = {
   create(context) {
     if (context.getFilename().endsWith('.snap')) {
       const lineLimit =
-        (context.options[0] && context.options[0].maxSize) || 50;
+        context.options[0] && Number.isFinite(context.options[0].maxSize)
+          ? context.options[0].maxSize
+          : 50;
 
       return {
         ExpressionStatement(node) {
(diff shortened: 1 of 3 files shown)


rules/no-large-snapshots.js
```diff
@@ -11,7 +11,11 @@ module.exports = {
   create(context) {
     if (context.getFilename().endsWith('.snap')) {
       const lineLimit =
-        (context.options[0] && context.options[0].maxSize) || 50;
+        context.options[0] &&
+        typeof context.options[0].maxSize === 'number' &&
+        isFinite(context.options[0].maxSize)
```
SimenB: Could we do `Number.isFinite`? Then you can drop the `typeof` check 🙂 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite#Description

rules/no-large-snapshots.js
```diff
@@ -22,7 +26,9 @@ module.exports = {
           if (lineCount > lineLimit) {
             context.report({
               message:
-                'Expected Jest snapshot to be smaller than {{ lineLimit }} lines but was {{ lineCount }} lines long',
+                lineLimit === 0
+                  ? 'Expected to not encounter a Jest snapshot but was found with {{ lineCount }} lines long'
```
SimenB: nice one!

rules/no-large-snapshots.js
paularmstrong: Sure thing!

paularmstrong:
Thanks @SimenB !

PR #134, opened 2018-07-29
feat(rules): add prefer-strict-equal

tryggvigy:
## This PR will 

Implement the `prefer-strict-equal` rule. This rule serves to encourage stricter equality checks on objects.

## Why?
The believe is that imposing stricter equality results in safer tests.

### DIFF
--- a/docs/rules/prefer-strict-equal.md
+++ b/docs/rules/prefer-strict-equal.md
@@ -1,0 +1,25 @@
+# Suggest using `toStrictEqual()` ` (prefer-strict-equal)
+
+`toStrictEqual` not only checks that two objects contain the same data but also
+that they have the same shape. The believe is that imposing a stricter equality
+results in safer tests.
+
+## Rule details
+
+This rule triggers a warning if `toEqual()` is used to assert equality.
+
+This rule is enabled by default.
+
+### Default configuration
+
+The following pattern is considered warning:
+
+```js
+expect({ a: 'a', b: undefined }).toEqual({ a: 'a' }); // true
+```
+
+The following pattern is not warning:
+
+```js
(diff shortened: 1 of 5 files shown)


tryggvigy:
The desire to make strict equality the default behaviour [has been discussed](https://github.com/facebook/jest/pull/6032#issuecomment-382698991). This rule at least makes that enforceable in codebases. Maybe this should be a part of the `recommended` rules?

docs/rules/prefer-strict-equal.md
```diff
@@ -0,0 +1,25 @@
+# Suggest using `toStrictEqual()` (prefer-strict-equal)
+
+`toStrictEqual` not only checks that two objects contain the same data but also
+that they have the same shape. The believe is that imposing a stricter equality
```
SimenB: "The believe is that" sounds odd. Could you re-phrase?

docs/rules/prefer-strict-equal.md
```diff
@@ -0,0 +1,25 @@
...
+results in safer tests.
+
+## Rule details
+
+This rule triggers a warning if `toEqual()` is used to assert equality.
+
+This rule is enabled by default.
```
SimenB: no it's not 🙂

tryggvigy: Thanks, addressed in a7ca3b2

docs/rules/prefer-strict-equal.md
tryggvigy: I tweaked and expanded on this a bit in a7ca3b2

SimenB:
Thanks!

PR #136, opened 2018-08-09
feat(rules): add `no-test-return-statement`

mattphillips:
Adds a rule for not returning inside of a test. Related issue in Jest: https://github.com/facebook/jest/issues/6516

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ for more information about extending configuration files.
 | [no-jest-import][]           | Disallow importing `jest`                                         | ![recommended][] |                     |
 | [no-large-snapshots][]       | Disallow large snapshots                                          |                  |                     |
 | [no-test-prefixes][]         | Disallow using `f` & `x` prefixes to define focused/skipped tests |                  | ![fixable-green][]  |
+| [no-test-return-statement][] | Disallow explicitly returning from tests                          |                  |                     |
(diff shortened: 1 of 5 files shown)


SimenB:
What about `test('hello', () => somePromise);`?

SimenB:
If that should be allowed, let's call it `no-test-return-statement` or something

README.md
```diff
@@ -90,6 +90,7 @@ for more information about extending configuration files.
 | [no-jest-import][]           | Disallow importing `jest`                                         | ![recommended][] |                     |
 | [no-large-snapshots][]       | Disallow large snapshots                                          |                  |                     |
 | [no-test-prefixes][]         | Disallow using `f` & `x` prefixes to define focused/skipped tests |                  | ![fixable-green][]  |
+| [no-test-return-statement][] | Disallow returning from tests                                     |                  |                     |
```
SimenB: "Disallow explicitly returning from tests"?

docs/rules/no-test-return-statement.md
```diff
@@ -0,0 +1,47 @@
+# Disallow returning from tests (no-test-return-statement)
+
+Tests in Jest are void and cannot return values.
```
SimenB: "should be"? Seeing as you can return a promise, and that's unlikely to change (I don't think it's possible to differentiate between an `async` function and a promise-returning one (at least not if it's transpiled))

PR #155, opened 2018-09-21
fix(no-disabled-tests): fix false positives for pending() usage

### LINKED ISSUE #149: a function named pending () triggered the 'no-disabled-tests' error
I just read that pending is a jasmine global that allows you not run the test. 
effectively disabling it. 

Our code imports a value and then runs its it. 
It appears the parser is confusing it with the jasmine global.
without being to careful with the syntax, This is basically what we're doing:
```js

import { pending } from './myActions' ; 

describe('the pending action creator should return correct object', ()=>{
    it('should work correctly' , ()=>{
      expect(pending()).toStrictEqual({type:'FETCH_ALL_ALBUMS_PENDING'});
   })
})
```

does this rule notice if you import the function before assuming you're using the jasmine global and hence, disabling the test?  
or is it an issue with the code?

macklinu:
Resolves #149 

This PR adds some failing test cases and fixes for false positive warnings of `pending()` in the no-disabled-tests rule.

I'll add some comments inline to help explain my solution - feedback welcome! 😺

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,7 @@ module.exports = {
   },
   env: {
     node: true,
+    es6: true,
   },
   rules: {
     eqeqeq: ['error', 'smart'],
--- a/rules/__tests__/no-disabled-tests.test.js
+++ b/rules/__tests__/no-disabled-tests.test.js
@@ -3,7 +3,11 @@
 const RuleTester = require('eslint').RuleTester;
 const rule = require('../no-disabled-tests');
 
-const ruleTester = new RuleTester();
+const ruleTester = new RuleTester({
+  parserOptions: {
+    sourceType: 'module',
+  },
+});
 
 ruleTester.run('no-disabled-tests', rule, {
   valid: [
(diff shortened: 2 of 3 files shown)


.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
   },
   env: {
     node: true,
+    es6: true,
```
macklinu: I used a `Set()` in my solution, and this was needed for ESLint to understand the `Set` global.

rules/__tests__/no-disabled-tests.test.js
```diff
@@ -3,7 +3,11 @@
 const RuleTester = require('eslint').RuleTester;
 const rule = require('../no-disabled-tests');
 
-const ruleTester = new RuleTester();
+const ruleTester = new RuleTester({
+  parserOptions: {
+    sourceType: 'module',
```
macklinu: This allows for writing examples that use ES modules (i.e. `import` statements).

rules/__tests__/no-disabled-tests.test.js
```diff
@@ -17,6 +21,35 @@ ruleTester.run('no-disabled-tests', rule, {
...
+      '  const pending = getPending()',
+      '  expect(pending()).toEqual({})',
+      '})',
+    ].join('\n'),
+    [
+      'test("foo", () => {',
+      '  expect(pending()).toEqual({})',
+      '})',
+      '',
+      'function pending() {',
```
macklinu: We want to ensure that a local function called `pending()` will not trigger a no-disabled-tests violation.

rules/no-disabled-tests.js
```diff
@@ -19,6 +19,33 @@ function getName(node) {
   return null;
 }
 
+function collectReferences(scope) {
```
macklinu: This function takes the current scope (so wherever the `pending()` function is called) and walks up the tree to collect all local references and [unresolved](https://eslint.org/docs/developer-guide/scope-manager-interface#through) references. This information is helpful for determining what local variables or functions have been defined in a file, as well as what variables that ESLint could not resolve (which likely means those are globals, like `describe()` & `test()`).

rules/no-disabled-tests.js
```diff
@@ -58,7 +85,19 @@ module.exports = {
...
+            const references = collectReferences(context.getScope());
+
+            if (
+              // `pending` was found as a local variable or function declaration.
+              references.locals.has('pending') ||
+              // `pending` was not found as an unresolved reference,
+              // meaning it is likely not an implicit global reference.
+              !references.unresolved.has('pending')
+            ) {
+              break;
```
macklinu: Let me know if this if statement is clear enough with the comment or if you see a clearer way to communicate this - I want this change to be as understandable and maintainable as possible. ❤️

PR #158, opened 2018-09-25
refactor(no-jest-import): use ESQuery syntax for selectors

macklinu:
ESLint visitor selectors support [ESQuery](https://github.com/estools/esquery) syntax, which is similar to a CSS style selector pattern. I'm opening this PR partly to show that it's capable to move some logic into these selectors (like `[callee.name="require"]`), but also to get some feedback if there is interest in introducing these into the codebase. Learned about them this weekend and wanted to share in the context of an ESLint plugin. 🤓 No worries on my end if this is merged or closed, more just want to start a discussion.

### DIFF
--- a/rules/no-jest-import.js
+++ b/rules/no-jest-import.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const getDocsUrl = require('./util').getDocsUrl;
-const getNodeName = require('./util').getNodeName;
+
 const message = `Jest is automatically in scope. Do not import "jest", as Jest doesn't export anything.`;
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


SimenB:
I like it! @thymikee thoughts?

thymikee:
Strings are not statically analyzable, but we're not using any typings here, so that shouldn't be an issue. Is this some stable syntax and won't go away quickly? If it's good, then I'm cool :)

macklinu:
The documentation is hidden somewhat on the ESLint site, but there is a guide for selectors and how to integrate them into your rules: https://eslint.org/docs/developer-guide/selectors

One public ESLint rule that utilizes this syntax is `no-restricted-syntax`, which allows user to pass these ESQuery selector strings in their ESLint configuration. See here: https://eslint.org/docs/developer-guide/selectors#restricting-syntax-with-selectors

TL;DR I don't think this is going away soon :)

rules/no-jest-import.js
```diff
@@ -12,32 +12,20 @@ module.exports = {
...
-          context.report({
-            loc: {
-              end: {
-                column: node.arguments[0].loc.end.column,
-                line: node.arguments[0].loc.end.line,
-              },
-              start: node.arguments[0].loc.start,
+      'CallExpression[callee.name="require"][arguments.0.value="jest"]'(node) {
+        context.report({
+          loc: {
```
SimenB: can we do `loc: node.arguments[0].loc` instead of the whole thing?

macklinu: Good call, will push up a commit.

SimenB:
This is awesome! Would love to see us migrate over.

PR #160, opened 2018-09-29
Add new rule: require-tothrow-message

### LINKED ISSUE #154: Ensure toThrow always has a message
Consider adding a new rule, which checks that `toThrow` always has one argument, to match the error's message. 

This would have helped me a few times, when I used `toThrow` without any arguments, and it caught a general `TypeError`, etc. when instead I wanted it to catch a specific custom error.

garyking:
This closes #154.

This rule checks if `toThrow` or `toThrowError` has a message.

### DIFF
--- a/rules/require-tothrow-message.js
+++ b/rules/require-tothrow-message.js
@@ -15,7 +15,7 @@ module.exports = {
 
         // Look for `toThrow` calls with no arguments.
         if (
-          ['toThrow', 'toThrowError'].includes(propertyName) &&
+          ['toThrow', 'toThrowError'].indexOf(propertyName) > -1 &&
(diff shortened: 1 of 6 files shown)


rules/require-tothrow-message.js
```diff
@@ -0,0 +1,32 @@
...
+    },
+  },
+  create(context) {
+    return {
+      CallExpression(node) {
+        const propertyName = node.callee.property && node.callee.property.name;
+
+        // Look for `toThrow` calls with no arguments.
+        if (
+          ['toThrow', 'toThrowError'].indexOf(propertyName) > -1 &&
```
SimenB: can use `includes` whenever we merge the `next` branch 🙂

PR #161, opened 2018-09-29
Some fixes for `require-tothrow-message`

garyking:
I allow for any argument type, which would include `TemplateLiteral`, `CallExpression`, etc.

I adjusted the tests by throwing actual errors, adding comments, adding more tests. 

I couldn't add a test for `TemplateLiteral` since it requires `babel-eslint`. 

I couldn't break up the commit into multiple commits because one of the precommit hooks staged and commited the entire file, even when I staged only a few lines at a time.


SimenB:
>I couldn't add a test for TemplateLiteral since it requires babel-eslint

That doesn't make any sense. Would it be enough to bump `ecmaVersion`?

>I couldn't break up the commit into multiple commits because one of the precommit hooks staged and commited the entire file, even when I staged only a few lines at a time.

What I usually do is just do `--no-verify` to skip the hook. Hopefully it'll be better in the future: https://github.com/okonet/lint-staged/pull/75

rules/__tests__/require-tothrow-message.js
```diff
@@ -7,23 +7,44 @@ const ruleTester = new RuleTester();
...
+    // String
+    "expect(function() { throw new Error('a'); }).toThrow('a');",
+    "expect(function() { throw new Error('a'); }).toThrowError('a');",
+
+    // Template literal
+    // {
+    //   code:
+    //     "const a = 'a'; expect(() => { throw new Error('a'); })" +
+    //     '.toThrow(`${a}`);',
+    //   parser: 'babel-eslint',
```
SimenB: I think it should work if you set `ecmaVersion`: https://eslint.org/docs/user-guide/configuring#specifying-parser-options

garyking:
@SimenB 	Okay changes have been made.

PR #163, opened 2018-09-29
prefer-strict-equal: Only run if `expect` is used

garyking:
Only run the rule `prefer-strict-equal` on `CallExpression`s where `expect` is used.


garyking:
@SimenB 	Okay done, please check it.

PR #164, opened 2018-09-29
New rule: no-alias-methods

garyking:
This rule ensures that only the canonical method name from the Jest docs are used for methods, rather than their alias forms. This ensures consistency in codebases, and makes it easier to search for all occurrences of a method.


SimenB:
We can follow up with that later if people want it

PR #168, opened 2018-10-02
Detect usage like 'test.only.each()' and 'describe.only.each()'

msvab:


msvab:
Oh yes, forgot a duplicate code in, should be fixed now!

PR #169, opened 2018-10-02
refactor(no-disabled-tests): use ESQuery selector syntax

macklinu:
Following up to #158, this PR refactors `no-disabled-tests` to use [ESQuery selector syntax](https://eslint.org/docs/developer-guide/selectors).

What do you think about these changes? Does this seem more maintainable / cleaner, or does this make things harder to follow? Open to feedback and not tied to this getting merged - curious what your thoughts are. 😄

### DIFF
--- a/rules/no-disabled-tests.js
+++ b/rules/no-disabled-tests.js
@@ -57,90 +57,73 @@ module.exports = {
     let testDepth = 0;
 
     return {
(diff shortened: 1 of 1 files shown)


macklinu:
Thanks for the reviews @thymikee @SimenB 🥇

PR #171, opened 2018-10-02
refactor(expect-expect): clean up and use ESQuery syntax

macklinu:
This PR refactors to use [selector syntax](https://eslint.org/docs/developer-guide/selectors). Open to feedback on if this is a helpful refactor! 😄 

Relates to #158, #169, #170

### DIFF
--- a/rules/expect-expect.js
+++ b/rules/expect-expect.js
@@ -26,45 +26,25 @@ module.exports = {
     ],
   },
   create(context) {
-    // variables should be defined here
     const unchecked = [];
     const assertFunctionNames =
       context.options[0] && context.options[0].assertFunctionNames
         ? context.options[0].assertFunctionNames
         : ['expect'];
 
-    //----------------------------------------------------------------------
-    // Helpers
-    //----------------------------------------------------------------------
-    const isExpectCall = node =>
-      // if we're not calling a function, ignore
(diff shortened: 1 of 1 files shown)


rules/expect-expect.js
```diff
@@ -26,45 +26,25 @@ module.exports = {
...
-        if (!isExpectCall(node)) {
-          return;
-        }
-        // here, we do have a call to expect
-        // use `some` to return early (in case of nested `it`s
-        context.getAncestors().some(ancestor => {
+      'CallExpression[callee.name=/^it|test$/]'(node) {
+        unchecked.push(node);
+      },
+      [`CallExpression[callee.name=/^${assertFunctionNames.join('|')}$/]`]() {
```
macklinu: This statement replaces the `isExpectCall()` function (same behavior).

rules/expect-expect.js
```diff
@@ -26,45 +26,25 @@ module.exports = {
...
+        // Return early in case of nested `it` statements.
+        for (const ancestor of context.getAncestors()) {
```
macklinu: I moved from using `Array#some()` to using a `for..of` loop because it felt more semantically correct. When we find the element we are looking for, call `unchecked.splice()` then `break`, instead of returning `true` in `Array#some()`.

PR #174, opened 2018-10-08
chore: add prefer-to-contain rule

### LINKED ISSUE #100: New rule prefer-to-contain
Prefer `expect(a).toContain(b)` rather than `expect(a.includes(b)).toBe(true)`.

The error messages are often times better.

This rule is similar to `prefer-to-have-length`.

katakonst:
fixes #100

### DIFF
--- a/docs/rules/prefer-to-contain.md
+++ b/docs/rules/prefer-to-contain.md
@@ -1,0 +1,47 @@
+# Suggest using `toContain()` (prefer-to-contain)
+
+In order to have a better failure message, `toContain()` should be used upon
+asserting expectations on an array containing an object.
+
+## Rule details
+
+This rule triggers a warning if `toBe()` or `isEqual()` is used to assert object
+inclusion in an array
+
+```js
+expect(a.includes(b)).toBe(true);
+```
+
+```js
+expect(a.includes(b)).not.toBe(true);
+```
+
+```js
+expect(a.includes(b)).toBe(false);
+```
+
+### Default configuration
+
+The following patterns are considered a warning:
+
+```js
+expect(a.includes(b)).toBe(true);
+```
+
+```js
+expect(a.includes(b)).not.toBe(true);
+```
+
+```js
+expect(a.includes(b)).toBe(false);
+```
+
+The following patterns are not a warning:
+
+```js
+expect(a).toContain(b);
+```
+
+```js
+expect(a).not.toContain(b);
+```
--- a/rules/__tests__/prefer-to-contain.test.js
+++ b/rules/__tests__/prefer-to-contain.test.js
@@ -1,0 +1,207 @@
+'use strict';
+
+const RuleTester = require('eslint').RuleTester;
+const rule = require('../prefer-to-contain');
+
+const ruleTester = new RuleTester();
+
+ruleTester.run('prefer-to-contain', rule, {
+  valid: [
+    'expect(a).toContain(b);',
+    "expect(a.name).toBe('b');",
+    'expect(a).toBe(true);',
+    `expect(a).toEqual(b)`,
+    `expect(a.test(c)).toEqual(b)`,
+    `expect(a.includes(b)).toEqual()`,
+    `expect(a.includes(b)).toEqual("test")`,
+    `expect(a.includes(b)).toBe("test")`,
+    `expect(a.includes()).toEqual()`,
+    `expect(a.includes()).toEqual(true)`,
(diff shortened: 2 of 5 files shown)


rules/__tests__/prefer-to-contain.test.js
```diff
@@ -0,0 +1,71 @@
...
+    "expect(a.name).toBe('b');",
+    'expect(a).toBe(true);',
+    `expect(a).toEqual(b)`,
+    `expect(a.test(c)).toEqual(b)`,
+    `expect(a.includes(b)).toEqual(false)`,
+    `expect(a.includes(b)).toEqual()`,
+    `expect(a.includes(b)).toEqual("test")`,
+    `expect(a.includes(b)).toBe("test")`,
+    `expect(a.includes()).toEqual()`,
+    `expect(a.includes(b)).toBe(false)`,
```
SimenB: could we support this, and suggest `expect(a).not.toContain(b)`?

macklinu: Agree on supporting the `.not.toContain()` syntax. 👍

katakonst: Added `not.toContain` suggestions

docs/rules/prefer-to-contain.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+The following patterns are not a warning:
+
+```js
+expect(a).toContain(b);
+```
+
+```js
+expect(a).toContain(b);
```
SimenB: missing `not`?

rules/prefer-to-contain.js
```diff
@@ -0,0 +1,76 @@
...
+            node.parent.parent.arguments.length === 1) &&
+          (argument(node).value === true || argument(node).value === false) &&
+          node.arguments[0] &&
+          node.arguments[0].callee &&
+          node.arguments[0].callee.property &&
+          node.arguments[0].callee.property.name === 'includes' &&
+          node.arguments[0].arguments.length === 1 &&
+          node.arguments[0].arguments[0]
+        ) {
+          const sourceCode = context.getSourceCode();
```
SimenB: could we move all these variables into the the `fix` closure to avoid the hit of calculating them unless people call `fix`?

rules/prefer-to-contain.js
```diff
@@ -0,0 +1,76 @@
...
+  meta: {
+    docs: {
+      url: getDocsUrl(__filename),
+    },
+    fixable: 'code',
+  },
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (
```
SimenB: this `if` is really verbose, could you clean it up? Not a bigge (we can do it later), but 15 lines seems excessive

rules/prefer-to-contain.js
```diff
@@ -0,0 +1,76 @@
...
+          !(
+            expectNotCase(node) ||
```
SimenB: what about e.g. `expect(a.includes(b)).not.toEqual(true);`? might be weird enough that we don't need to support it, I guess 🙂

rules/prefer-to-contain.js
katakonst: Broke it in multiple methods

rules/prefer-to-contain.js
katakonst: It'a a little weird but it's a nice to have. Added it to the rule!

docs/rules/prefer-to-contain.md
katakonst: Fixed

rules/prefer-to-contain.js
katakonst: Moved

rules/prefer-to-contain.js
```diff
@@ -0,0 +1,131 @@
...
+          context.report({
+            fix(fixer) {
+              const sourceCode = context.getSourceCode();
+
+              let fixArr = getCommonFixes(node, sourceCode, fixer);
+              if (isEqualityNegation(node)) {
+                return getNegationFixes(node, sourceCode, fixer).concat(fixArr);
+              }
+
+              const equalityArgument = isEqualityNegation(node)
```
SimenB: CI is complaining that this line is not covered by tests. Mind adding a test for the other case (or removing the ternary)?

katakonst: It was not used so removed it, the coverage is fine now.

PR #176, opened 2018-10-13
fix(expect-expect): support MemberExpressions in assertFunctionNames

### LINKED ISSUE #175: [expect-expect]: Unable to allow methods in `assertFunctionNames`
In testing the `expect-expect` rule on one of my projects, I found that it was reporting a number of tests that didn't have expect calls.  In these tests, I was using `td.verify` from [testdouble.js](https://github.com/testdouble/testdouble.js/) which verifies that a particular test double was called.

I attempted to use the `assertFunctionNames` option to configure `td.verify` as a known expect function, but was unable to get it to work.  The tests in questions were still reported as violating this rule.

I tried both `assertFunctionNames: ["expect", "td.verify"]` and `assertFunctionNames: ["expect", "verify"]` and neither approach worked.

I was able to work around the issue by doing the following, but this is not the normal way to use testdouble.js and I'd prefer not to have to do it everywhere I use `td.verify`.

```js
const { verify } = td;

test("...", () => {
  // ...
  verify(someFunctionCall());
}
```

I think the more general problem here is that the rule is not able to match anything but bare functions; it can't handle methods on an object.

I don't know the AST syntax well enough to know how to fix this, but I thought I'd at least report it here in the hope that someone more familiar can take care of it.

macklinu:
Fixes #175

### DIFF
--- a/rules/expect-expect.js
+++ b/rules/expect-expect.js
@@ -6,6 +6,7 @@
  */
 
 const getDocsUrl = require('./util').getDocsUrl;
+const getNodeName = require('./util').getNodeName;
 
 module.exports = {
   meta: {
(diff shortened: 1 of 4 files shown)


rules/no-disabled-tests.js
```diff
@@ -1,23 +1,7 @@
 'use strict';
 
 const getDocsUrl = require('./util').getDocsUrl;
-
-function getName(node) {
```
macklinu: I replaced the implementation of `util/getNodeName` with this implementation, which I think handles member expressions a little bit better than the `expect-expect` implementation.

All tests continue to pass with the refactor too. 👍

rules/expect-expect.js
```diff
@@ -33,16 +34,18 @@ module.exports = {
...
-        unchecked.push(node);
-      },
-      [`CallExpression[callee.name=/^${assertFunctionNames.join('|')}$/]`]() {
-        // Return early in case of nested `it` statements.
-        for (const ancestor of context.getAncestors()) {
-          const index = unchecked.indexOf(ancestor);
-          if (index !== -1) {
-            unchecked.splice(index, 1);
-            break;
+      CallExpression(node) {
```
macklinu: Using the selector query syntax became a little confusing now having to support MemberExpressions, so all CallExpression logic lives in this visitor function now. 😃

PR #177, opened 2018-10-15
fix(no-disabled-tests): stop picking up all funcs that start it/test

jackfranklin:
A bug in the regex means that this version picks up functions it shouldn't and warns that they don't have two arguments. For example, these lines in our codebase fail:

```
const dummyItem = itemBuilder({ // fails here
```

```
const showOutOfStockBadge = itemEntirelyOutOfStock(this.props.item);
```

Commit includes a test that was failing and the fix which is just to wrap brackets around the match in the regex.


jackfranklin:
(the test fail is odd - it looks unrelated, the actual unit tests are passing)

macklinu:
I'm going to kick the build off again and see if that fixes the build issue. Thanks for this PR. 👍

SimenB:
Thanks!

PR #179, opened 2018-10-21
feat(rules): add `no-test-callback` rule

SimenB:
Using a `done` callback is old-school, all the cool kids should use Promises these days. Especially because of `async-await`. We might also want to remove `done` from a future version of Jest, and this is a nice start to make people migrate.

/cc @rubennorte, since we talked about something like this

### DIFF
--- a/docs/rules/no-test-callback.md
+++ b/docs/rules/no-test-callback.md
@@ -46,7 +46,7 @@ will report the assertion error correctly.
 
 ## Rule details
 
-This rule triggers a warning if you have a `done` callback in your test
+This rule triggers a warning if you have a `done` callback in your test.
 
 The following patterns are considered warnings:
 
--- a/docs/rules/no-test-callback.md
+++ b/docs/rules/no-test-callback.md
@@ -1,0 +1,76 @@
+# Avoid using a callback in asynchronous tests (no-test-callback)
+
+Jest allows you to pass a callback to test definitions, typically called `done`,
+that is later invoked to indicate that the asynchronous test is complete.
+
+However, that means that if your test throws (e.g. because of a failing
+assertion), `done` will never be called unless you manually use `try-catch`.
+
+```js
+test('some test', done => {
+  expect(false).toBe(true);
+  done();
+});
+```
+
+The test above will time out instead of failing the assertions, since `done` is
+never called.
+
+Correct way of doing the same thing is to wrap it in `try-catch`.
+
+```js
+test('some test', done => {
+  try {
+    expect(false).toBe(true);
+    done();
+  } catch (e) {
+    done(e);
+  }
+});
+```
+
+However, Jest supports a second way of having asynchronous tests - using
+promises.
+
+```js
+test('some test', () => {
+  return new Promise(done => {
+    expect(false).toBe(true);
+    done();
+  });
+});
+```
+
+Even though `done` is never called here, the Promise will still reject, and Jest
+will report the assertion error correctly.
+
+## Rule details
+
+This rule triggers a warning if you have a `done` callback in your test
+
+The following patterns are considered warnings:
+
+```js
+test('myFunction()', done => {
+  // ...
+});
+
+test('myFunction()', function(done) {
+  // ...
+});
+```
+
+The following patterns are not considered warnings:
+
+```js
+test('myFunction()', () => {
+  expect(myFunction()).toBeTruthy();
+});
+
+test('myFunction()', () => {
+  return new Promise(done => {
+    expect(myFunction()).toBeTruthy();
+    done();
+  });
+});
+```
--- a/rules/__tests__/no-test-callback.test.js
+++ b/rules/__tests__/no-test-callback.test.js
@@ -14,7 +14,6 @@ ruleTester.run('no-test-callback', rule, {
     'test("something", () => {})',
     'test("something", async () => {})',
     'test("something", function() {})',
-    'test("something", function () {})',
     'test("something", async function () {})',
     'test("something", someArg)',
   ],
--- a/rules/__tests__/no-test-callback.test.js
+++ b/rules/__tests__/no-test-callback.test.js
@@ -1,0 +1,90 @@
+'use strict';
+
+const RuleTester = require('eslint').RuleTester;
+const rule = require('../no-test-callback');
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 8,
+  },
+});
+
+ruleTester.run('no-test-callback', rule, {
+  valid: [
+    'test("something", () => {})',
+    'test("something", function() {})',
+    'test("something", function () {})',
+  ],
+  invalid: [
+    {
+      code: 'test("something", done => {done();})',
+      errors: [
+        {
+          message: 'Illegal usage of test callback',
+          line: 1,
+          column: 19,
+        },
+      ],
+      output:
+        'test("something", () => {return new Promise(done => {done();})})',
+    },
+    {
+      code: 'test("something", (done) => {done();})',
+      errors: [
+        {
+          message: 'Illegal usage of test callback',
+          line: 1,
+          column: 20,
+        },
+      ],
+      output:
+        'test("something", () => {return new Promise((done) => {done();})})',
+    },
+    {
+      code: 'test("something", done => done())',
+      errors: [
+        {
+          message: 'Illegal usage of test callback',
+          line: 1,
+          column: 19,
+        },
+      ],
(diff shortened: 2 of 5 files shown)


rules/no-test-callback.js
```diff
@@ -0,0 +1,80 @@
...
+          callback.params.length !== 1
+        ) {
+          return;
+        }
+
+        const argument = callback.params[0];
+        context.report({
+          node: argument,
+          message: 'Illegal usage of test callback',
+          fix(fixer) {
```
thymikee: ❤️

rules/__tests__/no-test-callback.test.js
```diff
@@ -0,0 +1,128 @@
...
+      code: 'test("something", async done => {done();})',
+      errors: [
+        {
+          message: 'Illegal usage of test callback',
+          line: 1,
+          column: 25,
+        },
+      ],
+      output:
+        'test("something", async () => {await new Promise(done => {done();})})',
```
thymikee: a test with a bit more complex code would be nice to validate fixer

SimenB: Feel free to push some examples 🙂 Or write them here, and I can add them

rubennorte:
@SimenB looks very good! :D

rubennorte:
@SimenB looks very good! :D

rubennorte:
@SimenB looks very good! :D

rubennorte:
@SimenB looks very good! :D

rubennorte:
@SimenB looks good! :D

docs/rules/no-test-callback.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+However, Jest supports a second way of having asynchronous tests - using
+promises.
+
+```js
+test('some test', () => {
+  return new Promise(done => {
+    expect(false).toBe(true);
+    done();
```
macklinu: I think wrapping this code in a Promise as a fixer is a good first step and a consistent way to wrap code in the "return a Promise" paradigm. Do we also want to document using [`.resolves`](https://jestjs.io/docs/en/expect#resolves), [`.rejects`](https://jestjs.io/docs/en/expect#rejects) or returning a Promise / await statement as other suggested alternatives?

rules/no-test-callback.js
```diff
@@ -0,0 +1,80 @@
...
+          !/^(Arrow)?FunctionExpression$/.test(callback.type) ||
+          callback.params.length !== 1
+        ) {
+          return;
+        }
+
+        const argument = callback.params[0];
+        context.report({
+          node: argument,
+          message: 'Illegal usage of test callback',
```
macklinu: What about something like "Prefer returning a Promise instead of using a test callback"? Not sure if we could be more up front about explaining the issue in the message?

docs/rules/no-test-callback.md
SimenB: The thinking is that returning a promise is a safe change, as it will always work (I think?). Messing with the body of the tests are both way harder and more brittle.

rules/__tests__/no-test-callback.test.js
```diff
@@ -0,0 +1,128 @@
...
+    ecmaVersion: 8,
+  },
+});
+
+ruleTester.run('no-test-callback', rule, {
+  valid: [
+    'test("something", () => {})',
+    'test("something", async () => {})',
+    'test("something", function() {})',
+    'test("something", function () {})',
```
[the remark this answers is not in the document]
SimenB: yeah, makes sense

rules/no-test-callback.js
SimenB: Yeah, I actually had that in my first iteration locally, but landed on this to more closely match the name of the rule, and general language of the eslint rules I looked at for the implementation of the fixer (`arrow-parens` and `arrow-body-style`). Thoughts?

>Not sure if we could be more up front about explaining the issue in the message?

How so? Isn't that more the job of the docs?

macklinu: Yeah, I think leaving that detail to the docs is the best option here - makes sense. Thanks for providing that context too. 👍

SimenB: I'd love PRs improving the language of both the lint error and the docs - I suck at both 🙂

SimenB:
I've ran this on the Jest codebase without it crashing, or reporting false positives, BTW. So should be _fairly_ good. I'd love for people to try it out though!

docs/rules/no-test-callback.md
SimenB: oh, I now understood that you meant the docs... I thought you talked about the fixer, sorry!

Yeah, we should definitely include some prose about using at least `async-wait` over returning the promise (async functions returns promises implicitly, so it's almost always a better idea anyways). Mentioning `resolves` and `rejects` (especially `rejects` as `resolves` i better handled by `await`) is probably a good idea as well.

Would you mind sending a PR for it? 🙏

rubennorte: @SimenB rejects is very useful to avoid using logic in try/catch blocks (like `.toThrow` but for async functions), so I would keep it (and `resolves` for consistency) unless we come up with an equally good alternative.

SimenB: Yeah, for sure.

What I want is same as ava (which returns the error from `t.throws`, but I'm not sure if that makes sense for `expect`

rubennorte: I just saw that ava has `throwsAsync` for that :D

SimenB: Ah, new in v1, it was just `throws` before. But yeah, I like that pattern. Doesn't fit into `expect`, though

ron23:
Hey @SimenB Thanks for this rule! It's great and I wanna be part of the cool kids using as/aw.
I have a question about the `--fix` though: It seems like it's wrapping the `done` arg with a promise, (basically using the done as a resolver) but I don't think it really needs to.
Why not simply remove the `done` argument all together and add a return? Is that much more complicated? 

This is my original test:
```
    it('should work', done => {
      mockApi.onGet(`/account/someone`).reply(200, {});
      return Api.fetch(someone).then(result => {
        expect(result).toEqual([]);
        done();
      });
    });
```

And this is that the fix did:
```
    it('should work', () => {return new Promise(done => { => {
      mockApi.onGet(`/account/someone`).reply(200, {});
      return Api.fetch(someone).then(result => {
        expect(result).toEqual([]);
        done();
      });
    });
```

And this is what I want it to be:
```
    it('should work', () => {
      mockApi.onGet(`/account/someone`).reply(200, {});
      return Api.fetch(someone).then(result => {
        expect(result).toEqual([]);
      });
    });
```

Sorry for bumping old thread. If you want I can open a new ticket for it.

G-Rath:
@ron23 this is because of the complexity around doing that, specifically finding the right `done`:

```
it('should work', done => {
  mockApi.onGet(`/account/someone`).reply(200, {});

  return new Promise(done => {
    doSomething();
    done();
  }).then(() =>
    Api.fetch(someone).then(result => {
      expect(result).toEqual([]);
      done();
    })
  );
});
```

In the above, how do you identify the `done` that should be removed? :)

SimenB:
Ideally your original example should actually be an error since you're both returning a promise _and_ using a `done` callback. Mocha actually errors in this case - Jest really should do the same. Using both methods of signalling a test as asynchronous is usually an error, and it's definitely confusing.

So if we say you only use the done callback and don't return the promise (which is the intended way):

```js
it('should work', done => {
  mockApi.onGet(`/account/someone`).reply(200, {});
  Api.fetch(someone).then(result => {
    expect(result).toEqual([]);
    done();
  });
});
```

Then the only (relatively) safe transform we can do is wrapping the whole thing in a promise and replacing the `done` call with `resolve` from the promise executor.

Ideally, you'd just return the promise as you suggest, though, and have no `new Promise` wrapper 👍

> To be honest, this rule should probably provide a suggestion rather than a fix :/

This rule predates suggestions existing in eslint by a year. But yes, I agree - this is not really a safe fix to apply in all cases

ron23:
Thanks @SimenB and @G-Rath . Good stuff!
I used find & replace and ended up refactoring all of the tests that had that format. I can now safety enable the rule to prevent people from using `done`!

vvo:
Hey @SimenB I just got warned about this, then I tested the example you gave that should timeout and it does not times out, it just show the error:

https://codesandbox.io/s/jest-example-lcpkg

image

Could you precise maybe in the documentation what's really wrong with this pattern? Like with a use case involving maybe a time where it will really timeout? Thanks!

PR #187, opened 2018-10-24
fix(ci): update commitlint

JohanBrorson:
## This PR will
hopefully fix the failing build on node 11 by updating `commitlint/cli` and `commitlint/config-conventional` to the latest versions.

Also, stop running commitlint on node 4, since the new version of commitlint no longer supports node 4.

## Why?
The old version of `commitlint` doesn't work on node 11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ before_install:
 install: yarn --ignore-engines
 before_script: greenkeeper-lockfile-update
 script:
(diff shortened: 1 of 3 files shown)


SimenB:
Can you send this PR to the `next` branch? We've dropped node 4 there. I'll just remove node 11 from Travis for now

SimenB:
Actually, we've already upgraded on that branch, so we should be good. Need to sync it with master and add node 11 to it, though. I'll do that now.

Thanks for the PR, though!

PR #193, opened 2018-10-31
Next major version

SimenB:
Let's just do this.

Any other changes we wanna make to the recommended config (or other breaking changes)? I removed the stylistic matcher stuff. We could export something like `plugin:jest/stylistic` people can `extend`? Not sure


index.js
```diff
@@ -36,12 +36,16 @@ module.exports = {
         'jest/globals': true,
       },
       rules: {
+        'jest/no-alias-methods': 'warn',
```
thymikee: damn, I guess I like the aliased methods better :D

SimenB: I think it makes sense? The justification in the docs of the rule makes sense to me, +1 for consistency.

If the aliased names are better, maybe we should swap them in Jest? :D

thymikee: I like them because they're shorter, but don't care too much because Flow autocompletes it for me. Consistency matters though. I wish it was autofixable though!

SimenB: PR welcome? 😀

SimenB: Wait, it _is_ fixable: https://github.com/jest-community/eslint-plugin-jest/blob/db0fdd835050bbe6e96d4ce79706d45a34760bad/rules/no-alias-methods.js#L49-L51

thymikee: I was just about to write that!

SimenB: Updated the readme to include the badge. Also updated the readme in this PR to reflect which rules are recommended, I forgot that

thymikee: How about adding `prefer-to-contain`, `prefer-to-be-undefined` and `prefer-to-be-null`? They're pretty useful and all fixable

SimenB: What about my suggestion in the OP?

>I removed the stylistic matcher stuff. We could export something like `plugin:jest/stylistic`

thymikee: Makes sense. Maybe jest/style would be easier to type

rules/valid-expect-in-promise.js
```diff
@@ -76,12 +75,12 @@ const getFunctionBody = func => {
...
 
 const getTestFunction = node => {
-  let parent = node.parent;
+  let { parent } = node;
   while (parent) {
     if (isFunction(parent) && isTestFunc(parent.parent)) {
       return parent;
     }
-    parent = parent.parent;
+    ({ parent } = parent);
```
macklinu: This statement is a little confusing - could you help explain what's going on with this destructure statement?

index.js
macklinu: `jest/style` sounds good to me too.

rules/valid-expect-in-promise.js
SimenB: Yeah, I just removed it. The syntax is confusing.

index.js
SimenB: done!

PR #207, opened 2018-11-26
Add no-truthy-falsy rule

noahbenham:
This resolves #204.

This is my first real submission to this repo. If there's anything missing here, please let me know.

I've tested this against my company's private repo and it appears to be working as expected. This rule is not safely auto-fixable.

@KevinHerklotz @macklinu @SimenB

PR #208, opened 2018-11-26
Danger fails for new rules without README changes

noahbenham:
This came up in a discussion on #207 so I split this change off into a separate PR.


SimenB:
Haha.

Maybe check that there has been changes (added files?) In `rules/`?

noahbenham:
Yeah I probably want `find`, not `filter`. At least we know it works 😁

PR #210, opened 2018-11-30
Fix: no-alias-methods supports .resolves & .rejects

noahbenham:
The autofix for `no-alias-methods` didn't catch everything it could've in our internal repo. Specifically `.resolves` & `.rejects` were missed.

This PR assumes that the full list of `methodNames` can be prefixed with `.resolves` and `.rejects`.

There may be other helpers I've overlooked that could be added here. For now, I erred on the side of caution.


SimenB:
Does this cover `not` as well?

noahbenham:
Added, thanks for the feedback @SimenB.

SimenB:
Thank you!

Issue #217, opened 2019-01-27
[new rule] Prefer test.todo()


macklinu:
Introduced in Jest 24 (see [this blog post](https://jestjs.io/blog/2019/01/25/jest-24-refreshing-polished-typescript-friendly#testtodo)), it seems like we could support a rule that would prefer the following:

```js
test.todo('i need to write this test')
```

instead of:

```js
test('i need to write this test')
// and/or
test('i need to write this test', () => {})
```

Haven't thought much about specifics, rule naming, etc. Just came here from the [Jest 24 blog post](https://jestjs.io/blog/2019/01/25/jest-24-refreshing-polished-typescript-friendly#testtodo) and wanted to jot this down.

doniyor2109:
Hi. Can I work on this?

SimenB:
Yeah, go for it!

It should also warn against `test.skip('asdasdad', () => {})`

PR #218, opened 2019-01-27
feat(rules): add prefer-todo rule

doniyor2109:
Added prefer-todo rule as mentioned #217

### DIFF
--- a/docs/rules/prefer-todo.md
+++ b/docs/rules/prefer-todo.md
@@ -11,8 +11,6 @@ This rule triggers a warning if empty test case is used without 'test.todo'.
 test('i need to write this test');
 ```
 
-This rule is enabled by default.
-
 ### Default configuration
 
 The following pattern is considered warning:
--- a/docs/rules/prefer-todo.md
+++ b/docs/rules/prefer-todo.md
@@ -18,7 +18,9 @@ This rule is enabled by default.
 The following pattern is considered warning:
 
 ```js
-test('i need to write this test');
+test('i need to write this test'); // Unimplemented test case
+test('i need to write this test', () => {}); // Empty test case body
+test.skip('i need to write this test', () => {}); // Empty test case body
 ```
 
 The following pattern is not warning:
--- a/docs/rules/prefer-todo.md
+++ b/docs/rules/prefer-todo.md
@@ -1,0 +1,28 @@
+# Suggest using `test.todo` (prefer-todo)
+
+When test cases are empty then it is better to mark them as `test.todo` as it
+will be highlighted in the summary output.
+
+## Rule details
+
+This rule triggers a warning if empty test case is used without 'test.todo'.
+
+```js
+test('i need to write this test');
+```
+
+This rule is enabled by default.
+
+### Default configuration
+
+The following pattern is considered warning:
+
+```js
+test('i need to write this test');
+```
+
+The following pattern is not warning:
+
+```js
+test.todo('i need to write this test');
+```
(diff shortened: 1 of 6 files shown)


doniyor2109:
I came up with `prefer-todo` name for this rule

SimenB:
Thanks! Can you add code for `test.skip` as well?

rules/__tests__/prefer-todo.test.js
```diff
@@ -0,0 +1,49 @@
...
+const rule = require('../prefer-todo');
+
+const ruleTester = new RuleTester({
+  parserOptions: { ecmaVersion: 2015 },
+});
+
+ruleTester.run('prefer-todo', rule, {
+  valid: [
+    'test.todo("i need to write this test");',
+    'test("stub", () => expect(1).toBe(1));',
```
SimenB: https://github.com/facebook/jest/blob/665a93eec70c4f2d7809d1848fbc8ca60cfd8361/e2e/test-in-root/spec.js#L9

rules/__tests__/prefer-todo.test.js
```diff
@@ -0,0 +1,49 @@
...
+    `
+      supportsDone && params.length < test.length
```
SimenB: https://github.com/facebook/jest/blob/6081f7a4f02f28a5446857bb9e33a54f84a61333/packages/jest-each/src/bind.js#L179-L181

rules/prefer-todo.js
```diff
@@ -34,12 +34,20 @@ function removeSecondArg({ arguments: [first, second] }, fixer) {
...
+  return firstArg && firstArg.type === 'SpreadElement';
+}
+
 function create(context) {
   return {
     CallExpression(node) {
       if (isTestCase(node)) {
         const combineFixers = composeFixers(node);
 
+        if (isFirstArgSpread(node)) {
```
SimenB: you also need to somehow not touch https://github.com/facebook/jest/blob/6081f7a4f02f28a5446857bb9e33a54f84a61333/packages/jest-each/src/bind.js#L219-L220

that doesn't use spread

SimenB: Yeah, probably. It might also be fine to use ignores in jest's code base. But being on the safe side seems better.

Not sure if it makes sense to only autofix if there's a string, but still warn?

doniyor2109: I think developers only left tests empty with strings to make them later. If there is other than strings then they probably going to make some utils. I think we should apply this rule for strings only. What do yo think?

SimenB: Yeah, let's start with that

docs/rules/prefer-todo.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Rule details
+
+This rule triggers a warning if empty test case is used without 'test.todo'.
+
+```js
+test('i need to write this test');
+```
+
+This rule is enabled by default.
```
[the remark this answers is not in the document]
doniyor2109: Thanks for feedback

rules/prefer-todo.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function isFirstArgString({ arguments: [firstArg] }) {
+  return firstArg && isString(firstArg);
+}
+
+function create(context) {
+  return {
+    CallExpression(node) {
+      if (isTestCase(node) && isFirstArgString(node)) {
```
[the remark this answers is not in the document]
SimenB: You cannot call `fit.todo` or `xit.todo`. Only a single modifier can be applied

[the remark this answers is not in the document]
doniyor2109: So we should apply this rule for `it`, `test`, `it.skip`, `test.skip`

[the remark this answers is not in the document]
SimenB: yes

SimenB:
Thank you!

PR #221, opened 2019-01-31
fix: use range instead of start/end properties

armano2:
This PR replaces usage of non standard fields `start`, `end` with `range`.

Properties `start`, `end` should not be used (they may be removed in feature)

currently those rules are crashing because of that if you decide to use ([@typescript-eslint/parser](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser)) 

see: https://eslint.org/docs/developer-guide/working-with-custom-parsers#all-nodes

fixes: https://github.com/typescript-eslint/typescript-eslint/issues/175

### DIFF
--- a/rules/lowercase-name.js
+++ b/rules/lowercase-name.js
@@ -78,8 +78,8 @@ module.exports = {
               const description = testDescription(node);
 
               const rangeIgnoringQuotes = [
-                firstArg.start + 1,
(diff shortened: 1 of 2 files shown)


SimenB:
https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/6

:(

Issue #109, opened 2018-04-23
[new rule] prefer toHaveBeenCalledTimes()


robcresswell:
This is a new rule suggestion; it would be nice to enforce `toHaveBeenCalledTimes()` over `toHaveBeenCalled()`, to ensure functions are only being executed a set amount of times. It's unlikely making this fixable would be useful, as we would have to make an assumption about the number of times, which could well be incorrect.

I'm happy to make a PR to support this, but don't have a definite timeframe. I thought it was worth opening the issue for discussion first.

SimenB:
I think that's a great idea!

It would also be cool to fix thing like `expect(someMock.mock.calls.length).toBe(3)` into `expect(someMock).toHaveBeenCalledTimes(3)`, but it might give some false positives?

robcresswell:
That's a good idea, I hadn't thought of that. I can't think of any examples of how that would cause a false positive off the top of my head, but thats only off a couple thousand lines of experience so hardly a big sample. If nobody beats me to it, I'll have a go at writing the rule out. Should get time tomorrow evening.

SimenB:
e.g.:

```js
const someMock = {
  mock: {
    calls: ['something', 'weird']
  }
}
```

Not sure how realistic that is, though.

edbrannin:
@SimenB I don't quite follow your last comment.  Were you trying to illustrate a situation where changing `expect(foo.mock.calls.length).toBe(x)` to `expect(foo).toHaveBeenCalledTimes(x)` might break something?

(Because I came here wondering if someone had done that rule yet)

SimenB:
`toHaveBeenCalledTimes` verifies it's called with a mock function - so if you have an object that's _not_ a mock function, autofixing would break your test

G-Rath:
@SimenB sounds like this is a fix we should *suggest* 😉

SimenB:
hah, you _called_ it

benmonro:
I was just going to suggest this feature as well.  the rule should also handle `expect(some.mock.calls).toHaveLength(x)` and `expect(some.mock.calls[0][0]).toBe("foo")`

first one could be fixable, 2nd... might be trickier.

PR #224, opened 2019-02-14
Add new prefer-called-with rule

bz2:
Suggest using `toBeCalledWith()` or `toHaveBeenCalledWith()` instead of the forms without argument checking.

This is a rule I've been enforcing on an ad-hoc basis at work, and I think a better general suggestion than `toBeCalledTimes(1)` suggested in https://github.com/jest-community/eslint-plugin-jest/issues/109 for robust assertions.

Similar to [prefer-toHaveBeenCalledWith from eslint-plugin-jasmine](https://github.com/tlvince/eslint-plugin-jasmine/pull/134).

This rule has no opinion on which matcher alias should be used.


bz2:
@SimenB Thanks for landing and releasing so quickly!

PR #235, opened 2019-03-10
feat(rules): add `short` parameter to `no-alias-methods`

### LINKED ISSUE #234: Add an option to `no-alias-methods` to prefer the alias methods.
I find the rule very useful, because the consistent naming allows better search and replacement.
However, the "canonical" names are almost twice as long.
I appreciate that there may be more references to the "canonical" names out in the wild, but I find the official documentation sufficient, and the aliases search well enough there.

qm3ster:
Fixes #234 

@macklinu please have a look if you think this should be improved

### DIFF
--- a/docs/rules/no-alias-methods.md
+++ b/docs/rules/no-alias-methods.md
@@ -1,15 +1,28 @@
-# Don't use alias methods (no-alias-methods)
+# Don't mix alias methods (no-alias-methods)
 
 Several Jest methods have alias names, such as `toThrow` having the alias of
 `toThrowError`. This rule ensures that only the canonical name as used in the
-Jest documentation is used in the code. This makes it easier to search for all
-occurrences of the method within code, and it ensures consistency among the
+Jest documentation is used in the code. Alternatively, it ensures that only the
+shorter aliases are used in the code. Both make it easier to search for all
+occurrences of the method within code, and it ensure consistency among the
 method names used.
 
 ## Rule details
 
-This rule triggers a warning if the alias name, rather than the canonical name,
-of a method is used.
+This rule can be configured as follows
+
+```js
+{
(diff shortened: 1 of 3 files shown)


qm3ster:
Yup, [only one build failed](https://travis-ci.org/jest-community/eslint-plugin-jest/builds/504182577), not my fault

macklinu:
With this option added, I feel like the `no-alias-methods` rule name doesn't make sense anymore. Maybe something like `consistent-alias-methods`? That would be a breaking change, but I'm curious to think about the addition to this rule and what would be the best naming / clearest API for it and see if we'd want to introduce a breaking change for it or not.

Thoughts from others (maintainers and community)?

qm3ster:
Thank you for your feedback.
Yeah, I just wanted to make it non-breaking. Felt like that would make it easiest to get it in right away.

I think the perfect API would be `alias-methods: "never"|"short"` (Doubt `"always"` is a good one here)
It would be easy to add it in a non-breaking fashion by duplicating the rule under the new name and in the current or future version marking the `no-` one deprecated.

I can envision someone wanting to just manually configure which names are used, but I'm sure that `"short"` would be the most popular strategy after `"canonical"`. Can always add `"custom"` variant for first positional argument to take more parameters.

SimenB:
Sorry about the late response!

I'm down with the idea of a separate rule, marking the current one as deprecated in the docs, and removing it in the next major

SimenB:
@qm3ster ping 🙂

qm3ster:
Yes, it me!
I agree with that viewpoint.
I suppose this PR should be closed.
If I go ahead with that, it should be a new one.

SimenB:
That works for me 🙂 Feel free to proceed however you like

PR #237, opened 2019-03-10
fix(jest/no-identical-title): not reporting when using backticks

### LINKED ISSUE #232: jest/no-identical-title doesn't report if using backticks
Just checked and it appears that `jest/no-identical-title` doesn't report an error when specifying the title using backticks (even when not using string interpolation).

**Error:**
```js
describe('an odd case where identical titles are not reported', () => {
    it('this works as expected', () => {
        //  ...test code here
    });
    it('this works as expected', () => {
        //  ...test code here
    });
});
```

**Not reported:**
```js
describe('an odd case where identical titles are not reported', () => {
    it(`this does not work as expected`, () => {
        //  ...test code here
    });
    it(`this does not work as expected`, () => {
        //  ...test code here
    });
});
```

himynameisdave:
Currently `jest/no-identical-title` only checks against regular strings. As a result, this wouldn't get reported:

```
describe('an odd case where identical titles are not reported', () => {
    it(`this does not work as expected`, () => {
        //  ...test code here
    });
    it(`this does not work as expected`, () => {
        //  ...test code here
    });
});
```

This ticket fixes that by testing against `TemplateLiteral` nodes, but will never report if that node is using string interpolation (ie: has expressions, `${}` inside it).

Resolves #232

### DIFF
--- a/rules/__tests__/no-identical-title.test.js
+++ b/rules/__tests__/no-identical-title.test.js
@@ -12,6 +12,8 @@ ruleTester.run('no-identical-title', rule, {
       '   it("it", function() {});',
       '});',
     ].join('\n'),
+    ['describe();describe();'].join('\n'),
+    ['it();it();'].join('\n'),
     [
       'describe("describe1", function() {',
       '   it("it1", function() {});',
--- a/rules/__tests__/no-identical-title.test.js
+++ b/rules/__tests__/no-identical-title.test.js
@@ -81,6 +81,17 @@ ruleTester.run('no-identical-title', rule, {
       '    });',
       '});',
     ].join('\n'),
+    {
+      code: [
+        'describe("describe", () => {',
+        '    it(`testing ${someVar} with the same title`, () => {});',
+        '    it(`testing ${someVar} with the same title`, () => {});',
+        '});',
+      ].join('\n'),
+      env: {
+        es6: true,
+      },
+    },
   ],
 
   invalid: [
@@ -209,6 +220,25 @@ ruleTester.run('no-identical-title', rule, {
             'Describe block title is used multiple times in the same describe block.',
           column: 1,
           line: 4,
+        },
+      ],
+    },
+    {
+      code: [
+        'describe("describe", () => {',
+        '    it(`testing backticks with the same title`, () => {});',
+        '    it(`testing backticks with the same title`, () => {});',
+        '});',
+      ].join('\n'),
+      env: {
+        es6: true,
+      },
+      errors: [
+        {
+          message:
+            'Test title is used multiple times in the same describe block.',
+          column: 5,
+          line: 3,
         },
       ],
     },
--- a/rules/no-identical-title.js
+++ b/rules/no-identical-title.js
@@ -1,6 +1,12 @@
 'use strict';
 
-const { getDocsUrl, isDescribe, isTestCase } = require('./util');
+const {
+  getDocsUrl,
+  isDescribe,
+  isTestCase,
+  isString,
+  hasExpressions,
+} = require('./util');
 
 const newDescribeContext = () => ({
   describeTitles: [],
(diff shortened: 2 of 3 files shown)


rules/__tests__/no-identical-title.test.js
```diff
@@ -81,6 +81,17 @@ ruleTester.run('no-identical-title', rule, {
       '    });',
       '});',
     ].join('\n'),
+    {
+      code: [
+        'describe("describe", () => {',
+        '    it(`testing ${someVar} with the same title`, () => {});',
+        '    it(`testing ${someVar} with the same title`, () => {});',
```
macklinu: I feel like this case could be statically analyzed and marked as invalid, but I'm also not super worried about digging into handling that edge case in this PR (the code would probably be a little complicated to validate this). Thoughts from others?

rules/no-identical-title.js
```diff
@@ -51,10 +64,10 @@ module.exports = {
         if (isDescribe(node)) {
           contexts.push(newDescribeContext());
         }
-        if (!isFirstArgLiteral(node)) {
+        const [firstArgument] = node.arguments;
```
macklinu: I think that `firstArgument` can technically be undefined. I'm not 100% sure we are protecting against that case. For example, the following code:

```js
describe()
```

produces the following AST

```json
{
  "type": "Program",
  "start": 0,
  "end": 10,
  "range": [
    0,
    10
  ],
  "body": [
    {
      "type": "ExpressionStatement",
      "start": 0,
      "end": 10,
      "range": [
        0,
        10
      ],
      "expression": {
        "type": "CallExpression",
        "start": 0,
        "end": 10,
        "range": [
          0,
          10
        ],
        "callee": {
          "type": "Identifier",
          "start": 0,
          "end": 8,
          "range": [
            0,
            8
          ],
          "name": "describe"
        },
        "arguments": []
      }
    }
  ],
  "sourceType": "module"
}
```

So `node.arguments[0] === undefined`.

rules/util.js
```diff
@@ -134,6 +134,8 @@ const isString = node =>
   (node.type === 'Literal' && typeof node.value === 'string') ||
   node.type === 'TemplateLiteral';
 
+const hasExpressions = node => !!(node.expressions && node.expressions.length);
```
macklinu: The double `!!` was a little unclear reading at first. May I suggest?

```js
const hasExpressions = node => node.expressions && node.expressions.length > 0;
```

That way, the function is always returning a boolean without the extra type cast.

rules/__tests__/no-identical-title.test.js
SimenB: What if I assign to/mutate `someVar` between test declarations?
Feels like a slippery slope to start with usage analysis. I think it's fine to leave as-is

rules/no-identical-title.js
SimenB: Good catch, should be a test case

himynameisdave: Yep totally agree, will add 👍

rules/__tests__/no-identical-title.test.js
himynameisdave: I agree that this would be nice to have, but for the time being it's fine as is.

rules/util.js
himynameisdave: Thanks for the tip!

rules/__tests__/no-identical-title.test.js
macklinu: Definitely a slippery slope - didn't think about that case. 😅 Moving along... 😂

rules/util.js
himynameisdave: I think that my thinking was for if `node.expressions`  was not defined, the entire statement would evaluate to `undefined` (instead of an actual Boolean). Doesn't matter though because `undefined` is falsey, so I'll make that change to the test against `node.expressions.length` 👍

himynameisdave:
@macklinu `||` @SimenB : I've resolved the comments, just let me know if you need me to fix anything else up! 👍

PR #238, opened 2019-03-10
feat(rules): Add no-empty-title rule

himynameisdave:
Having an empty string as title is pretty useless. This adds a new rule to ensure that tests have real titles.

Resolves #226

---

The following patterns are considered warnings:

```js
describe('', () => {});
describe('foo', () => {
  it('', () => {});
});
it('', () => {});
test('', () => {});
xdescribe('', () => {});
xit('', () => {});
xtest('', () => {});
```

These patterns would not be considered warnings:

```js
describe('foo', () => {});
describe('foo', () => {
  it('bar', () => {});
});
test('foo', () => {});
it('foo', () => {});
xdescribe('foo', () => {});
xit('foo', () => {});
xtest('foo', () => {});
```

PR #240, opened 2019-03-15
fix: make sure to export all available rules

SimenB:
Ref https://github.com/jest-community/eslint-plugin-jest/pull/238#issuecomment-473419411


__tests__/rules.test.js
```diff
@@ -16,4 +19,13 @@ describe('rules', () => {
       }
     });
   });
+
+  it('should have the correct amount of rules', () => {
```
himynameisdave: giphub

PR #253, opened 2019-05-06
docs(valid-describe): add suggestion to disable for dynamic tests

### LINKED ISSUE #203: [jest/valid-describe] "First argument must be name" incorrect error
```
import { MY_REDUX_ACTION } from "./actions";

describe(MY_REDUX_ACTION, () => {
	//...
});
```

We're getting the error "First argument must be name" [jest/valid-describe]. This appears to be because the first argument is a variable instead of a string. 

However, our use case is testing redux store actions. Those action types are exported as strings. We use those strings to name our jest tests. 

This seems like a reasonable situation that is incorrectly marked as wrong.

chrisdothtml:
Fixes #203

### DIFF
--- a/rules/__tests__/valid-describe.test.js
+++ b/rules/__tests__/valid-describe.test.js
@@ -45,7 +45,8 @@ ruleTester.run('valid-describe', rule, {
       code: 'describe(() => {})',
       errors: [
         {
-          message: 'First argument must be name',
+          message:
+            'Name must be a string literal (disable this rule inline for dynamic tests)',
           line: 1,
           column: 10,
         },
--- a/docs/rules/valid-describe.md
+++ b/docs/rules/valid-describe.md
@@ -5,8 +5,25 @@ errors.
 
 ## Rule Details
 
-This rule validates that the second parameter of a `describe()` function is a
-callback function. This callback function:
+This rule validates the signature of `describe()` functions. It validates two
+things:
+
+#### 1. The first argument is a string literal
+
+NOTE: if you are using a dynamic value, disable the rule inline like so:
+
+```js
+// eslint-disable-next-line jest/valid-describe
+describe(myTestName, () => {
+```
+
+See
+[this comment](https://github.com/jest-community/eslint-plugin-jest/pull/253#issuecomment-491371038)
+for reasoning behind not supporting dynamic values.
+
+#### 2. The second argument is a callback function
+
+This callback function:
 
 - should not be
   [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)
--- a/rules/valid-describe.js
+++ b/rules/valid-describe.js
@@ -47,7 +47,7 @@ module.exports = {
           if (!isString(name)) {
             context.report({
               message:
-                'Name must be a string literal (if writing dynamic tests, disable this rule)',
+                'Name must be a string literal (disable this rule inline for dynamic tests)',
               loc: paramsLocation(node.arguments),
             });
           }
--- a/rules/valid-describe.js
+++ b/rules/valid-describe.js
@@ -46,7 +46,8 @@ module.exports = {
           const [, callbackFunction] = node.arguments;
           if (!isString(name)) {
             context.report({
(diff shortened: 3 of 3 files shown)


chrisdothtml:
Think CI just needs to be restarted

jeysal:
@chrisdothtml Thanks for taking time to fix this!
I wonder though: If we allow `x.name`, why not e.g. `x.getName()` (another kind of expression)? Both could be incorrect just as well as both could be correct. And if we allow both (in fact, all kinds of expressions that we cannot infer the type of), does the plugin provide a big benefit over Jest checking what type the name really is at runtime?
I guess you could error only if it's clearly something invalid like an object expression. Is that better than allowing only clearly valid things (the by far most common case) and recommending eslint-disable / template string-ish workarounds in special cases though? Just something that came to my mind looking at the issue thread :)
cc @macklinu @SimenB

chrisdothtml:
> If we allow `x.name`, why not e.g. `x.getName()`

Ah, good point; I was just covering the cases mentioned in #203

> Both could be incorrect just as well as both could be correct

Even further than that, `x.name` could also be incorrect if using a Proxy or getter; but that's Javascript for ya 🤷‍♂. I think the benefit this plugin provides is more for syntactical/signature checking rather than type checking.

In that light, perhaps the best way to resolve this is to just remove the `First argument must be name` error and rely on `Describe requires name and callback arguments` to check the amount of arguments

SimenB:
If we could extract some type info that it's a string bit might make sense, but at that point it's not really a lint rule's responsibility.

Thoughts on making docs and the error say something like "must be string literal" or something, and recommending disabling the rule if you have dynamic names (like you mention)?

chrisdothtml:
> Thoughts on making docs and the error say something like "must be string literal" or something, and recommending disabling the rule if you have dynamic names (like you mention)?

Yeah I think that's the best solution. Updated my PR to just a docs update

rules/__tests__/valid-describe.test.js
```diff
@@ -45,7 +45,8 @@ ruleTester.run('valid-describe', rule, {
       code: 'describe(() => {})',
       errors: [
         {
-          message: 'First argument must be name',
+          message:
```
SimenB: Look at the code - this should keep the old message. Cab we differentiate between missing first arg/wrong type and a variable?

chrisdothtml: I don't think I follow. Could you explain what exactly you're looking for?

jeysal: I think the point is that the new message suggests that the mistake was to pass a function instead of a string, when the mistake is really that the name is just missing before the function. The old message was neutral in this regard.
And I guess @SimenB's suggestion (second sentence) was to print something like "First argument must be name" in this case because we can infer that the user probably just forgot the name in before the function, and only if both args are there but the first is not a string literal print the new one :)

SimenB: Yes, that's what I meant. Thanks! 😀

thisissami:
Seeing as how this hasn't been merged in yet, I'm just gonna chime in with a different opinion: the `linter` should not enforce a particular form of strings, period. The first argument for a `describe` should be a string, and _any_ valid way of generating a string there should be valid. If there is no way to do this, then this should not be the linters responsibility, and this rule should be completely removed. 

If people want typechecking, they can use `typescript`, which `jest` fully supports; it will complain if you don't pass in a string. 

If the community insists on this being a rule for the linter, then the message that it expresses should make it really clear that this might not actually be an issue. E.g. instead of 

> First argument must be name

it could be

> First argument must be name. Also if you are using a string, then please ignore this message.

^if something like that is needed, then... it's a code smell and probably shouldn't be something the linter handles. 

My $.02.

pedro-mass:
What @thisissami suggested, or provide a config object that allows ignoring the name check

chrisdothtml:
Closing as I don't plan on finishing this PR, so anyone can free to pick up the change. Doesn't seem like there's a solid verdict on how this should be handled though

PR #255, opened 2019-05-09
Assert that async `expect`s are awaited or returned - Closes #54, #254

### LINKED ISSUE #54: Assert that async `expect`s are awaited or returned

### LINKED ISSUE #254: Bug Report: expectRejectCase & expectResolveCase is not working
Hello, I was interested to solve the issue https://github.com/jest-community/eslint-plugin-jest/issues/54 and I was looking at the code base. 

https://github.com/jest-community/eslint-plugin-jest/blob/05eb11ace731d1f675711adbc0730a61ee36f2a9/rules/util.js#L20-L28

I realised that expectRejectCase & expectResolveCase functions are actually not working. `methodName(node) === 'reject';` here the method name must be `rejects` I think, and the same goes with `resolves`.

I found the usages for those functions and removing them actually doesn't break anything, means they were not working.

So, would it be ok if i fix these 2 functions and use them to solve issue #54 ?

yatki:
## Acceptance Criteria

- [x] Improve `valid-expect` rule
- [x] Add test coverage
- [x] Update docs
- [x] Enable `alwaysAwait` option in plugin config

closes #54,
closes #254

### DIFF
--- a/docs/rules/valid-expect.md
+++ b/docs/rules/valid-expect.md
@@ -20,7 +20,7 @@ or when a matcher function was not called, e.g.:
 expect(true).toBeDefined;
 ```
 
-or when an async assertion was not awaited or returned, e.g.:
+or when an async assertion was not `await`ed or returned, e.g.:
 
 ```js
 expect(Promise.resolve('Hi!')).resolves.toBe('Hi!');
@@ -76,8 +76,8 @@ test('test2', () => expect(Promise.resolve(2)).resolves.toBe(2));
 
 ### `allowPromiseMethods`
 
-When set to true, disables triggers on async assertions that were used inside a
-Promise method.
+When set to true, disables triggering warnings on async assertions that were
+used inside a Promise method, even if not immediately `await`ed or returned.
 
 Examples of **correct** code for the { "allowPromiseMethods": **true** } option:
 
--- a/docs/rules/valid-expect.md
+++ b/docs/rules/valid-expect.md
@@ -20,7 +20,96 @@ or when a matcher function was not called, e.g.:
 expect(true).toBeDefined;
 ```
 
+or when an async assertion was not awaited or returned, e.g.:
+
+```js
+expect(Promise.resolve('Hi!')).resolves.toBe('Hi!');
+```
+
 This rule is enabled by default.
+
+## Options
+
+```js
+{
+  type: 'object',
+  properties: {
+    alwaysAwait: {
+      type: 'boolean',
+      default: false,
+    },
+    allowPromiseMethods: {
+      type: 'boolean',
+      default: false,
+    },
+  },
+  additionalProperties: false,
+}
+```
+
+### `alwaysAwait`
+
+Enforces to use `await` inside block statements. Using `return` will trigger a
+warning. Returning one line statements with arrow functions is _always allowed_.
+
+Examples of **incorrect** code for the { "alwaysAwait": **true** } option:
+
+```js
+// alwaysAwait: true
+test('test1', async () => {
+  await expect(Promise.resolve(2)).resolves.toBeDefined();
+  return expect(Promise.resolve(1)).resolves.toBe(1); // `return` statement will trigger a warning
+});
+```
+
+Examples of **correct** code for the { "alwaysAwait": **true** } option:
+
+```js
+// alwaysAwait: true
+test('test1', async () => {
+  await expect(Promise.resolve(2)).resolves.toBeDefined();
+  await expect(Promise.resolve(1)).resolves.toBe(1);
+});
+
+test('test2', () => expect(Promise.resolve(2)).resolves.toBe(2));
+```
+
+### `allowPromiseMethods`
+
+When set to true, disables triggers on async assertions that were used inside a
+Promise method.
+
+Examples of **correct** code for the { "allowPromiseMethods": **true** } option:
+
+```js
+// allowPromiseMethods: true
+test('test1', () => {
+  Promise.all([
+    expect(Promise.resolve(1)).resolves.toBeDefined(),
+    expect(Promise.resolve(2)).resolves.toBeDefined(),
+  ]);
+});
+```
+
+Examples of **correct** code for the { "allowPromiseMethods": **false** }
+option:
+
+```js
+// allowPromiseMethods: false
+test('test1', async () => {
+  await Promise.all([
+    expect(Promise.resolve(1)).resolves.toBeDefined(),
+    expect(Promise.resolve(2)).resolves.toBeDefined(),
+  ]);
+});
+
+test('test2', () => {
+  return Promise.all([
+    expect(Promise.resolve(1)).resolves.toBeDefined(),
+    expect(Promise.resolve(2)).resolves.toBeDefined(),
+  ]);
+});
+```
 
 ### Default configuration
 
@@ -33,6 +122,8 @@ expect('something', 'else');
 expect('something');
 expect(true).toBeDefined;
 expect(Promise.resolve('hello')).resolves;
+expect(Promise.resolve('hello')).resolves.toEqual('hello');
+Promise.resolve(expect(Promise.resolve('hello')).resolves.toEqual('hello'));
 ```
 
 The following patterns are not warnings:
@@ -41,5 +132,8 @@ The following patterns are not warnings:
 expect('something').toEqual('something');
 expect([1, 2, 3]).toEqual([1, 2, 3]);
 expect(true).toBeDefined();
-expect(Promise.resolve('hello')).resolves.toEqual('hello');
+await expect(Promise.resolve('hello')).resolves.toEqual('hello');
+await Promise.resolve(
+  expect(Promise.resolve('hello')).resolves.toEqual('hello'),
+);
 ```
--- a/docs/rules/valid-expect.md
+++ b/docs/rules/valid-expect.md
@@ -38,10 +38,6 @@ This rule is enabled by default.
       type: 'boolean',
       default: false,
     },
-    allowPromiseMethods: {
-      type: 'boolean',
-      default: false,
-    },
   },
   additionalProperties: false,
 }
@@ -74,43 +70,6 @@ test('test1', async () => {
 test('test2', () => expect(Promise.resolve(2)).resolves.toBe(2));
 ```
 
-### `allowPromiseMethods`
-
-When set to true, disables triggering warnings on async assertions that were
-used inside a Promise method, even if not immediately `await`ed or returned.
-
-Examples of **correct** code for the { "allowPromiseMethods": **true** } option:
-
-```js
-// allowPromiseMethods: true
-test('test1', () => {
-  Promise.all([
-    expect(Promise.resolve(1)).resolves.toBeDefined(),
-    expect(Promise.resolve(2)).resolves.toBeDefined(),
-  ]);
-});
-```
-
-Examples of **correct** code for the { "allowPromiseMethods": **false** }
-option:
-
-```js
-// allowPromiseMethods: false
-test('test1', async () => {
-  await Promise.all([
-    expect(Promise.resolve(1)).resolves.toBeDefined(),
-    expect(Promise.resolve(2)).resolves.toBeDefined(),
-  ]);
-});
-
-test('test2', () => {
-  return Promise.all([
-    expect(Promise.resolve(1)).resolves.toBeDefined(),
-    expect(Promise.resolve(2)).resolves.toBeDefined(),
-  ]);
-});
-```
-
 ### Default configuration
 
 The following patterns are considered warnings:
@@ -124,6 +83,10 @@ expect(true).toBeDefined;
 expect(Promise.resolve('hello')).resolves;
 expect(Promise.resolve('hello')).resolves.toEqual('hello');
 Promise.resolve(expect(Promise.resolve('hello')).resolves.toEqual('hello'));
+Promise.all([
+  expect(Promise.resolve('hello')).resolves.toEqual('hello'),
+  expect(Promise.resolve('hi')).resolves.toEqual('hi'),
+]);
 ```
 
 The following patterns are not warnings:
@@ -135,5 +98,9 @@ expect(true).toBeDefined();
 await expect(Promise.resolve('hello')).resolves.toEqual('hello');
 await Promise.resolve(
   expect(Promise.resolve('hello')).resolves.toEqual('hello'),
+);
+await Promise.all(
+  expect(Promise.resolve('hello')).resolves.toEqual('hello'),
+  expect(Promise.resolve('hi')).resolves.toEqual('hi'),
 );
 ```
--- a/src/rules/valid-expect.js
+++ b/src/rules/valid-expect.js
@@ -15,12 +15,54 @@ const {
 } = require('./util');
 
 const expectProperties = ['not', 'resolves', 'rejects'];
+const promiseArgumentTypes = ['CallExpression', 'ArrayExpression'];
 
-const getParentCallExpressionNode = node => {
+/**
+ * expect statement can have chained matchers.
+ * We are looking for the closest CallExpression
+ * to find `expect().x.y.z.t()` usage.
+ *
+ * @Returns CallExpressionNode
+ */
+const getClosestParentCallExpressionNode = node => {
+  if (!node || !node.parent || !node.parent.parent) {
+    return null;
+  }
+
   if (node.parent.type === 'CallExpression') {
     return node.parent;
   }
-  return getParentCallExpressionNode(node.parent);
+  return getClosestParentCallExpressionNode(node.parent);
+};
+
+/**
+ * Async assertions might be called in Promise
+ * methods like `Promise.x(expect1)` or `Promise.x([expect1, expect2])`.
+ * If that's the case, Promise node have to be awaited or returned.
+ *
+ * @Returns CallExpressionNode
+ */
+const getPromiseCallExpressionNode = node => {
+  if (
+    node &&
+    node.type === 'ArrayExpression' &&
+    node.parent &&
+    node.parent.type === 'CallExpression'
+  ) {
+    node = node.parent;
+  }
+
+  if (
+    node.type === 'CallExpression' &&
+    node.callee &&
+    node.callee.type === 'MemberExpression' &&
+    node.callee.object.name === 'Promise' &&
+    node.parent
+  ) {
+    return node;
+  }
+
+  return null;
 };
 
 const checkIfValidReturn = (parentCallExpressionNode, allowReturn = true) => {
@@ -28,8 +70,12 @@ const checkIfValidReturn = (parentCallExpressionNode, allowReturn = true) => {
   if (allowReturn) {
     validParentNodeTypes.push('ReturnStatement');
   }
-  return validParentNodeTypes.includes(parentCallExpressionNode.parent.type);
+
+  return validParentNodeTypes.includes(parentCallExpressionNode.type);
 };
+
+const promiseArrayExceptionKey = ({ start, end }) =>
+  `${start.line}:${start.column}-${end.line}:${end.column}`;
 
 module.exports = {
   meta: {
@@ -52,12 +98,28 @@ module.exports = {
           alwaysAwait: {
             type: 'boolean',
           },
+          ignoreInPromise: {
+            type: 'boolean',
+          },
         },
         additionalProperties: false,
       },
     ],
   },
   create(context) {
+    // Context state
+    const arrayExceptions = {};
+
+    const pushPromiseArrayException = loc => {
+      const key = promiseArrayExceptionKey(loc);
+      arrayExceptions[key] = true;
+    };
+
+    const promiseArrayExceptionExists = loc => {
+      const key = promiseArrayExceptionKey(loc);
+      return !!arrayExceptions[key];
+    };
+
     return {
       CallExpression(node) {
         // checking "expect()" arguments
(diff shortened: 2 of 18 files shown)


yatki:
@SimenB please take a look to this wip PR. After getting your feedback, I'll update the docs.

index.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
         'jest/no-jasmine-globals': 'warn',
         'jest/no-test-prefixes': 'error',
         'jest/valid-describe': 'error',
-        'jest/valid-expect': 'error',
+        'jest/valid-expect': ['error', { alwaysAwait: true }],
```
SimenB: it should default to `false` to avoid a breaking change 🙂

rules/valid-expect.js
```diff
@@ -5,23 +5,59 @@
...
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
   },
   create(context) {
     return {
       CallExpression(node) {
         const calleeName = node.callee.name;
```
SimenB: inline this?

yatki:
@SimenB addressed :)

rules/valid-expect.js
```diff
@@ -5,23 +5,57 @@
...
+};
+
+const checkIfValidReturn = (context, parentCallExpressionNode) => {
+  const validParentNodeTypes = ['ArrowFunctionExpression', 'AwaitExpression'];
+  const { options } = context;
+  if (!options[0] || !options[0].alwaysAwait) {
+    validParentNodeTypes.push('ReturnStatement');
+  }
+  return (
+    validParentNodeTypes.indexOf(parentCallExpressionNode.parent.type) > -1
```
jeysal: `includes`

rules/valid-expect.js
```diff
@@ -108,6 +142,26 @@ module.exports = {
...
+          expectResolvesCase(node) ||
+          expectRejectsCase(node) ||
+          expectNotResolvesCase(node) ||
+          expectNotRejectsCase(node)
+        ) {
+          const parentCallExpressionNode = getParentCallExpressionNode(node);
+          if (!checkIfValidReturn(context, parentCallExpressionNode)) {
+            const { options } = context;
+            const messageReturn =
+              !options[0] || !options[0].alwaysAwait ? ' or returned' : '';
```
jeysal: Extract `!options[0] || !options[0].alwaysAwait` to `allowReturn`? It's used in two places and not very readable

jeysal: (passing just the flag to `checkIfValidReturn` instead of `context`)

yatki: @jeysal good points, addressed.

rules/util.js
```diff
@@ -5,30 +5,39 @@ const { version } = require('../package.json');
 
 const REPO_URL = 'https://github.com/jest-community/eslint-plugin-jest';
 
-const expectCase = node =>
-  node.callee.name === 'expect' &&
-  node.arguments.length === 1 &&
-  node.parent &&
-  node.parent.type === 'MemberExpression' &&
-  node.parent.parent;
+const expectCase = node => node.callee.name === 'expect';
```
jeysal: After taking a closer look, I'm a bit worried that removing all the parent checks will break some cases (and some of the new code like `getParentCallExpression` also doesn't do any checks).
Are you sure these are no longer needed, and if so, mind explaining why this can't break?

yatki: I think the change is fine for several reasons:

1. All the tests are passing. I didn't check all test cases but if we can rely on the tests, the change should be fine.
2. I've looked into all `expectCase` usages, 8 files are using this function in total. And almost in all files, there are further checks after detecting if the node is an `expect` CallExpression. So, intiutively, I can say it looks ok.
3. Actually initial implementation was making sure that expect always gets minimum one argument and expect has a property like `expect(x).property`.  Thus it was missing `expect()` &  `expect(x)` statements. 
  The way I implemented is only checking if it's `expect()`. Rest is up to developers to check further.

Actually after taking another look, now I think we don't even need these checks:
https://github.com/yatki/eslint-plugin-jest/blob/3bafe954fbeb50168116e14a8f350352c806d705/rules/util.js#L12

Because all it does is to make sure there is another property coming after `.not`, `.resolves`, `.rejects` properties. I don't think we need them.

My only concern is during typing the code, linter may complain too early as soon as it detects `expectCase()`. But I couldn't see such case.

I'll try to test the rules manually when I have time tomorrow. Until then, if you notice anything weird, you can just let me know.

If you don't see any other problems, I can update the docs and mark the PR ready for review 🖖🏻

jeysal: Hey, I just ran this on Jest itself. 
There's one error that appears to be correct:
```
jest/packages/expect/src/__tests__/matchers.test.js
  1178:7  error  Async assertions must be awaited or returned  jest/valid-expect
```
Then there's four (really two) errors involving `Promise.all`:
```
jest/packages/jest-mock/src/__tests__/index.test.ts
  589:9  error  Async assertions must be awaited or returned  jest/valid-expect
  590:9  error  Async assertions must be awaited or returned  jest/valid-expect
  614:9  error  Async assertions must be awaited or returned  jest/valid-expect
  615:9  error  Async assertions must be awaited or returned  jest/valid-expect
```
I think those might just be edge cases suitable for `eslint-disable` though, you can't cover every form of waiting for promises anyway. @SimenB

jeysal: Regarding my comment about `parent`s:

As far as `expectCase` is concerned, I think you're likely right that it won't break as long as it's called with `CallExpression`s as a node because those will always have at least an `ExpressionStatement` and a `Program` above them. I'd personally still prefer to keep the parents check in because as a rule author I'd expect `expectCase` to be safe (never throw regardless of what node I give it). Removing the `arguments.length` check sounds good to me!

As far as `getParentCallExpression` is concerned, here's an example that crashes the rule: `expect().resolves.asdf;`
(A weird line of course, but still linting should never crash, only report errors)

yatki: @jeysal thank you very much for the feedbacks.

1. I noticed that it's possible to use `.not`, after `.resolves`. I'll make sure that it doesn't break anything.
  https://github.com/facebook/jest/blob/db5e472b51cfc68787c9da7bb21b26f192b0ab78/packages/expect/src/__tests__/matchers.test.js#L1179

2. Regarding having the assertion inside a Promise method, for `Promise.fn(expect())` or `Promise.fn([expect()])` usages, I can also check if Promise was awaited or returned. I think it's straightforward to solve for this case and we can save devs from some pain?

3. I'll look into expectCase function and add the necessary checks. I didn't notice it was crashing. 

Also, is there already a setup to test the plugin on an existing project? How did you run it on jest for instance? Is there some instructions to setup that? Thanks.

SimenB: `Promise.all` & `Promise.race`, and the new `Promise.allSettled` & `Promise.any` should all be supported and count as a new promise, unless implementing that is super complicated

yatki: @SimenB it's not complicated at all. I already did that just couldn't push it yet. I simply check for all `Promise.x` usages where x can be any of those methods. I will push it this weekend.

SimenB: Sounds perfect 🙂

jeysal: > Also, is there already a setup to test the plugin on an existing project? How did you run it on jest for instance?

1. `yarn link` in the `eslint-plugin-jest` dir
2. `yarn link eslint-plugin-jest` in the `jest` dir
3. `yarn lint` in the `jest` dir

:)

index.js
yatki: @SimenB After a second thought, since we are extending valid-expect capabilities, I think it would be more convenient to enable the options by default. Community have to upgrade their plugins to the minor version anyway, they will have this features enabled by default? Am I missing something?

yatki:
@SimenB @jeysal first of all, sorry for the delay. I couldn't find time to push this earlier.

Now we validate awaiting `Promise.x(expect)` and `Promise.x([expect1, expect2])` usages. I also introduced `ignoreInPromise` option to disable reporting errors when async asserts are in promise methods. I also tried to make sure nothing breaks when we type. 

Please take a look if you experience any problems and let me know what do you think about the latest changes. 

Cheers.

rules/__tests__/valid-expect.test.js
```diff
@@ -3,16 +3,85 @@
...
+  },
+});
 
 ruleTester.run('valid-expect', rule, {
   valid: [
     'expect("something").toEqual("else");',
     'expect(true).toBeDefined();',
     'expect([1, 2, 3]).toEqual([1, 2, 3]);',
     'expect(undefined).not.toBeDefined();',
-    'expect(Promise.resolve(2)).resolves.toBeDefined();',
```
jeysal: Hmm, why was this allowed in the first place? Can't think of a use case

index.js
```diff
@@ -31,7 +31,10 @@ module.exports = {
         'jest/no-jasmine-globals': 'warn',
         'jest/no-test-prefixes': 'error',
         'jest/valid-describe': 'error',
-        'jest/valid-expect': 'error',
+        'jest/valid-expect': [
+          'error',
+          { alwaysAwait: true, ignoreInPromise: true },
```
jeysal: If those are defaults for the preset, they should probably also be defaults for the rule?

jeysal: Also, might want to default `alwaysAwait` to `false` because
```js
test('1', () => expect(doStuff(1)).resolves.toEqual(11));
test('2', () => expect(doStuff(2)).resolves.toEqual(12));
```
probably has some use cases where a whole lot of tests look nicer as one liners, I imagine.

SimenB: Should default to `false` yes

SimenB:
@jeysal why is this breaking? Seems like a bug fix it it uncovers more cases

jeysal:
@SimenB 
Because of `alwaysAwait: true` in the preset (although that might be changed now) and because
```js
expect(Promise.resolve(2)).resolves.toBeDefined();
```
changed from valid to invalid

index.js
yatki: > If those are defaults for the preset, they should probably also be defaults for the rule?

@jeysal you are right. I think setting default values for the rule and removing them from the preset makes more sense, right? Also I prefer to keep alwaysAwait: `true` & ignoreInPromise: `false` by default. I think that would be the most desired behaviour in most cases. What do you think about that?

> Also, might want to default `alwaysAwait` to `false` because
> 
> ```js
> test('1', () => expect(doStuff(1)).resolves.toEqual(11));
> test('2', () => expect(doStuff(2)).resolves.toEqual(12));
> ```
> 
> probably has some use cases where a whole lot of tests look nicer as one liners, I imagine.

Actually having alwaysAwait enabled doesn't conflict with one liners. See: https://github.com/jest-community/eslint-plugin-jest/pull/255/files#diff-a737bbc15714dae2905b79ec37ea79cdR69 

alwaysAwait just forces to use `await` instead of `return`.

jeysal: Hmm, I agree that `await` is better than `return` in a test function with a block body, just not sure I'd impose this preference on everyone by default. Why `ignoreInPromise: false`, is it too lenient and thus likely to produce false negatives?

yatki: > Why ignoreInPromise: false, is it too lenient and thus likely to produce false negatives?

I think so. In most cases devs would return or await `Promise.all([expect1, expect])` statements. I rarely saw Promise.all(...).then(...) usages in the tests and even then reminding devs to return or await Promises (that contains async assertions) could be more useful in most cases. 

But of course this is very opinionated, just talking from my experience. I'm ok with either way, since rule is already providing options to have control over these behaviours.

jeysal: Oh wait, I think I got confused here - `ignoreInPromise` is the other way around, it's less strict if you set it to true. Maybe something like "allowPromiseMethods" would be a clearer option name?
Anyway, I think I'd favor `alwaysAwait: false, ignoreInPromise: false`. @SimenB should probably make the final call ^^

yatki: Yeap, just to clarify. `ignoreInPromise: true` disables warnings for async assertions that were used in Promise methods. So setting it to `false` will complain for the usages below.

```js
// ignoreInPromise: false
// Bad
test("valid-expect", () => { 
  Promise.all([
    expect(Promise.resolve(2)).not.resolves.toBeDefined(),
    expect(Promise.resolve(3)).not.resolves.toBeDefined(),
  ]); 
});
```

```js
// ignoreInPromise: false
// Good
test("valid-expect", async () => { 
  await Promise.all([
    expect(Promise.resolve(2)).not.resolves.toBeDefined(),
    expect(Promise.resolve(3)).not.resolves.toBeDefined(),
  ]); 
});
```

```js
// ignoreInPromise: true
// Good
test("valid-expect", () => { 
  Promise.all([
    expect(Promise.resolve(2)).not.resolves.toBeDefined(),
    expect(Promise.resolve(3)).not.resolves.toBeDefined(),
  ]); 
});
```
I think when it's documented name of the options are fine but I'm ok with renaming options, of course.

Lastly, seems like tests on master are broken? Is it only me?

yatki:
@SimenB @jeysal just wanted to follow up, if you were able to discuss & decide what to do about the remaining topics? So I can update the docs & conclude the PR this week when I have time ^^. 

- Renaming the options
- Default values of the options

Cheers.

jeysal:
Simen agreed with `alwaysAwait` defaulting to false, so that's safe. Also doubt he'd be against the rename. Looking at your third really makes me want to default `allowPromiseMethods` to `false`, since that is quite a severe issue to miss (in fact, I'm not sure why you'd want to enable it at all if the behavior without it already properly considers `await`/`return`).

yatki:
@jeysal I also feel like `allowPromiseMethods` sounds better. And I agree, I also wouldn't never enable `allowPromiseMethods` but since supporting Promise methods was an additional feature, i just wanted to provide a way to easily disable it just in case somebody wouldn't like to use it. 

I'll set `alwaysAwait` to false by default, rename the option `ignoreInPromise` to `allowPromiseMethods` and set it to false by default and update the docs asap.

yatki:
@jeysal @SimenB changes were addressed. I also updated the docs. Please feel free to edit the docs, if you want to add or correct something.

Looking forward for your feedback and to use this rule with the plugin 🖖

docs/rules/valid-expect.md
```diff
@@ -20,8 +20,97 @@ or when a matcher function was not called, e.g.:
...
+  await expect(Promise.resolve(1)).resolves.toBe(1);
+});
+
+test('test2', () => expect(Promise.resolve(2)).resolves.toBe(2));
+```
+
+### `allowPromiseMethods`
+
+When set to true, disables triggers on async assertions that were used inside a
+Promise method.
```
jeysal:
```suggestion
Promise method, even if not immediately `await`ed or returned.
```

yatki: @jeysal addressed.

src/rules/valid-expect.js
```diff
@@ -5,23 +5,118 @@
...
+      {
+        type: 'object',
+        properties: {
+          alwaysAwait: {
+            type: 'boolean',
+            default: false,
+          },
+          allowPromiseMethods: {
+            type: 'boolean',
+            default: false,
```
SimenB: should this default to true? I don't care about `Promise.resovle`, but doing `return expect(Promise.all([prom1, prom2])).resolves.toEqual([val1, val2])` seems perfectly valid to me

yatki: But that's a different case. It shouldn't trigger any warnings on that anyways (if alwaysAwait: false). 

allowPromiseMethods are just for checking if `expect().resolves` was used in Promise.all, not the other way around.

jeysal: I think you got confused by the same thing as me - with `false`, this will already be allowed. With `true`, *any* `Promise.all`s will be allowed, even if they are not awaited or returned. TBH I have no idea why anyone would set it to true and not just eslint-disable if there's ever a use case for that.
Seeing that it was not just confusing for me and that IMO it makes the examples look like documentation mistakes, maybe we should just remove the setting?

SimenB:
```js
Promise.all([
  expect(prom1).resolves.toEqual(val1),
  expect(prom2).resolves.toEqual(val2),
]);
```

should also always be valid

yatki: @jeysal if somebody would have test cases using `done()` like below then they may wanna disable that check.

```js
Promise.all([
  expect(prom1).resolves.toEqual(val1),
  expect(prom2).resolves.toEqual(val2),
]).then(() => {
// some other operations & assertions
done();
});
```

@SimenB that's also valid. There shouldn't be any warnings with the examples you shared.

SimenB: > With `true`, _any_ `Promise.all`s will be allowed, even if they are not awaited or returned. 

Wat, them seems really weird. Adding `Promise.all` (or `Promise.resolve`, albeit that's weird) around the `expect` should make absolutely zero difference to this rule - it is about correctness, not style. The reason I brought is up is that I want us to understand it's the wrapping promise we want to be `return`ed or `await`ed, not for us to have any opinion on whether it's used or not

SimenB: Ok, read the docs, and tests, more thoroughly, an it does indeed behave the way I want it to. 😀 

However, it only does so provided `allowPromiseMethods` is not set to `false`. I think we should kill that option, I see no use cases for it. Can you describe one?

yatki: > The reason I brought is up is that I want us to understand it's the wrapping promise we want to be returned or awaited, not for us to have any opinion on whether it's used or not

@SimenB I agree. However since promises can be fulfilled with `.then` statements as well and developers can use `done()` to conclude their tests. That's why somebody can want to have the rule in general but just want to disable it for async assertions that were wrapped with Promise methods. See my comments above as an example :)

SimenB: I think an `eslint-disable-next-line jest/valid-expect` is better in that case.

yatki: @jeysal @SimenB I removed the option. I agree that in most cases nobody will need it. 

If that case annoys lots of people solution is already here. Somebody can revert the last commit in future if we need that option :D

SimenB:
Thanks! I'll run this over some code bases later today and merge if all is good :)

@jeysal @thymikee do you agree this is not breaking?

thymikee:
@SimenB it's turned off by default, so I think we should be safe? Would be cool to turn it on in next major

yatki:
> @SimenB it's turned off by default, so I think we should be safe? Would be cool to turn it on in next major

@thymikee As far as I see, in recommended settings the rule is enabled, only `alwaysAwait` option is disabled but the behaviour of the `valid-expect` changed anyways.

@SimenB I don't think it is breaking, we just extended the behavior of `valid-expect`. We were not triggering warnings before on misused async assertions but now in lots of project, linting step on CI will fail because of new checks in the rule. So including this feature to a `minor` release should be fine.

yatki:
Just to clarify `alwaysAwait` only enforces to use `await` instead of returning Promises. even alwaysAwait is false, the rule still would trigger warning on 

```js
expect(Promise.resolve(2)).resolves.toBeDefined();
```

yatki:
@SimenB @jeysal is there any update on this PR? Can you estimate when this can be released? Is it going to be released? Cheers,

SimenB:
Sorry, forgot about this! Mind merging in master (I assume a rebase with 20 commits will be painful, and we squash during merge anyways)? Once you've done so, I can run it over some code bases and merge (I know I said so in May... This time I mean it!)

yatki:
@SimenB done. I also created messageIds for the messages I wrote! So, I hope this should be fine now.

SimenB:
Awesome, thanks!

>I didn't squash my commits, tho would you like me to do that too?

No need, I'll do that when merging.

---

I just left for the day and won't have access to a computer until tomorrow. I'll test this first thing tomorrow though :)

SimenB:
Finally got around to testing this, this is great! It actually found an error in Jest's own tests (ironically enough when testing async matchers): https://github.com/facebook/jest/blob/ccf90c005232bf9e57ee11023637e31b404a3520/packages/expect/src/__tests__/matchers.test.js#L1220-L1222

---

However, changes introduced here also crashed in another repo (not `valid-expect` though - `no-alias-methods`). Failing snippet:

```js
expect(foobar);
```

It's not a valid `expect` call (there's no expectation), but it explodes with `TypeError: Cannot read property 'name' of undefined`

(backstory - this `expect` is from cypress so we've turned off `valid-expect` for these files, which is why it slipped through)

yatki:
@SimenB Interesting, the tests were passing. I think the change I made in `expectCase` helper function is causing that error. I'm looking into it.

yatki:
@SimenB The errors are caused because of the changes I did on `expectCase`. I'm going to fix it but I don't have time for today. I'll do it tomorrow probably. Something to think on, would you prefer me to revert the changes on `expectCase` or fix the rules that use `expectCase`. 

I feel like the expectCase function is better as it is now in this PR. It's doing what it should be doing: It's validating if the statement is `expect()` statement. Checking if that node has arguments or properties should be done afterwards depending on the logic of the rule. So, I suggest fixing the rules that's using expectCase:

- src\rules\no-alias-methods.js
- src\rules\no-truthy-falsy.js
- src\rules\prefer-called-with.js
- src\rules\prefer-strict-equal.js
- src\rules\prefer-to-contain.js
- src\rules\prefer-to-have-length.js
- src\rules\require-tothrow-message.js

Let me know what do you think.

SimenB:
I agree. Potentially adding a separate util that verifies it's `expect` _and_ has matchers so we don't have to duplicate it across rules

yatki:
Ok then. I will do it that way. Btw, great job on catching that problem. Could cause lots of headaches 😅 🖖🏻

SimenB:
I'll be landing #302 btw, which'll probably give you a few conflicts. Shouldn't be too bad, though 🙂

SimenB:
@yatki I merged in master, but this now has a failing test. Would you be able to take a look?

image

I'll merge as soon as it's green

(feel free to reapply the merge or rebase if you want, but I _think_ I merged correctly)

SimenB:
Finally! 🎉 Thank you for a mind-boggling awesome contribution @yatki!

If you feel up for it, I'd love your help converting this rule to TypeScript (#256) 🙂 I feel like you know this rule best by now 😛

yatki:
Wow, that's great :D I immediately updated the plugin to the latest version and seeing it working feels great :)

@SimenB thank you very much for the support and constructive feedback. It was a pleasure.

Regarding converting it to TS: https://github.com/jest-community/eslint-plugin-jest/pull/333 It's wip at the moment but I'll complete it when i have time this week.

Issue #256, opened 2019-05-10
Rewrite in TypeScript


SimenB:
Now that https://github.com/typescript-eslint/typescript-eslint/pull/425 has been released, it should be relatively straightforward to migrate.

I think porting one and one rule makes the most sense. Feel free to pick one 🙂 

- [x] `consistent-test-it` - #327
- [x] `expect-expect` - #325
- [x] `lowercase-name` - #258
- [ ] `no-alias-methods`
- [x] `no-commented-out-tests` - #305
- [x] `no-disabled-tests` - #315
- [x] `no-duplicate-hooks` - #318
- [ ] `no-empty-title`
- [x] `no-export` - #323
- [x] `no-focused-tests` - #314
- [x] `no-hooks` - #322
- [ ] `no-identical-title`
- [x] `no-if` - #324
- [x] `no-jasmine-globals` - #315
- [x] `no-jest-import` - #259
- [ ] `no-large-snapshots`
- [x] `no-mocks-import` - #309
- [x] `no-test-callback` - #321
- [x] `no-test-prefixes` - #328
- [x] `no-test-return-statement` - #320
- [ ] `no-truthy-falsy`
- [ ] `prefer-called-with`
- [ ] `prefer-expect-assertions`
- [x] `prefer-inline-snapshots` - #319
- [x] `prefer-spy-on` - #326
- [x] `prefer-strict-equal` - #329
- [ ] `prefer-to-be-null`
- [ ] `prefer-to-be-undefined`
- [ ] `prefer-to-contain`
- [ ] `prefer-to-have-length`
- [x] `prefer-todo` - #335
- [ ] `require-tothrow-message`
- [x] `valid-describe` - #308
- [ ] `valid-expect-in-promise`
- [ ] `valid-expect` - #333

G-Rath:
I've done a first pass of (`no-mocks-import`](https://github.com/G-Rath/eslint-plugin-jest/tree/ts-migration/no-mocks-import), which passes all the tests.

So far, it seems to me like this'll be a bit teeth pully, as `ts-estree` doesn't provide you with lots of nice [type guards like TS does](https://github.com/microsoft/TypeScript/blob/master/src/compiler/utilities.ts#L5392) - as a result, I've started to make my own in `tsUtils`, but am trying to be conservative for now, since I don't know `ts-estree` well enough to predict how wild things might get.

[...]

PR #257, opened 2019-05-10
chore: precompile with babel

SimenB:
Prerequisite for #256. This has no TS yet, but sets up babel for jest and publishing.

This also allows us to use newer syntax while maintaining node 6 compat


.eslintrc.js
```diff
@@ -32,6 +32,8 @@ module.exports = {
       },
     ],
     'prettier/prettier': 'error',
+    'node/no-unsupported-features/es-syntax': 'off',
```
SimenB: babel deals with syntax

package.json
```diff
@@ -16,24 +16,31 @@
...
   ],
+  "main": "lib/",
   "engines": {
     "node": ">=6"
   },
   "peerDependencies": {
     "eslint": ">=5"
   },
   "scripts": {
+    "postinstall": "yarn build",
```
jeysal: Is auto-build after install helpful in this case?
Unlike in the Jest repo, you'll be able to run `yarn jest` just fine.
In development, `yarn build --watch` will initially build everything anyway (afaik that's what Babel CLI `watch` does)

babel.config.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = {
+  presets: [['@babel/preset-env', { targets: { node: 6 } }]],
```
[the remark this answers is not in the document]
SimenB: No need to make a breaking change, exactly 0 benefit

package.json
SimenB: Lint will fail since we run with our own rules. And jest runs lint, so it'd fail

jeysal: Okay, didn't consider that 👍

babel.config.js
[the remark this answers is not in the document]
SimenB: whenever we wanna make a new major we can drop 6 though, no reason to keep it at that point

package.json
ranyitz: @SimenB `postInstall` runs when this package is being `npm/yarn` installed.

We can't assume that the user has `yarn` or any of the `devDependencies`.

I think we should revert it for now.

ranyitz: Maybe the intention was `prepublishOnly`?

PR #258, opened 2019-05-10
chore: port lowercase-name to TypeScript

SimenB:
For #256. Includes #257, so just review the second commit.

The diff looks bigger than it is since I had to rewrite all the helpers to include type guards. The implementation of the rule itself is identical.

I can extract the changes to `index` etc to a separate PR if wanted so this is focused on _just_ the rule porting

@jeysal any help here would be greatly appreciated 🙂


src/rules/__tests__/lowercase-name.test.ts
```diff
@@ -1,9 +1,11 @@
-'use strict';
+import { RuleTester as ESLintRuleTester } from 'eslint';
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+import rule from '../lowercase-name';
 
-const { RuleTester } = require('eslint');
-const rule = require('../lowercase-name');
+const RuleTester: TSESLint.RuleTester = ESLintRuleTester as any;
```
SimenB: https://github.com/typescript-eslint/typescript-eslint/blob/eb613ca451624c3f9db280cab195f2a565d5de88/packages/eslint-plugin/tests/RuleTester.ts#L5

src/rules/__tests__/lowercase-name.test.ts
```diff
@@ -1,9 +1,11 @@
...
 
 const ruleTester = new RuleTester({
+  parser: '@typescript-eslint/parser',
```
SimenB: I find it odd they require this - even if the rule is written in TS, it might not necessarily only run with that parser. Might be needed for all I know 🤷‍♂

src/rules/lowercase-name.ts
```diff
@@ -0,0 +1,129 @@
...
+): expression is TSESTree.Literal | TSESTree.TemplateLiteral =>
+  !!expression &&
+  (expression.type === AST_NODE_TYPES.Literal ||
+    expression.type === AST_NODE_TYPES.TemplateLiteral);
+
+const testDescription = (
+  firstArgument: TSESTree.Literal | TSESTree.TemplateLiteral,
+) => {
+  if (firstArgument.type === AST_NODE_TYPES.Literal) {
+    return firstArgument.value as string;
```
jeysal: Doesn't this hint at an actual error if it's a different kind of literal, e.g. `42`?

src/rules/lowercase-name.ts
```diff
@@ -0,0 +1,129 @@
...
+  create(context, [{ ignore }]) {
+    const ignoredFunctionNames = ignore.reduce<
+      Record<string, true | undefined>
+    >((accumulator, value) => {
+      accumulator[value] = true;
+      return accumulator;
+    }, Object.create(null));
+
+    const isIgnoredFunctionName = (node: TSESTree.CallExpression) =>
+      ignoredFunctionNames[(node.callee as TSESTree.Identifier).name];
```
jeysal: Would
```ts
isIdentifier(node.callee) && ignoredFunctionNames[node.callee.name]`
```
work?

src/rules/lowercase-name.ts
```diff
@@ -0,0 +1,129 @@
...
+  AST_NODE_TYPES,
+} from '@typescript-eslint/experimental-utils';
+
+const isIdentifier = (
+  callee: TSESTree.LeftHandSideExpression,
+): callee is TSESTree.Identifier => !!callee;
+
+const isItTestOrDescribeFunction = (
+  callee: TSESTree.LeftHandSideExpression,
+): callee is TSESTree.Identifier => {
```
jeysal: Doesn't matter that much in this case, but maybe it helps someone learn more about TS: This is technically an incorrect type guard, because the function returns true not for any `TSESTree.Identifier`, but only for specific ones (those with certain names). This matters as soon as you negate the guard (`!isItTestOrDescribeFunction`) - consider the following code:
```ts
const node: Identifier | Literal = someIdentifier;
if (!isItTestOrDescribeFunction) {
  node; // Inferred as Literal
  node.value; // Typechecks even though it's really an identifier that has no `value`
}
```
Hope that this explanation is reasonably easy to understand :)

In this case I'd recommend removing the type check and making the signature
```ts
isItTestOrDescribeFunction = (callee: Identifier): boolean => callee.name // === ...
```
so callers have to do `isIdentifier(callee) && isItTestOrDescribeFunction(callee)`.

P.S. If you *really* need to type something like this correctly, I did it [here](https://github.com/facebook/jest/blob/15e2cc68150e5a67cc9f4b18995954696c9bbf59/packages/jest-config/src/utils.ts#L233-L238), but it's a bit arcane :D

SimenB: Yeah, I probably butchered the implementation here 😅 mostly kept the helpers the same, but trying to make tsc shut up. Super happy to make them better if you wanna give it a whack :)

src/rules/lowercase-name.ts
SimenB: Hmm, maybe. We don't have a test for that? Probably not.

src/rules/lowercase-name.ts
SimenB: Probably. Unnecessary check though as we know it's true 100% of the time already, we just haven't convinced the compiler of this fact.

It also doesn't matter to the rule since it'd return undefined. Would of course be better to have the type checker happy without casts or duplicated code 😀

SimenB:
@jeysal thanks for the review! I reimplemented this now with what I think is a better approach, with my own internal types that are more specific, and type guarding for them. I like this approach, could you take a look?

src/rules/lowercase-name.ts
SimenB: You were right. I've added a test and removed the cast

src/rules/lowercase-name.ts
SimenB: I've redone it now to what I _think_ are valid type guards

jeysal:
Typings look nice! 😅 (hell yeah, statically verify everything 💪)
The one missing line of coverage looks like it could be covered by having a callee that is not an identifier (e.g. member expression `a.b()`)

SimenB:
Thanks @jeysal!

PR #259, opened 2019-05-12
chore: migrate no-jest-import to typescript

SimenB:
Quite easy when the rule's declarative 😅


src/rules/__tests__/no-jest-import.test.ts
```diff
@@ -0,0 +1,86 @@
...
+const RuleTester: TSESLint.RuleTester = ESLintRuleTester as any;
+
+const ruleTester = new RuleTester({
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    ecmaVersion: 6,
+  },
+});
+
+const moduleSourceType = { sourceType: 'module' } as const;
```
SimenB: can ditch all `as const` after https://github.com/typescript-eslint/typescript-eslint/pull/517 is released

PR #262, opened 2019-05-16
feat(rules): add no-commented-out rule

### LINKED ISSUE #261: New rule: no-commented-out-tests
There's already a rule to catch disabled tests. In our codebase, I'm seeing cases where people disable tests by commenting them out. While this could be easily caught during code reviews, it'd be nice to disallow this via a rule before it even makes it to the review.

Implementation-wise, we could probably detect cases of `describe`/`it`/etc. within comments and warn then.

kangax:
Closes #261


src/rules/no-commented-tests.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const { getDocsUrl } = require('./util');
+
+const message = 'Some tests seem to be commented';
+
+function hasAssertions(node) {
+  return /x?(test|it|describe)((\.skip|\[['"]skip['"]\]))?\(/.test(node.value);
```
kangax: I didn't bother with matching quotes here. I think this will be rare enough and the rule is fuzzy as it is.

src/rules/no-commented-tests.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const { getDocsUrl } = require('./util');
+
+const message = 'Some tests seem to be commented';
+
+function hasAssertions(node) {
```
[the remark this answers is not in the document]
kangax: hm, perhaps `hasTests`?

src/rules/no-commented-tests.js
```diff
@@ -0,0 +1,39 @@
...
+  return /x?(test|it|describe)((\.only|\.skip|\[['"]skip['"]\]))?\(/.test(
```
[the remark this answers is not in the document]
kangax: Ok, I'll make it match any word (optionally). Looks like we also missed `each`:

Screen Shot 2019-05-16 at 17 15 42

src/rules/__tests__/no-commented-tests.test.js
```diff
@@ -0,0 +1,161 @@
...
+    'it.only("foo", function () {})',
+    'test("foo", function () {})',
+    'test.only("foo", function () {})',
+    'var appliedSkip = describe.skip; appliedSkip.apply(describe)',
+    'var calledSkip = it.skip; calledSkip.call(it)',
+    '({ f: function () {} }).f()',
+    '(a || b).f()',
+    'itHappensToStartWithIt()',
+    'testSomething()',
+    [
```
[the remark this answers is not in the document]
kangax: This was just a copy from no-disabled-tests. Would you still want these updated to template literals?

kangax:
I agree that `no-commented-out-tests` would probably more of a common naming

src/rules/__tests__/no-commented-tests.test.js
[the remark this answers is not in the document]
jeysal: Doesn't matter that much. Maybe it was done to avoid deep indentation causing high col numbers

src/rules/no-commented-tests.js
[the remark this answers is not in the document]
kangax: Updated to match any word after

src/rules/no-commented-out-tests.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const { getDocsUrl } = require('./util');
+
+const message = 'Some tests seem to be commented';
+
+function hasTests(node) {
+  return /x?(test|it|describe)(\.\w+|\[['"]\w+['"]\])?\(.*?\)/.test(node.value);
```
jeysal: * Other than `xit`, there's also `fit`
* The closing parenthesis might cause trouble with multi-line tests in comments, we might need the `//s` regex modifier? Should add some multiline test cases

src/index.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
       rules: {
         'jest/no-alias-methods': 'warn',
         'jest/no-disabled-tests': 'warn',
+        'jest/no-commented-out-tests': 'warn',
```
SimenB: changing the recommended config is a breaking change.

Not sure this needs to be in the recommended config, at all, but we can revisit that before the next release 🙂

kangax: Ah, I didn't realise this will modify recommended config. On the other hand, I think it's ok to warn since warnings usually don't fail with tools like lint-staged, etc.

src/rules/no-commented-out-tests.js
kangax: Addressed both comments and added tests. The thing about newlines is that ESLint seems to create a separate comment for each line with line-level comments. We can't detect stuff like:

```js
// test
// (
// 'x
// )
```

..which, I think, is an OK limitation?

jeysal: For block comments as well though?

jeysal: Actually, best thing might be to just not look for the closing parenthesis? `// test(` already looks suspicious enough

kangax: Yep, I did just that in the last commit. Let me add a multiline comment test as well.

kangax:
@SimenB anything else you want me to do here?

src/index.js
SimenB: People can run eslint with a flag failing on warnings

kangax: removed it

mrtnzlml:
Hi @kangax! Here I am again... 😊 

I ran this rule on our codebase [just like I did here](https://github.com/gajus/eslint-plugin-flowtype/pull/400#issuecomment-492828739) and I found some cases where this rule reports false positives. You may want to check it so people don't have to disable this rule by default just like we did. All these cases are reporting issues incorrectly (+ none of them is a test file):

```text
// TODO: find out better way how to do it (type should be just an internal detail - see evaluateGlobalIdField)
```

```text
// TODO: turn this into regular Jest test (?)
```

```text
// TODO: unify with Git implementation from Shipit (?)
```

And also [this one](https://github.com/kiwicom/monorepo-utils/blob/4324189b587a75ff434debc1456899f9e1b04213/src/getChangedFiles.js#L8-L27) - reports issue on line 8 incorrectly. Cheers 🍻

kangax:
@mrtnzlml yep, I noticed some false positives in our codebase too just now. This should fix them — https://github.com/jest-community/eslint-plugin-jest/pull/271

PR #265, opened 2019-05-22
Do not run `yarn build` during postinstall hook

### LINKED ISSUE #263: Upgrade from 22.5.1 to 22.6.0 fails with error `/bin/sh: 1: babel: not found`
After upgrading to 22.6.0, we get the following build error when installing eslint-plugin-jest:

```bash
error /root/project/bootstrap-vue/node_modules/eslint-plugin-jest: Command failed.
Exit code: 1
Command: yarn build
Arguments: 
Directory: /root/project/bootstrap-vue/node_modules/eslint-plugin-jest
Output:
yarn run v1.13.0
$ babel --extensions .js,.ts src --out-dir lib
/bin/sh: 1: babel: not found
error Command failed with exit code 127.
```

Fails due to the `postinstall` command introduced via #257 which tries to compile files with babel.

ranyitz:
Maybe I'm missing something, but running build during `postInstall` doesn't seem to be the right way to go.

This PR moves it into `pretest` instead.

Fixes #263

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "eslint": ">=5"
   },
   "scripts": {
-    "postinstall": "yarn build",
     "lint": "eslint . --ignore-pattern '!.eslintrc.js' --ext js,ts",
     "prettylint": "prettylint docs/**/*.md README.md package.json",
     "prepublishOnly": "yarn build",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "eslint": ">=5"
   },
   "scripts": {
+    "pretest": "yarn build",
     "lint": "eslint . --ignore-pattern '!.eslintrc.js' --ext js,ts",
     "prettylint": "prettylint docs/**/*.md README.md package.json",
     "prepublishOnly": "yarn build",


SimenB:
Thanks, and sorry about this!

I don't wanna build before every test run, but I'll figure something out :)

swissspidy:
I thought that's what 9f40d2cf0ceef7d127af0c4160c58e022dcd63c6 should do?

SimenB:
I _don't_ want before every test run.

I'll probably add build to ci, and throw explicit error from eslintrc. But I'm currently in line to board a plane to Paris for react-europe, so this seems like a great quickfix in the meantime :)

swissspidy:
Oh, sorry, I misread. Thanks for clarification! Enjoy the conference 🎉

ranyitz:
@SimenB I agree that this is not a good solution for the long term. Thanks for merging :)

SimenB:
I think using `prepare` is the correct lifecycle hook, right?

ranyitz:
@SimenB It may be, but it still doesn't solve the local development problem.

I think we should go with adding a `yarn build` step in our [travis configuration](https://github.com/jest-community/eslint-plugin-jest/blob/master/.travis.yml#L17) (for CI flow)

And for local development, we should create a `watch` script that runs `babel`/`tsc` on every change. prepare would run the build locally after the first install but it won't be much fun running `yarn build` after each change. (while running `jest --watch` on a different terminal)

What do you think?

PR #278, opened 2019-06-10
chore: use messageId in all rules

### LINKED ISSUE #201: Use `messageId` in rules
This is a feature of newer verisons of eslint: https://eslint.org/docs/developer-guide/working-with-rules#messageids

Should be just a simple internal refactor, without impacting end users

SimenB:
Fixes #201

The TypeScript util thing forces us to use `messageId` anyways, so I thought to just do it now to reduce the diff when migrating to TS


SimenB:
I did the rules first, so the old tests which used `message` verified it spit the same message out the other side. Then I did the tests. So fairly certain this is correct 🙂

PR #282, opened 2019-06-21
feat(to-match-snapshot): show no-large-snapshots in js file

benmonro:
BREAKING CHANGE: this change will no longer report errors in the snap file, but instead report them on `.toMatchSnapshot()` calls instead.

Currently, when the `no-large-snapshots` rule runs, it highlights the problem in the .snap file.  The problem is you can't use `eslint-disable-next-line` in those files as they are generated and will get wiped out whenever `-u` is used.  This PR instead match the snapshot with the `toMatchSnapshot()` call thus allowing eslint-disable-next-line.  This is very useful for teams like ours which have a large number of large snapshots that we need to set to disable, while still blocking new large snapshots going forward.

The change here will still parse the snap files, but will keep track of them and align them with the test that is calling toMatchSnapshot.

### DIFF
--- a/src/rules/__tests__/no-large-snapshots.test.js
+++ b/src/rules/__tests__/no-large-snapshots.test.js
@@ -2,6 +2,7 @@
 
 const { RuleTester } = require('eslint');
 const rule = require('../no-large-snapshots');
+
 const noLargeSnapshots = rule.create;
 
 const ruleTester = new RuleTester({
@@ -78,7 +79,51 @@ describe('no-large-snapshots', () => {
     });
   });
 
-  describe('ExpressionStatement function', () => {
+  describe('toMatchSnapshot expression statement and call expressions', () => {
+    const testNode = {
+      type: 'CallExpression',
+      callee: {
+        type: 'Identifier',
+        name: 'it',
+      },
+      arguments: [
+        {
+          type: 'Literal',
+          value: 'a big component',
+          raw: "'a big component'",
+        },
+      ],
+    };
+    const expectNode = {
+      type: 'CallExpression',
+      callee: {
+        type: 'MemberExpression',
+        object: {
+          type: 'CallExpression',
+          callee: {
+            type: 'Identifier',
+            name: 'expect',
+          },
+          arguments: [
+            {
+              type: 'Identifier',
+              name: 'something',
+            },
+          ],
+        },
+        property: {
+          type: 'Identifier',
+          name: 'toMatchSnapshot',
+        },
+      },
+    };
+
+    afterEach(() => {
+      const mockContext = {
+        getFilename: () => 'mock-component.test.js',
+      };
+      noLargeSnapshots(mockContext)['CallExpression:exit'](testNode);
+    });
     it('should report if node has more than 50 lines of code and no sizeThreshold option is passed', () => {
       const mockReport = jest.fn();
       const mockContext = {
@@ -86,7 +131,25 @@ describe('no-large-snapshots', () => {
         options: [],
         report: mockReport,
       };
-      const mockNode = {
+
+      const snapshotNode = {
+        type: 'ExpressionStatement',
+        expression: {
+          left: {
+            property: {
+              type: 'TemplateLiteral',
+              quasis: [
+                {
+                  type: 'TemplateElement',
+                  value: {
+                    raw: 'a big component 1',
+                    cooked: 'a big component 1',
+                  },
+                },
+              ],
+            },
+          },
+        },
         loc: {
           start: {
             line: 1,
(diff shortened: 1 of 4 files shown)


benmonro:
@SimenB can we get this merged?  It is a breaking change since it's now reporting the long snapshot in the .js file instead of the .snap file.

benmonro:
@SimenB pretty please?  :) I think teams that have a large number of large snapshots are probably experiencing the same problem as me.  They want to block new large snapshots going forward, but they have a large quantity of them that they need to grandfather in using `//eslint-disable-next-line` which isn't pragmatic inside a .snap file.

src/rules/no-large-snapshots.js
```diff
@@ -35,19 +89,32 @@ module.exports = {
...
-          const propertyName =
-            node.callee.property && node.callee.property.name;
-          if (
-            propertyName === 'toMatchInlineSnapshot' ||
-            propertyName === 'toThrowErrorMatchingInlineSnapshot'
-          ) {
-            reportOnViolation(context, node);
+          if (isTestCase(node) || isDescribe(node)) {
+            const [firstArgument] = node.arguments;
+            titles.push(getStringValue(firstArgument));
```
jeysal: Does this work for
```js
const x = 42;
test(`${x}`, () => {});
```
?

src/rules/no-large-snapshots.js
```diff
@@ -19,6 +48,31 @@ const reportOnViolation = (context, node) => {
     });
   }
 };
+const captureToMatchSnapshot = (context, node) => {
+  const snapshotName = `${titles.join(' ')} ${++snapshotCall}`; //getSnapshotName(context, node);
```
jeysal: I think you're counting up within a test file here, while Jest counts up within a test case?

src/rules/__tests__/no-large-snapshots.test.js
```diff
@@ -78,15 +79,77 @@ describe('no-large-snapshots', () => {
     });
   });
 
-  describe('ExpressionStatement function', () => {
+  describe('toMatchSnapshot expression statement and call expressions', () => {
+    const testNode = {
```
jeysal: Can we maybe use ESLint's parser here to write these nodes as source code template literals in the tests instead? Would keep the file from exploding in size

src/rules/no-large-snapshots.js
jeysal: Nevermind just noticed the exit visitor. Should probably test both multiple snapshots in a test case and multiple snapshots in different test cases though

src/rules/__tests__/no-large-snapshots.test.js
jeysal: Would also make it a lot easier to see if we have all cases we can think of covered with tests

src/rules/no-large-snapshots.js
benmonro: good question, I will look for that and test it.

benmonro:
@jeysal thank you for the feedback, i will make sure all those things are addressed.  Sorry for being impatient, we'd like to roll this into our code ASAP so would like to get it merged soon.  we are bleeding large snapshots and i'd like to stop the hemorrhaging. :)

src/rules/no-large-snapshots.js
benmonro: @jeysal so the short answer is no.  The long answer is that i have a couple options.  One is to treat variables in a template literal as regular expressions when trying to resolve which snapshot they are associated with (not great). Another is to fallback to reporting the error in the snap for (also not great). Curious if you have any other thoughts.  Or if you know of a way to resolve the variables in the template literal so that the whole strong value can be reconstructed

src/rules/no-large-snapshots.js
benmonro: @jeysal added a test w/ 2 large snapshots in 2 separate tests.  resolving this, but feel free to re-open it below if you see any issues.

src/rules/__tests__/no-large-snapshots.test.js
benmonro: @jeysal added a test for snapshot in a describe, and multiple tests w/ mulitple snaps.  let me know if you can think of others.

SimenB:
@anescobar1991 what do you think about these changes?

benmonro:
closing this PR since we were able to accomplish our goal of whitelisting existing big snapshots using https://github.com/jest-community/eslint-plugin-jest/pull/288 it _is_ an interesting challenge though, just one that I don't have time to solve. :(

SimenB:
I think this is an improvement - you want the error displayed where it's actionable, which is by the `toMatchSnapshot` call.

Are we missing something to keep us from landing this?

benmonro:
@SimenB there's some big challenges around template literals with variables in them (ex describe.each). I'm not really convinced that those problems can be resolved without some consistent way of associating a test to a snapshot and Visa versa.  Will re open but the examples that @jeysal brought up highlight the issue.

j10sanders:
Just adding my support to this proposal.  Unfortunately, I don't even see the error raised on the snap file at this point.

G-Rath:
I'm going to close this for now as it seems to have gotten stale, and it sounds like there's no clear golden path yet on how to solve this issue.

PR #288, opened 2019-06-27
feat(whitelist-snapshots): option for whitelisting snapshots

benmonro:
This pull request adds an option to whitelist snapshots.  Since .snap files are generated, if you try to use `//eslint-disable-next-line` in a .snap file, the next time any snapshot is updated in that file, all of those comments will be removed.  For codebases with a large number of large snapshots, it is useful to be able to grandfather in existing large snapshots, while at the same time not allowing new large snapshots into the code.


benmonro:
@jeysal @SimenB This option is much simpler than trying to associate the .snap back to the test.  Thoughts? :)

benmonro:
Can someone merge this?  Or at least give feedback?

src/rules/no-large-snapshots.js
```diff
@@ -10,8 +10,29 @@ const reportOnViolation = (context, node) => {
...
+    context.options[0].whitelistedSnapshots;
+
+  let isWhitelisted = false;
+  if (whitelistedSnapshots) {
+    const fileName = path.relative(process.cwd(), context.getFilename());
+    const whitelistedSnapshotsInFile = whitelistedSnapshots[fileName];
+    if (whitelistedSnapshotsInFile) {
+      const snapshotName = getStringValue(node.expression.left.property);
+      isWhitelisted =
+        whitelistedSnapshotsInFile.find(name => {
```
SimenB: use `whitelistedSnapshotsInFile.some(...` instead and avoid the `!=null` at the end

src/rules/no-large-snapshots.js
```diff
@@ -10,8 +10,29 @@ const reportOnViolation = (context, node) => {
...
   const endLine = node.loc.end.line;
   const lineCount = endLine - startLine;
+  const whitelistedSnapshots =
+    context.options &&
+    context.options[0] &&
+    context.options[0].whitelistedSnapshots;
+
+  let isWhitelisted = false;
+  if (whitelistedSnapshots) {
+    const fileName = path.relative(process.cwd(), context.getFilename());
```
SimenB: this seems risky and potentially surprising - could we ask people to provide absolute paths instead? They'd have to use `.eslintrc.js` but I think that's fine. It's only js files mentioned in the eslint@6 migration guide: https://eslint.org/docs/6.0.0/user-guide/migrating-to-6.0.0. They'd then do `require.resolve('./path-to-my-file')` or something like it.

The reason I want thit is that I'd expect the relative path provided in config to be relative to the configuration file, not to `cwd`

src/rules/no-large-snapshots.js
```diff
@@ -10,8 +10,29 @@ const reportOnViolation = (context, node) => {
...
+            return name.test(snapshotName);
+          } else {
+            return name === snapshotName;
+          }
+        }) != null;
+    }
+  }
 
-  if (lineCount > lineLimit) {
+  if (!isWhitelisted && lineCount > lineLimit) {
```
SimenB: won't this ignore all long snapshots in the file, not the specific one from config (`snapshot name 1` in your example)? `isWhitelisted` will be `true` if the name is found in the file, regardless of which snapshot it is

SimenB:
Following #291, also please update the schema for this rule to include the option 🙂

src/rules/no-large-snapshots.js
benmonro: @SimenB absolsute paths work because we can’t check that into source control.  I’m good with doing it relative to the config file, is there some mechanism I can use to get that directory?

src/rules/no-large-snapshots.js
benmonro: no, `isWhiteListed` specific to each test/snapshot, not the entire file.  The tests I wrote should have all these bases covered, and when I run it locally against our codebase it works as expected.  Which is that the long snaps are whitelisted in each .snap file (but not the whole file).  The reason for having the .snap file in the whitelisting, is to avoid name collisions.  Suppose you have 2 different snap files with the same test name.  Keying off the path solves that problem.

src/rules/no-large-snapshots.js
SimenB: People can use js to calculate absolute paths within the config

benmonro: @SimenB what do you think of a special property for the root directory, so the root dir can be specified once, rather than in each file name?  so something like this:

```javascript
{
  whitelistedSnapshots: {
    "__root":"/absolute/path/to/root", //defaults to process.cwd()
    "relative/path/to/snapshot.snap": ["test1", "test2"],
    "relative/path/to/snapshot2.snap": ["test1", "test3"],
  }

```

That would be much better logistically for teams like us with hundreds and hundreds of large snapshots that we need to grandfather in.

src/rules/no-large-snapshots.js
benmonro: @SimenB  in other words `isWhitelisted` is only true when the file AND snapshot name match.  not just the filename.  I may be over explaining it, but does that make sense? 😄

benmonro: There's actually a test I wrote that covers this case: https://github.com/benmonro/eslint-plugin-jest/blob/feature/white-list-snapshots/src/rules/__tests__/no-large-snapshots.test.js#L220-L248

the title is about regex, but if you notice, it's got 2 large snapshots in the same file, but only reports on the non-whitelisted snapshot.

src/rules/no-large-snapshots.js
jeysal: We have `<rootDir>` in Jest and IMO the UX is mediocre. I don't think it's worth it to introduce a concept that people need to understand over using standard `path` mechanics.
Aside from the lack of support for JSON config, which may or may not be considered a big deal, it should be pretty easy to do something like
```js
const base = path.resolve(__dirname, 'src');
const whitelist = [path.resolve(base, 'file1.js'), path.resolve(base, 'file2.js')];
```
Could even introduce another function that applies the `base` argument already.

jeysal: I don't think banning relative paths outright is strictly necessary, but I also can't think of a good base directory (cwd might vary, config file location makes no sense if it's in a `config` directory) so idk if we have another option

benmonro: @jeysal the reason for doing this is because we have hundreds maybe even thousands of large snapshots that we need to whitelist so actually it would be quite painful.

jeysal: As soon as you make a helper function, possibly even one that takes multiple arguments and resolves them all from the base (`snapshots('1.js', '2.js')`), I don't see how it is more painful though?
On a side note, maybe for your scale you might want to make large snapshots follow a pattern that can be matched on or something?

src/rules/no-large-snapshots.js
benmonro: done

src/rules/no-large-snapshots.js
benmonro: @jeysal @SimenB ok i went ahead and got rid of the `process.cwd()` and updated the docs as well.  it will now assume an absolute path.

benmonro:
@SimenB @jeysal I have addressed all of the requests in the review, can we merge this now?  really want to try and stop the hemorrhaging of large snapshots on our team. :)

jeysal:
@benmonro looks like you have some leftover code ("jestx") in there, might want to remove that so it's ready when Simen has time for a final review and merge :)

benmonro:
@jeysal thanks, good catch, merged a branch i didn't mean to.  fixed.

benmonro:
Help me @SimenB you're my only hope! 
image

benmonro:
@SimenB I noticed some merges that went in, but this PR got skipped.  is there something I'm missing that you'd like to see included here?  We really could use this feature.  please advise.

docs/rules/no-large-snapshots.md
```diff
@@ -110,3 +110,27 @@ In an `eslintrc` file:
...
+In an `.eslintrc.js` file:
+
+```javascript
+...
+
+  "rules": {
+    "jest/no-large-snapshots": ["error",
+      {
+        "whitelistedSnapshots": {
+          "/path/to/file.js.snap": ["snapshot name 1", /a big snapshot \d+/]
```
SimenB: should we note somewhere in the docs people can use `path.resolve(__dirname, './relative-path')` or `require.resolve('./relative-path')` to get absolute paths that work between computers?

src/rules/no-large-snapshots.js
```diff
@@ -10,8 +10,30 @@ const reportOnViolation = (context, node) => {
   const startLine = node.loc.start.line;
   const endLine = node.loc.end.line;
   const lineCount = endLine - startLine;
+  const whitelistedSnapshots =
```
SimenB: we could verify all paths are absolute.

```js
const path = require('path');

const allPathsAreAbsolute = (whitelistedSnapshots || []).every(path.isAbsolute);

if (!allPathsAreAbsolute) {
  throw new Error('All paths must be absolute. You can use JS config and `path.resolve`');
}
```

That way it might be easier for users to use this rule correctly

benmonro:
@SimenB Thanks for the approve! :) . Can someone please merge this now?

docs/rules/no-large-snapshots.md
benmonro: done.

src/rules/no-large-snapshots.js
benmonro: done

benmonro:
@SimenB @jeysal ok guys, I have addressed everything, including Simen's latest comments. Now that Simen approved, can we please please please please merge this?  I've been very patient.  In the meantime my team is writing large snapshots and we need this desperately.  please merge.  pretty please... with cherries on top. 😄

benmonro:
Thanks @SimenB !

PR #293, opened 2019-07-06
feat(rules): adds `no-if` rule

cartogram:
If statements in tests is usually a bad idea. This PR prevents them.

 ```js
it('foo', () => {
  if ('bar') {
    // an if statement here is invalid
    // you are probably testing too much
  }
});

 it('foo', () => {
  const bar = foo ? 'bar' : null;
});
```

 ```js
it('foo', () => {
  // only test the 'foo' case
});

 it('bar', () => {
  // test the 'bar' case separately
});

 it('foo', () => {
  function foo(bar) {
    // nested functions are valid
    return foo ? bar : null;
  }
});
```

### DIFF
--- a/src/rules/no-if.js
+++ b/src/rules/no-if.js
@@ -1,0 +1,88 @@
+import { getDocsUrl, testCaseNames, getNodeName, isTestCase } from './util';
+
+const isTestArrowFunction = node =>
+  node !== undefined &&
+  node.type === 'ArrowFunctionExpression' &&
+  node.parent.type === 'CallExpression' &&
+  testCaseNames.has(getNodeName(node.parent.callee));
+
+export default {
+  meta: {
+    docs: {
+      description: 'Disallow conditional logic',
+      category: 'Best Practices',
+      recommended: false,
+      uri: getDocsUrl('jest/no-if'),
+    },
+    messages: {
+      noIf: [
+        'Tests should not contain if statements.',
+        'This is usually an indication that you',
+        'are attempting to test too much at once',
+        'or not testing what you intend to.',
+        'Consider breaking the if statement out',
+        'into a separate test to resolve this error.',
+      ].join(' '),
+      noConditional: [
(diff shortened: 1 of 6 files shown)


src/rules/no-if.js
```diff
@@ -0,0 +1,93 @@
...
+        'using a non-null assertion operator (!) instead.',
+      ].join(' '),
+    },
+  },
+
+  create(context) {
+    const stack = [];
+
+    function validate(node) {
+      if (!stack[stack.length - 1]) {
```
SimenB: Could you refactor this a bit? I had to read it once or twice and look at what's pushed to `stack` to understand this. Maaaaybe something like


```js
const lastElementInStack = stack[stack.length - 1];

if (lastElementInStack === false) {
  return;
}
```

The reason I added the explicit `=== false` here was because I initially read it as you just check if the element was there, which is not what happens (I _think_) 🙂

cartogram: @SimenB I updated the PR with this change. The `!` (versus explicit `=== false`) also caught cases when the stack was empty so I amended with a more explicit check for that case too.

src/rules/no-if.js
```diff
@@ -0,0 +1,88 @@
...
+  meta: {
+    docs: {
+      description: 'Disallow conditional logic',
+      category: 'Best Practices',
+      recommended: false,
+      uri: getDocsUrl('jest/no-if'),
+    },
+    messages: {
+      noIf: [
+        'Tests should not contain if statements.',
```
SimenB: I wonder if the first sentence (for both messages) is enough. If people want more background, they can open up the docs

src/rules/no-if.js
SimenB: thanks, easier to follow now!

src/rules/no-if.js
cartogram: @SimenB makes sense to me 👍

PR #298, opened 2019-07-11
New Rule - no-duplicate-hooks

### LINKED ISSUE #231: New rule: no duplicated hooks
A rule that disallows duplicated hooks in the same describe.
For example this code will fail, because you should combine the two `beforeEach`s into one:

```
describe('my describe', () => {
    beforeEach(() => {...});
    beforeEach(() => {...});

    it('my it', () => {...})
});
```

M4rk9696:
Fix #231

### DIFF
--- a/src/rules/__tests__/no-duplicate-hooks.test.js
+++ b/src/rules/__tests__/no-duplicate-hooks.test.js
@@ -27,6 +27,21 @@ ruleTester.run('basic describe block', rule, {
       '  some_fn();',
       '})',
     ].join('\n'),
+    [
+      'describe("foo", () => {',
+      '  beforeAll(() => {',
+      '  }),',
+      '  beforeEach(() => {',
+      '  }),',
+      '  afterEach(() => {',
+      '  }),',
+      '  afterAll(() => {',
+      '  }),',
+      '  test("bar", () => {',
+      '    some_fn();',
+      '  })',
+      '})',
+    ].join('\n'),
   ],
 
   invalid: [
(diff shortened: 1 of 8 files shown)


src/rules/util.js
```diff
@@ -99,6 +99,13 @@ const testCaseNames = Object.assign(Object.create(null), {
   xtest: true,
 });
 
+const testHookNames = Object.assign(Object.create(null), {
```
jeysal: Can we use a Set for this? (and change the code above as well)
This way of doing it is probably from before Sets were supported everywhere.

src/rules/__tests__/no-duplicate-hooks.test.js
```diff
@@ -0,0 +1,311 @@
...
+const rule = require('../no-duplicate-hooks');
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 6,
+  },
+});
+
+ruleTester.run('basic describe block', rule, {
+  valid: [
```
jeysal: I think an additional test checking that different kinds of hooks in one describe block are allowed would be good

src/rules/util.js
M4rk9696: I thought it was the standard followed in the project, will change it

jeysal: It probably was so far, but I imagine only for historical reasons so it's a good time to change that when touching this code :)

src/rules/no-focused-tests.js
```diff
@@ -2,16 +2,14 @@
...
 
-const testFunctions = Object.assign(Object.create(null), {
-  describe: true,
-  it: true,
-  test: true,
-});
+const testFunctions = new Set(['describe', 'it', 'test']);
 
-const matchesTestFunction = object => object && testFunctions[object.name];
+const matchesTestFunction = object => object && testFunctions.has(object.name);
```
SimenB: way better, thanks!

PR #303, opened 2019-07-16
fix(require-tothrow-message): require throw messages on async functions

chrisblossom:
This PR fixes `require-tothrow-message` to properly require messages for async functions that use `.rejects.toThrow(Error)`.

### DIFF
--- a/src/rules/__tests__/require-tothrow-message.test.js
+++ b/src/rules/__tests__/require-tothrow-message.test.js
@@ -12,70 +12,50 @@ const ruleTester = new RuleTester({
 ruleTester.run('require-tothrow-message', rule, {
   valid: [
     // String
+    "expect(() => { throw new Error('a'); }).toThrow('a');",
+    "expect(() => { throw new Error('a'); }).toThrowError('a');",
     `test('string', async () => {
-        const error = new Error('a');
-        const throwErrorSync = () => { throw new Error(error) };
-        const throwErrorAsync = async () => { throw new Error(error) };
-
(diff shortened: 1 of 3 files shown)


src/rules/__tests__/require-tothrow-message.test.js
```diff
@@ -5,28 +5,80 @@ const rule = require('../require-tothrow-message');
...
   parserOptions: {
-    ecmaVersion: 6,
+    ecmaVersion: 8,
   },
 });
 
 ruleTester.run('require-tothrow-message', rule, {
   valid: [
     // String
-    "expect(() => { throw new Error('a'); }).toThrow('a');",
```
SimenB: can you keep the old tests and just add new ones covering the fixed behavior? _If_ they start failing, it's easier to fix if they are as minimal as possible

chrisblossom: Is there a better way to use `await` without having to wrap the whole test inside an `async function`?

SimenB: Nope (afaik)

chrisblossom: Do you prefer force push to keep commits clean or commit normally and you'll merge them later?

SimenB: Commit normally, please :) easier to follow, and we use squash merges

chrisblossom: I think I updated as you requested. Let me know if there is anything else!

PR #306, opened 2019-07-17
fix(require-tothrow-message): rename rule to require-to-throw-message

### LINKED ISSUE #296: rename require-tothrow-message to require-to-throw-message
Would a PR be accepted that would rename `require-tothrow-message` to `require-to-throw-message`? `tothrow` is not a word and for those of us using a spell checker in our IDE (webstorm / vscode plugin) it is distracting.

I think the PR would not remove `require-tothrow-message`, but rename all instances of it and possibly deprecate the old name.

chrisblossom:
This change makes consistent with other rule naming format and removes spelling error message from editors.

This is not ready to be merged. Just more of a proof of concept / there is probably a better way within eslint to rename rules. Also `src/__tests__/rules.test.js` currently fails tests.

Closes #296.


jeysal:
Seems like a reasonably gentle way to do it if we decide to do the rename. For the test, just increasing the count by 1 until it's removed in a major is probably fine, the docs test might need a whitelist.

src/rules/require-tothrow-message.js
```diff
@@ -1,41 +1,13 @@
...
-        if (!expectCase(node)) {
-          return;
-        }
+const replacedBy = ['require-to-throw-message'];
 
-        let targetNode = method(node);
-        if (targetNode.name === 'rejects') {
-          targetNode = method(node.parent);
-        }
+const requireTothrowMessage = Object.assign({}, requireToThrowMessage, {
```
SimenB: you can use object spread (we transpile with babel)

chrisblossom: I didn't see this repo was transpiled...I was waiting for the TS conversion to do this. I'll make that change.

src/rules/__tests__/require-tothrow-message.test.js
```diff
@@ -1,13 +1,13 @@
 import { RuleTester } from 'eslint';
```
SimenB: can you rename the test file as well?

src/__tests__/rules.test.js
```diff
@@ -2,7 +2,14 @@ import { existsSync } from 'fs';
 import { resolve } from 'path';
 import { rules } from '../';
 
-const ruleNames = Object.keys(rules);
+const ruleNames = Object.keys(rules).filter(rule => {
+  const excludeRules = [
```
SimenB: can you create this array outside of the `filter`?

Also maybe a separate test that all "excluded" rules have the `meta.deprecated` property set?

chrisblossom: > Also maybe a separate test that all "excluded" rules have the meta.deprecated property set?

I thought about that, but I think that is only wanted for renamed rules. If a rule is deprecated without renaming, you'll probably want to keep it inside this test. Right?

SimenB: We use `import` & `export` 🙂 And yeah, I landed the Babel change to make the TS migration have a smaller diff

src/rules/require-tothrow-message.js
SimenB: We use `import` & `export` 🙂 And yeah, I landed the Babel change to make the TS migration have a smaller diff

src/__tests__/rules.test.js
chrisblossom: > We use import & export

Ha yeah it should have been obviously. Just wasn't thinking. Silly me.

SimenB: good point!

src/rules/require-tothrow-message.js
chrisblossom: Updated to object spread

src/rules/__tests__/require-tothrow-message.test.js
chrisblossom: I renamed the file. Not sure why this isn't showing "outdated"?

SimenB: the comment itself is not (since the line I commented on is the same), even though the file has been renamed.

(at least that's my guess 😛)

src/rules/require-tothrow-message.js
```diff
@@ -1,41 +1,18 @@
...
+  ...requireToThrowMessage,
   meta: {
+    ...requireToThrowMessage.meta,
     docs: {
-      url: getDocsUrl(__filename),
-    },
-    messages: {
-      requireRethrow: 'Add an error message to {{ propertyName }}()',
+      ...requireToThrowMessage.meta.docs,
+      replacedBy,
```
SimenB: I don't thing this should be under `docs`, should it? just `meta`

chrisblossom: I'm not sure. I copied what others did and I found examples of both:

[eslint/rules/no-native-reassign.js#L19](https://github.com/eslint/eslint/blob/v3.8.1/lib/rules/no-native-reassign.js#L14-L37)
[typescript-eslint/rules/prefer-interface.ts#L20](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/rules/prefer-interface.ts#L6-L21)

SimenB: In the documentation it's not under `docs`: https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

chrisblossom: Removed `meta.docs.replacedBy`.

SimenB:
Just landed the TS rewrite of this rule, mind rebasing? Absolutely no rush as we're still some way off a new major, but I thought I'd flag it early 🙂

chrisblossom:
I wasn't sure how to properly rebase so I manually added the changes and force pushed.

[Here are the commits before the rebase.](https://github.com/chrisblossom/eslint-plugin-jest/commits/rename-require-tothrow-message-before-rebase)

@SimenB This PR does not have any breaking changes because it deprecates the old `rename-require-tothrow-message` rule instead of removing it, so if accepted, this should not need to wait for a major version. When there is a major released, we can remove the deprecated rule completely.

src/rules/require-to-throw-message.ts
```diff
@@ -0,0 +1,50 @@
...
+          matcher &&
+          matcher.arguments &&
+          matcher.arguments.length === 0 &&
+          ['toThrow', 'toThrowError'].includes(matcher.name) &&
+          (!modifier ||
+            !(modifier.name === ModifierName.not || modifier.negation))
+        ) {
+          // Look for `toThrow` calls with no arguments.
+          context.report({
+            messageId: 'requireRethrow', // todo: rename to 'addErrorMessage'
```
G-Rath: I wouldn't object if you wanted to do these `todo`s for me 😉

src/__tests__/rules.test.ts
```diff
@@ -2,7 +2,15 @@ import { existsSync } from 'fs';
 import { resolve } from 'path';
 import plugin from '../';
 
-const ruleNames = Object.keys(plugin.rules);
+const excludeRules = [
+  // require-tothrow-message has been renamed to require-to-throw-message, remove in major version bump
+  'require-tothrow-message',
+];
+
+const ruleNames = Object.keys(plugin.rules).filter(rule => {
```
G-Rath: This could be short-handed

README.md
```diff
@@ -193,7 +193,7 @@ https://github.com/dangreenisrael/eslint-plugin-jest-formatting
 [prefer-to-contain]: docs/rules/prefer-to-contain.md
 [prefer-to-have-length]: docs/rules/prefer-to-have-length.md
 [prefer-todo]: docs/rules/prefer-todo.md
-[require-tothrow-message]: docs/rules/require-tothrow-message.md
```
G-Rath: It might be nicer to have both docs, & change `require-tothrow-message` to had "deprecated" + a link to the replacement as it's body.

That way any existing links would keep working, and would save having to change `rules.test` twice.

src/rules/require-to-throw-message.ts
chrisblossom: Would that be a breaking change? I am not too familiar how `messageId` is used.

src/__tests__/rules.test.ts
chrisblossom: I generally don't use implicit returns on arrow functions because I think they are harder to work with, but it doesn't matter to me either way. Do you want me to change it?

README.md
chrisblossom: I personally think it is more confusing to have them both listed and is better solved in the back with code. But if this is wanted I'll make the change no problem.

src/rules/require-to-throw-message.ts
G-Rath: No, it's not breaking - `messageId` is just used internally by `eslint` to make it easier to test, [so you don't have to write things out multiple times](https://eslint.org/docs/developer-guide/working-with-rules#messageids).

chrisblossom: I've updated this PR to include the changes in TODO. Please verify that the `context.report.data.propertyName` -> `context.report.data.matcherName` was wanted.

chrisblossom:
I've resolved the conflicts. Let me know how I should proceed with the remaining PR comments.

src/__tests__/rules.test.ts
G-Rath: I think I would, if you don't mind; we use shorthand arrow functions in most other places, and imo they can make filters easier to read by providing a clearer distinction between simple and advance filters.

For example, in this case the filter can just be `filter(rule => !excludeRules.includes(rules));`, which reads as "filter rules not included in excludeRules".

Any conditional more complex than this and I'd agree w/ you that it should be long form.

They tend to be nicer w/ TypeScript anyway, since you don't have to worry about returns being used by mistake if you short-hand.

G-Rath:
@chrisblossom Sorry for the delay in getting back to you. I'm happy w/ this aside from the requested change.

Cheers for actioning the `TODO` stuff for me - you did it as it was wanted 😄

chrisblossom:
Thanks @G-Rath for the response! I think I've made the final requested change. If there is anything else please let me know.

G-Rath:
@chrisblossom sorry I forgot about this 😬 

I'm prepping a `next` branch for release right now, so if you could find some time in the next couple of days to resolve the conflict, that'd be great, as I can then pull it from master into `next` :)

SimenB:
resolved the conflict, made breaking, and changed to target `next`. if CI is happy I'll merge into `next`

PR #307, opened 2019-07-17
feat(no-export): new rule for no-export

benmonro:
I ran into this issue when we had a set of helper functions in 1 test file that
got exported directly from a `-test.js` file. The test that was importing this
file then executes both it's own tests as well as the tests in the imported
file. We started seeing phantom snapshots show up that we weren't able to find
the origin of until we finally tracked it down and realized that a test file was
exporting shared functions. We fixed this by moving those functions out into
seperate files, but that was the inspiration of this rule.

This rule aims to eliminate duplicate runs of tests by exporting things from
test files. If you import from a test file, then all the tests in that file will
be run in each imported instance, so bottom line, don't export from a test, but
instead move helper functions into a seperate file when they need to be shared
across tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ installations requiring long-term consistency.
 | [no-commented-out-tests][]   | Disallow commented out tests                                      |                  |                     |
 | [no-duplicate-hooks][]       | Disallow duplicate hooks within a `describe` block                |                  |                     |
 | [no-empty-title][]           | Disallow empty titles                                             |                  |                     |
+| [no-export][]                | Disallow export from test files                                   | ![recommended][] |                     |
 | [no-focused-tests][]         | Disallow focused tests                                            | ![recommended][] |                     |
 | [no-hooks][]                 | Disallow setup and teardown hooks                                 |                  |                     |
 | [no-identical-title][]       | Disallow identical titles                                         | ![recommended][] |                     |
@@ -163,6 +164,7 @@ https://github.com/dangreenisrael/eslint-plugin-jest-formatting
 [no-duplicate-hooks]: docs/rules/no-duplicate-hooks.md
 [no-commented-out-tests]: docs/rules/no-commented-out-tests.md
 [no-empty-title]: docs/rules/no-empty-title.md
+[no-export]: docs/rules/no-export.md
 [no-focused-tests]: docs/rules/no-focused-tests.md
 [no-hooks]: docs/rules/no-hooks.md
 [no-identical-title]: docs/rules/no-identical-title.md
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ installations requiring long-term consistency.
 | [no-commented-out-tests][]   | Disallow commented out tests                                      |                  |                     |
 | [no-duplicate-hooks][]       | Disallow duplicate hooks within a `describe` block                |                  |                     |
 | [no-empty-title][]           | Disallow empty titles                                             |                  |                     |
-| [no-export][]                | Disallow export from test files                                   | ![recommended][] |                     |
+| [no-export][]                | Disallow export from test files                                   |                  |                     |
 | [no-focused-tests][]         | Disallow focused tests                                            | ![recommended][] |                     |
 | [no-hooks][]                 | Disallow setup and teardown hooks                                 |                  |                     |
 | [no-identical-title][]       | Disallow identical titles                                         | ![recommended][] |                     |
--- a/docs/rules/no-export.md
+++ b/docs/rules/no-export.md
@@ -1,12 +1,7 @@
 # no export from test file (no-export)
 
-I ran into this issue when we had a set of helper functions in 1 test file that
-got exported directly from a `-test.js` file. The test that was importing this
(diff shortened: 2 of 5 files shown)


docs/rules/no-export.md
```diff
@@ -0,0 +1,51 @@
+# no export from test file (no-export)
+
+I ran into this issue when we had a set of helper functions in 1 test file that
```
SimenB: While it's really useful to understand why you want this rule, I don't think we need it in the docs

benmonro:
@SimenB actually not hard at all to check for a test and a great idea! I'll add that

docs/rules/no-export.md
benmonro: fair enough.  fixed.

README.md
```diff
@@ -114,6 +114,7 @@ installations requiring long-term consistency.
 | [no-commented-out-tests][]   | Disallow commented out tests                                      |                  |                     |
 | [no-duplicate-hooks][]       | Disallow duplicate hooks within a `describe` block                |                  |                     |
 | [no-empty-title][]           | Disallow empty titles                                             |                  |                     |
+| [no-export][]                | Disallow export from test files                                   | ![recommended][] |                     |
```
SimenB: Not recommended (changing recommended rules is a breaking change)

src/rules/no-export.js
```diff
@@ -0,0 +1,49 @@
...
+        exportNodes.push(node);
+      },
+      'AssignmentExpression > MemberExpression'(node) {
+        let { object, property } = node;
+
+        if (object.type === 'MemberExpression') {
+          ({ object, property } = object);
+        }
+
+        if (object.name === 'module' && !!property.name.match(/^exports?$/)) {
```
SimenB: `/^exports?$/.test(...)` instead of `match` and you don't need the double bang

src/rules/no-export.js
```diff
@@ -0,0 +1,49 @@
+import { getDocsUrl, isTestCase } from './util';
+
+let exportNodes = [];
```
SimenB: These should be inside the `create` to not leak between tests

src/rules/no-export.js
```diff
@@ -0,0 +1,49 @@
...
+let hasTestCase = false;
+const messageId = 'unexpectedExport';
+export default {
+  meta: {
+    docs: {
+      url: getDocsUrl(__filename),
+    },
+    messages: {
+      [messageId]: `Do not export from a test file.`,
```
SimenB: No need to use a variable for `messageId`. I find it easier to read as a literal and we'll type check it when migrating to TS

benmonro:
should be good to merge whenever you can @SimenB

README.md
benmonro: what's the process of getting it to be a recommended change?  This rule prevents very hard to track down problems....

SimenB: Just remembering when we release a new major :)

benmonro: i'll remove it for now, but I would like to move this to recommended in another PR, if you don't mind providing guidance on that process.

chrisblossom:
Sometimes you need to add an empty `export {}` in a typescript file to workaround getting typescript error `error TS1208: Cannot compile namespaces when the '--isolatedModules' flag is provided` https://github.com/Microsoft/TypeScript/issues/15230

Because of that, I think it would be helpful to let `.test.ts` or `.test.tsx` files have an empty export function.

src/rules/no-export.js
benmonro: done, thanks!

src/rules/no-export.js
benmonro: done, thx

src/rules/no-export.js
benmonro: done

SimenB:
>Because of that, I think it would be helpful to let `.test.ts` or `.test.tsx` files have an empty export function.

I'm down with an option (`allowEmpty`, or something), but I don't think it should be the default behaviour. People can use `overrides` in their config if they want 🙂

README.md
SimenB: Sure! Just open up a PR updating the readme and add it to the config exported from `index` and I'll make sure to merge it before making a major release

benmonro:
can you merge @SimenB ? or give me permission to merge? :)

benmonro:
I can do the option in another PR

benmonro:
@SimenB any guidance you can provide on what it takes to get a rule to be 'recommended'?  The bugs that come about are quite nefarious and time consuming to track down, especially for junior devs...

benmonro:
@SimenB looks like the build failed on master. :( https://travis-ci.org/jest-community/eslint-plugin-jest/jobs/560608618

SimenB:
> @SimenB any guidance you can provide on what it takes to get a rule to be 'recommended'?

Not much guidance needed, just add the badge to the readme and add the rule here: https://github.com/jest-community/eslint-plugin-jest/blob/b3a360d3d033e22a940a49c7834d929ddb14a1b7/src/index.js#L43-L55

I'll make sure to merge the PR for the major 🙂

PR #310, opened 2019-07-18
ts-migration/convert-is-utils-fns

G-Rath:
Converts a couple of `isX` utility methods into their TS variants, so that it's easier to convert rules :)

### DIFF
--- a/src/rules/lowercase-name.ts
+++ b/src/rules/lowercase-name.ts
@@ -2,17 +2,15 @@ import {
   AST_NODE_TYPES,
   TSESTree,
 } from '@typescript-eslint/experimental-utils';
-import { createRule } from './tsUtils';
-
-type JestFunctionName = 'it' | 'test' | 'describe';
-
-interface JestFunctionIdentifier extends TSESTree.Identifier {
-  name: JestFunctionName;
-}
-
-interface TestFunctionCallExpression extends TSESTree.CallExpression {
-  callee: JestFunctionIdentifier;
-}
+import {
+  DescribeAlias,
+  JestFunctionCallExpression,
+  JestFunctionName,
+  TestCaseName,
+  createRule,
+  isDescribe,
+  isTestCase,
+} from './tsUtils';
 
 type ArgumentLiteral = TSESTree.Literal | TSESTree.TemplateLiteral;
 
@@ -20,22 +18,10 @@ interface FirstArgumentStringCallExpression extends TSESTree.CallExpression {
   arguments: [ArgumentLiteral];
 }
 
-type CallExpressionWithCorrectCalleeAndArguments = TestFunctionCallExpression &
+type CallExpressionWithCorrectCalleeAndArguments = JestFunctionCallExpression<
+  TestCaseName.it | TestCaseName.test | DescribeAlias.describe
+> &
   FirstArgumentStringCallExpression;
-
-const isItTestOrDescribeFunction = (
-  node: TSESTree.CallExpression,
-): node is TestFunctionCallExpression => {
-  const { callee } = node;
-
-  if (callee.type !== AST_NODE_TYPES.Identifier) {
-    return false;
-  }
-
-  const { name } = callee;
-
-  return name === 'it' || name === 'test' || name === 'describe';
-};
 
 const hasStringAsFirstArgument = (
   node: TSESTree.CallExpression,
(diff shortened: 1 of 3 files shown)


src/rules/tsUtils.ts
```diff
@@ -28,6 +43,26 @@ export const isFunction = (node: TSESTree.Node): node is FunctionExpression =>
   node.type === AST_NODE_TYPES.FunctionExpression ||
   node.type === AST_NODE_TYPES.ArrowFunctionExpression;
 
+export const isHook = (node: TSESTree.CallExpression): boolean => {
```
SimenB: I wonder if we can make this (and the others) into a type guard. Sort of like what I've done in https://github.com/jest-community/eslint-plugin-jest/blob/6958f04ce09bee8988b84453b3e84e9358d2482c/src/rules/lowercase-name.ts#L7-L51

It narrows down the types as we go

(might wanna move the ones I linked to in here as well?)

SimenB:
Hmm, CI also fails since these helpers are not used... Can add an istanbul ignore comment above until we use them

src/rules/tsUtils.ts
G-Rath: I'd advise against that - it's actually something I've recently been talking about with a few different people.

The problem is that type guards imply if something is *or is not* a given type, which will cause massive problems if you start putting business logic in your guards.

For example:

```
const isPositiveNumber = (n: unknown): n is number => typeof n === 'number' && n > 0;

const fn = (p: number | string): string => {
  if(isPositiveNumber(p)) {
    return p.toString();
  } //        ^ type: number

  return p.toUpperCase();
  //        ^ type: string | !number
}
```

Another way (and maybe more accurate) to say is it that type guards don't just narrow, they also *negate*.

G-Rath: However in looking over your code, I think you might already know that, and has accounted for it :)

If you're happy with it, I'm happy with it :joy:

SimenB: Yeah, they should definitely be logic free, just checking the _type_ of things, not if they pass some constraint. But e.g. verifying the name of a function and the type of its arguments (or at least, say, the first n arguments) is really useful as we then don't have to use type casts to please the compiler. As long as we're careful to just do narrowing that types can verify, I think we're fine

G-Rath: Yeah totally - your code I think is perfect; it wouldn't work w/ the `number` example (that needs [this](https://github.com/Microsoft/TypeScript/issues/15480)), but it should be totally spot on for this.

The only thing I think is that the enums will need to be strings:

```
enum DescribeAlias {
  'describe' = 'describe',
  'fdescribe' = 'fdescribe',
  'xdescribe' = 'xdescribe',
}
```

Otherwise, we can't use them as both type values and as enums, since `keyof DescribeAlias` currently returns `0,1,2,etc` :(

SimenB: That's fine 🙂 Not sure what the cleanest is there, but easy enough to change if a better alternative comes up at some point

src/rules/lowercase-name.ts
```diff
@@ -48,7 +34,9 @@ const hasStringAsFirstArgument = (
 const isJestFunctionWithLiteralArg = (
   node: TSESTree.CallExpression,
 ): node is CallExpressionWithCorrectCalleeAndArguments =>
-  isItTestOrDescribeFunction(node) && hasStringAsFirstArgument(node);
+  (isTestCase(node) || isDescribe(node)) &&
+  ['it', 'test', 'describe'].includes(node.callee.name) &&
```
SimenB: doesn't `(isTestCase(node) || isDescribe(node))` cover this line?

src/rules/tsUtils.ts
G-Rath: I've pushed up an implementation, which I am actually really happy w/.

Despite how it might first appear, I think it shuffles around the current learning curve rather than adding to it, while making it super quick to implement scoping on rules to target specific jest functions :D

Let me know what you think - I've applied it to `lowercase-name` as a showcase, but can cherry-pick that into another PR if you'd like.

src/rules/lowercase-name.ts
G-Rath: I was going to ask you about that :)

Previously the check was just for these three functions, and the description & tests for the rule only focus on those three:

> Enforce `it`, `test` and `describe` to have descriptions that begin with a lowercase letter. This provides more readable test failures.

I think it should be fine to remove that line, but wanted to implement the original behaviour + ask first.

G-Rath:
btw I've had to remove the `commitlint` hook from `package.json`, as it just refuses to work properly.

Could you confirm for me if it's working on your end, and what OS you're using?

I'm on Win10 w/ WSL - I've tried removing & reinstalling from both cmd & wsl, plus updating husky, replacing just the `commitmsg` hook w/ the latest version from husky while still using`v1`, etc all to no success :(

G-Rath:
I've started converting the `expectCase` methods too.

Let me know if you'd like them in a separate PR; otherwise I'll push them up to this one soon :)

I'm applying the typeguard patten where possible, renaming such methods to "is<x>" and leaving an exported const alias in place of the old version w/ a `@deprecated` link.

src/rules/lowercase-name.ts
SimenB: ah right, since we'd now get `xtest`, `ftest` etc? I think it's fine to detect all - they're just modifiers on the same functions

SimenB: We can do that in a follow-up, though

PR #311, opened 2019-07-19
chore(package): update `@commitlint` to `^7.0.0`

G-Rath:
This resolves problems w/ committing on Windows, where the hooks would
fail due to `$HUSKY_GIT_PARAMS` issues.

@SimenB Apparently this is [a common thing]? Either way, w/ this the commit hooks now all work fine for me :)

Let me know if you want me to make another PR into the TS branch, or if you'll handle that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,8 @@
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.4",
     "@babel/preset-env": "^7.4.4",
-    "@commitlint/cli": "^6.0.0",
-    "@commitlint/config-conventional": "^6.0.0",
+    "@commitlint/cli": "^7.0.0",
+    "@commitlint/config-conventional": "^7.0.0",
     "babel-eslint": "^10.0.2",
     "babel-jest": "^24.8.0",
     "eslint": "^5.1.0",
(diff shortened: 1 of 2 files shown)


SimenB:
The reason I haven't upgraded is that this requires node 8 and we support node 6. Could of course run `yarn` with `--ignore-engines` on CI. Would it work for you to only have this upgrade locally, but not push it? It only affect package.json and the lockfil, so you shouldn't need to juggle changes in a file for what to push and what not to

G-Rath:
Ah yeah of course - didn't notice this with all the rushing around.

> Would it work for you to only have this upgrade locally, but not push it?

The only problem is it's preventing me from merging my branch, since that reverts back to the original `package.json`, which has the deprecated usage of `-e`.

So far I'm doing everything on a [single generic branch](https://github.com/G-Rath/eslint-plugin-jest/tree/ts-migration/general) anyway, so it's not a big deal (then I can just cherry-pick commits out for PR later).

Sorry about that!

SimenB:
Happy to release a major dropping node 6 (it's eol after all). Just need to figure out which other changes we want to make at the same time

G-Rath:
Node6 can't be dropped fast enough... 😭

SimenB:
I thought maybe to merge the ta rewrite, update the recommended rules, and make a major release

G-Rath:
Being done in [next](#443)

PR #314, opened 2019-07-19
ts-migration/no-focused-tests

G-Rath:
The meta information should be checked to make sure it's correct :)


src/rules/no-focused-tests.ts
```diff
@@ -0,0 +1,85 @@
+import {
+  AST_NODE_TYPES,
+  TSESTree,
+} from '@typescript-eslint/experimental-utils';
+import { DescribeAlias, TestCaseName, createRule } from './tsUtils';
+
+const testFunctions = new Set(['describe', 'it', 'test']);
+
+const matchesTestFunction = (
```
SimenB: It'd be great to use the type guards from your other PR here... Let's land that first then use them here

SimenB: merged the utils PR 🙂

G-Rath: Those guards actually won't help here, but looking over this has highlighted that the guards are not entirely correct.

Right now the guards only narrow to a `CallExpression` w/ a `callee` of type `Identifier` and who's name is a "jest method", *but* they also return true for a `CallExpression` w/ a `callee` of type `MemberExpression`, whos `object` is of type `Identifier` and who's name is a "jest method".

That's fine and good, and I have reimplemented both the guards & `lowercase-name` to account for this; but that's also why they can't be used here: this rule is testing *three levels down*, in order to support `describe.only.each()`.

Visually:

```
describe(); // callee
describe.only(); // callee.object
describe.only.each() // callee.object.object
```

However, this means we can't use the guards *at all* here, since if they return true, they're saying that the `node` is *at most* two levels deep - and so any check to the contrary isn't valid b/c there's no overlap:

> Error:(62, 11) TS2367: This condition will always return 'false' since the types 'AST_NODE_TYPES.Identifier' and 'AST_NODE_TYPES.MemberExpression' have no overlap. 

I think for now the best course of action might be to just land this PR, and once more rules are converted look at re-scoping the guards, as that way it'll be easier to asset the implications.

G-Rath: #316 has the type guard improvements.

As I said in its description, I've been left a bit paranoid now b/c I don't know the plugin well enough to be able to scope out the impact of changing the behaviour of the guards so much (since we can't rely on the tests until we convert a lot more rules).

Let me know how'd you like to proceed. I'm happy to sit on this rule until we've got more converted if you want :)

SimenB: Yeah, we can convert over more rules then go back and see if we can create better type guards when it's easier to assess the impact

src/rules/no-focused-tests.ts
```diff
@@ -0,0 +1,85 @@
+import {
+  AST_NODE_TYPES,
+  TSESTree,
+} from '@typescript-eslint/experimental-utils';
+import { DescribeAlias, TestCaseName, createRule } from './tsUtils';
```
SimenB: You forgot to export these new ones

image

G-Rath: O.o I feel like I need to [commit sudoku](https://knowyourmeme.com/memes/commit-sudoku) - how did I miss that...?

G-Rath: Actually no it just needs to be merged w/ `reapply-ts` :)

SimenB: ah, you've cherry-picked. Makes sense 🙂

G-Rath: Yeah it got lost in the shuffle. The whole guard thing + timezones threw me for a loop :joy:

PR #315, opened 2019-07-19
ts-migration/convert-some-utils-functions

G-Rath:
This might cause issues with #310 but these three functions are the only ones from `tsUtils` that are required for about three or four rules, so getting these across the line will let me land those straight away :)

### DIFF
--- a/src/rules/__tests__/no-jasmine-globals.test.ts
+++ b/src/rules/__tests__/no-jasmine-globals.test.ts
@@ -54,6 +54,10 @@ ruleTester.run('no-jasmine-globals', rule, {
       output: 'jest.setTimeout(5000);',
     },
     {
+      code: 'jasmine.DEFAULT_TIMEOUT_INTERVAL = function() {}',
+      errors: [{ messageId: 'illegalJasmine', column: 1, line: 1 }],
+    },
+    {
       code: 'jasmine.addMatchers(matchers)',
       errors: [
         {
--- a/src/rules/__tests__/no-jasmine-globals.test.ts
+++ b/src/rules/__tests__/no-jasmine-globals.test.ts
@@ -1,7 +1,7 @@
-import { RuleTester } from 'eslint';
+import { TSESLint } from '@typescript-eslint/experimental-utils';
 import rule from '../no-jasmine-globals';
 
-const ruleTester = new RuleTester();
+const ruleTester = new TSESLint.RuleTester();
 
 ruleTester.run('no-jasmine-globals', rule, {
   valid: [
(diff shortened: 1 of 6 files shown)


SimenB:
Coverage isn't happy... Would it be quick to port a rule which uses these?

SimenB:
I've merged #310 as well, fwiw. Doesn't seem like it created a merge conflict 🙂

G-Rath:
@SimenB done - I've pushed `no-jasmine-globals` :)

src/rules/no-jasmine-globals.ts
```diff
@@ -1,11 +1,14 @@
...
 
-export default {
+export default createRule({
+  name: __filename,
   meta: {
     docs: {
-      url: getDocsUrl(__filename),
+      category: 'Best Practices',
+      description: 'Disallow Jasmine globals',
+      recommended: false,
```
SimenB: This is recommended :)

src/rules/no-jasmine-globals.ts
```diff
@@ -109,18 +117,26 @@ export default {
...
-                fix(fixer) {
-                  return [
-                    fixer.replaceText(
-                      node.parent,
-                      `jest.setTimeout(${node.parent.right.value})`,
-                    ),
-                  ];
-                },
+                fix:
+                  right.type !== AST_NODE_TYPES.Literal
```
SimenB: Can you flip it so the `true` case is first in the ternary? Or do a `return` above this if it's of the wrong type

G-Rath:
@SimenB this is good to go!

I pulled in another rule (`no-disabled-tests`) to help improve the coverage, along w/ adding a couple more tests & re-arranging some type-narrowing if conditions so that they get hit before non-narrowing ones :)

Everything's now back at 100% coverage, and `utils.js` has gotten smaller since those two rules were the only ones using some of the methods :D

src/rules/__tests__/no-jasmine-globals.test.ts
```diff
@@ -53,6 +53,10 @@ ruleTester.run('no-jasmine-globals', rule, {
       errors: [{ messageId: 'illegalJasmine', column: 1, line: 1 }],
       output: 'jest.setTimeout(5000);',
     },
+    {
+      code: 'jasmine.DEFAULT_TIMEOUT_INTERVAL = function() {}',
```
SimenB: is this valid jasmine code?

SimenB: I'm happy to warn on it either way

G-Rath: I don't think so, but it's technically something that could be done. Eitherway, it triggers the previously uncovered branch :)

PR #316, opened 2019-07-20
ts-migration/improve-guards

G-Rath:
I didn't type the guards correctly the first time - see [this comment](https://github.com/jest-community/eslint-plugin-jest/pull/314#discussion_r305560370) for some more info.

This is something that should be looked into at some point; in JS the type guards are implemented using `getNodeName`, which'll give you the following:

```
describe(); // describe
describe.only(); // describe.only
describe.only.each('') // describe.only.each
```

This is then checked using `#has` against a `Set`, which has since become an enum.
However the original `Set` contained both functions and function chains:

```
const describeAliases = new Set([
  'describe',
  'describe.only',
  'describe.skip',
  'fdescribe',
  'xdescribe',
]);
```

I've retyped the guards so that they now match the actual runtime checks that are happening, but this means that there is a behavioural difference from the JS version that I think should be commented on.

The behavioural change is that now things like `describe.<anything>()` will be cause `isDescribe` to return true.

It's all left me a bit paranoid right now, b/c I've not got enough info to know how big of an impact this'll have on everything - so far I've not had any problems (i.e no tests have failed), but it's technically possible that it means the plugin could try and lint things it shouldn't (which wouldn't necessarily cause an error, since it'll be doing other checks) which we might care about?

This PR should definitely be merged, as it only changes the types to reflect the runtime checks.

### DIFF
--- a/src/rules/tsUtils.ts
+++ b/src/rules/tsUtils.ts
@@ -38,16 +38,42 @@ export enum HookName {
 
 export type JestFunctionName = DescribeAlias | TestCaseName | HookName;
 
-interface JestFunctionIdentifier<FunctionName extends JestFunctionName>
+export interface JestFunctionIdentifier<FunctionName extends JestFunctionName>
   extends TSESTree.Identifier {
   name: FunctionName;
 }
 
-export interface JestFunctionCallExpression<
-  FunctionName extends JestFunctionName = JestFunctionName
+export interface JestFunctionMemberExpression<
+  FunctionName extends JestFunctionName
+> extends TSESTree.MemberExpression {
+  object: JestFunctionIdentifier<FunctionName>;
+}
+
+export interface JestFunctionCallExpressionWithMemberExpressionCallee<
+  FunctionName extends JestFunctionName
+> extends TSESTree.CallExpression {
+  callee: JestFunctionMemberExpression<FunctionName>;
+}
+
+export interface JestFunctionCallExpressionWithIdentifierCallee<
+  FunctionName extends JestFunctionName
 > extends TSESTree.CallExpression {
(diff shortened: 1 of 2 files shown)


src/rules/tsUtils.ts
```diff
@@ -59,13 +77,10 @@ export const isFunction = (node: TSESTree.Node): node is FunctionExpression =>
...
-): node is JestFunctionCallExpression<HookName> => {
+): node is JestFunctionCallExpressionWithIdentifierCallee<HookName> => {
   return (
-    (node.callee.type === AST_NODE_TYPES.Identifier &&
-      node.callee.name in HookName) ||
-    (node.callee.type === AST_NODE_TYPES.MemberExpression &&
-      node.callee.object.type === AST_NODE_TYPES.Identifier &&
-      node.callee.object.name in HookName)
+    node.callee.type === AST_NODE_TYPES.Identifier &&
+    node.callee.name in HookName
```
G-Rath: I've changed this b/c jest doesn't expect hooks to expose a method property, and this way the two rules that use them don't need to be changed :D

SimenB: I don't follow - you can do `describe.skip` (etc). Or do I misunderstand your comment?

G-Rath: But `describe` isn't a hook - the `HookName` enum is for `beforeAll`, `beforeEach`, `afterEach` & `afterAll`.

That's why it's the only one I made this change to - the others I left as it b/c in the JS version they have member calls (i.e `describe.skip`)

SimenB: hah, good point! There are indeed no identifiers on _hooks_

Issue #268, opened 2019-05-22
"Cannot read property 'meta' of undefined" after update to 22.6.2


hbroer:
`Cannot read property 'meta' of undefined` after update to 22.6.2

```
> eslint --ext .ts,.js ./

TypeError: Cannot read property 'meta' of undefined
    at rules.forEach (C:\Users\[...]\projects\ioc\node_modules\eslint\lib\cli.js:97:34)
    at Map.forEach (<anonymous>)
    at printResults (C:\Users\[...]\projects\ioc\node_modules\eslint\lib\cli.js:96:11)
    at Object.execute (C:\Users\[...]\projects\ioc\node_modules\eslint\lib\cli.js:217:17)
    at Object.<anonymous> (C:\Users\[...]\projects\ioc\node_modules\eslint\bin\eslint.js:78:28)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
```

Greenkeeper is going mad on my repo ;)

Repository: [github.com/owja/ioc](https://github.com/owja/ioc)

hbroer:
Downgrading to `22.5.1` and it works again. I will pin that version.

SimenB:
Should be a new release whenever https://www.githubstatus.com/incidents/nf4gcjtcsqdb is gone

PR #321, opened 2019-07-20
ts-migration/no-test-callbacks

G-Rath:
This is a bit of a tricky one, b/c it's got a bit of "hypothetical" nulls that I can't figure out how to reproduce in action.

As a result, coverage is currently failing, and I've stuck a "please file a github issue" throw as we could use that code to actually test those branches :joy:

We could eliminate at least one check by having `isFunction` narrow to `ArrowFunctionExpression` rather than `ArrowFunctionExpression | FunctionExpression`, as only the latter has `body` defined as being nullable (for reasons I can't figure out - anything I try either has a body or gets thrown out as a syntax error).

Personally, I think they might be typed like that for if you're constructing AST nodes? But I can't find anything solid explanation :/

The other alternative is to use `!`, but personally I'm not really a fan since it mainly means you'll just have messy errors if it ever is `null`.

### DIFF
--- a/src/rules/no-test-callback.ts
+++ b/src/rules/no-test-callback.ts
@@ -30,16 +30,17 @@ export default createRule({
         }
 
         const [argument] = callback.params;
-        const { body } = callback;
-
-        if (!body || !('name' in argument)) {
-          return;
-        }
 
         context.report({
           node: argument,
           messageId: 'illegalTestCallback',
           fix(fixer) {
+            const { body } = callback;
+
+            if (!body) {
+              throw new Error('Unexpected null - please file a github issue');
+            }
+
             const sourceCode = context.getSourceCode();
             const firstBodyToken = sourceCode.getFirstToken(body);
             const lastBodyToken = sourceCode.getLastToken(body);
@@ -47,6 +48,7 @@ export default createRule({
             const tokenAfterArgument = sourceCode.getTokenAfter(argument);
 
             if (
+              !('name' in argument) ||
               !firstBodyToken ||
               !lastBodyToken ||
               !tokenBeforeArgument ||
--- a/src/rules/no-test-callback.ts
+++ b/src/rules/no-test-callback.ts
@@ -37,8 +37,11 @@ export default createRule({
           fix(fixer) {
             const { body } = callback;
 
+            /* istanbul ignore if */
             if (!body) {
-              throw new Error('Unexpected null - please file a github issue');
+              throw new Error(
+                `Unexpected null when attempting to fix ${context.getFilename()} - please file a github issue at https://github.com/jest-community/eslint-plugin-jest`,
+              );
             }
 
             const sourceCode = context.getSourceCode();
@@ -47,6 +50,7 @@ export default createRule({
             const tokenBeforeArgument = sourceCode.getTokenBefore(argument);
             const tokenAfterArgument = sourceCode.getTokenAfter(argument);
 
+            /* istanbul ignore if */
             if (
               !('name' in argument) ||
               !firstBodyToken ||
@@ -54,7 +58,9 @@ export default createRule({
               !tokenBeforeArgument ||
               !tokenAfterArgument
             ) {
-              throw new Error('Unexpected null - please file a github issue');
+              throw new Error(
+                `Unexpected null when attempting to fix ${context.getFilename()} - please file a github issue at https://github.com/jest-community/eslint-plugin-jest`,
+              );
             }
 
             const argumentInParens =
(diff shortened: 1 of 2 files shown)


G-Rath:
@SimenB this might now need rebasing? Not sure why it's got so many extra commits, when it was originally based off `reapply-ts` about 5 minutes ago.

src/rules/no-test-callback.ts
```diff
@@ -20,24 +25,36 @@ export default {
...
-          callback.params.length !== 1
-        ) {
+        if (!isFunction(callback) || callback.params.length !== 1) {
           return;
         }
 
         const [argument] = callback.params;
+        const { body } = callback;
+
+        if (!body || !('name' in argument)) {
```
SimenB: We want to report either way, don't we? If we only use this in the fixer the checks should be moved into it

G-Rath: Yeah, I originally had them there b/c otherwise I have to duplicate the throw, since `body` is used for `firstBodyToken` & `lastBodyToken`.

SimenB:
@G-Rath I think adding a "please file an issue" is fine if we're unable to figure out why it's `null`. What I usually do is run the eslint rule over large codebases using Jest, and see if anything explodes. E.g. Jest itself, webpack, babel, lerna etc.. If all is good, I think you're current approach is fine.

---

That said, it might be worth an issue in https://github.com/typescript-eslint/typescript-eslint if you're unable to to reproduce what the type info tells you. Type comes from https://github.com/typescript-eslint/typescript-eslint/blob/807bc2d4f5dbd62aebfcbd01012d41c37066480f/packages/typescript-estree/src/ts-estree/ts-estree.ts#L485

G-Rath:
@SimenB Awesome. Do you have any secret tips on how to quickly test out the new code? I imagine copying & pasting into `node_modules` should be enough.

SimenB:
I do `yarn link` as it's easier to make adjustments that way. However, copying over the rule and the utils should work (as long as the other repo also has `@typescript-eslint/experimental-utils`, which at least `jest` does)

SimenB:
I've rebased this. Are you able to verify the `body` thing we discussed and either add a test here or an issue at the TS eslint repo?

G-Rath:
I tried to test it out against one of my repos, but for some reason eslint kept erroring with `Cannot read property 'meta' of undefined`.

My IDE somehow was able to run it just fine (which I double checked a couple of times w/ some `throw new Error`s).

The rule worked fine, w/ this code being reported:

```
test('some test', done => {
  expect(false).toBe(true);
  done();
});
```

and "correctly" fixed:

```
test('some test', () => {return new Promise(done => {
  expect(false).toBe(true);
  done();
})});
```

The fixed code violates `jest/no-test-return-statement`, but we can worry about that another time.

For now, I think I'm just going to open up an issue on the TSEslint repo :)

SimenB:
Seems like I'll need to investigate #268 then...

G-Rath:
I can't confirm that my repo was using the latest code - I just fired from the hip by checking out this branch & copying the `lib` after running `yarn build` into `node_modules`.

I played around w/ updating a few things just to be sure.

You can see the project structure I use [in this repo](https://github.com/G-Rath/strongly-typed-event-emitter), just in case it matters that I'm using an extending `.eslintrc` :)

SimenB:
Cool thanks, pushed a fix to `reapply-ts` 🙂

src/rules/no-test-callback.ts
```diff
@@ -20,24 +25,38 @@ export default {
...
         const [argument] = callback.params;
+
         context.report({
           node: argument,
           messageId: 'illegalTestCallback',
           fix(fixer) {
-            const sourceCode = context.getSourceCode();
             const { body } = callback;
+
+            if (!body) {
```
SimenB: could we add an istanbul ignore comment here and revisit if ts eslint change their types?

src/rules/no-test-callback.ts
```diff
@@ -20,24 +25,38 @@ export default {
...
+              throw new Error('Unexpected null - please file a github issue');
```
SimenB:
```suggestion
              throw new Error('Unexpected null - please file a github issue at https://github.com/jest-community/eslint-plugin-jest');
```

src/rules/no-test-callback.ts
G-Rath: We sure can!

The issue got traction pretty quick, and it sounds like it will be changed to not null.

I'm going to try and catch some zzzs before my 4am flight, so am happy for you to make the change and merge it. Otherwise I'll do it when I land tomorrow morning 😊

SimenB: enjoy your flight 😀

PR #322, opened 2019-07-20
ts-migration/no-hooks

G-Rath:
It might be best to rebase `reapply-ts` on `master`, then rebase this PR 😬 

@SimenB can I leave that to you?


SimenB:
>@SimenB can I leave that to you?

fo sho

src/rules/no-hooks.ts
```diff
@@ -0,0 +1,50 @@
...
+    type: 'suggestion',
+  },
+  defaultOptions: [{ allow: [] } as { allow: readonly HookName[] }],
+  create(context, [{ allow }]) {
+    const whitelistedHookNames = allow.reduce((hashMap, value) => {
+      hashMap[value] = true;
+      return hashMap;
+    }, Object.create(null));
+
+    const isWhitelisted = (node: { callee: { name: string } }) =>
```
SimenB:
```suggestion
    const isWhitelisted = (node: JestFunctionCallExpressionWithIdentifierCallee<HookName>) =>
```

SimenB: Could we possibly do `export type JestHookFunction = JestFunctionCallExpressionWithIdentifierCallee<HookName>` or something? 🙂

G-Rath: > Could we possibly do...

fo sho

>The name is so long... 😛

Yeah, it is - I might do a clean up PR soon, and rename them to something like `Fn` to make them a touch shorter.

Sadly, that's just pretty much what always happens when working w/ AST nodes - there's so much contextual type information that you end up w/ really very verbose but long type names :joy:

SimenB: Yeah, we can go through after everything is migrated. I bet we'll see some patterns or something we can abstract out to make it a bit more readable at that point

PR #324, opened 2019-07-20
ts-migration/no-if

G-Rath:


src/rules/no-if.ts
```diff
@@ -1,36 +1,54 @@
...
-const isTestArrowFunction = node =>
+const testCaseNames = new Set<string | null>([
+  ...Object.keys(TestCaseName),
+  'it.only',
+  'it.skip',
+  'test.only',
+  'test.skip',
+]);
+
+const isTestArrowFunction = (node: TSESTree.Node) =>
```
SimenB: This function is only used by `ArrowFunctionExpression`, so you can do `node: TSESTree.ArrowFunctionExpression` and ditch the `undefined` and `node.type` checks

src/rules/no-if.ts
```diff
@@ -1,36 +1,54 @@
-import { getDocsUrl, isTestCase } from './util';
+import { TestCaseName, createRule, getNodeName, isTestCase } from './tsUtils';
+import {
+  AST_NODE_TYPES,
+  TSESTree,
+} from '@typescript-eslint/experimental-utils';
 
-const isTestArrowFunction = node =>
+const testCaseNames = new Set<string | null>([
```
SimenB: null? seems wrong

src/rules/no-if.ts
```diff
@@ -1,36 +1,54 @@
...
+    type: 'suggestion',
   },
-
+  defaultOptions: [],
   create(context) {
-    const stack = [];
+    const stack: Array<boolean> = [];
 
-    function validate(node) {
+    function validate(node: TSESTree.Node) {
```
SimenB:
```suggestion
    function validate(node: TSESTree.ConditionalExpression | TSESTree.IfStatement) {
```

src/rules/no-if.ts
G-Rath: Yeah that's just b/c of how the typings work in relation to `#has`.

Basically, it's defined like this:

```
interface Set<T> {
  has(item: T): boolean;
}
```

That means if `T` is just `string`, you can't do `has` checks against things like `null`, which is a possible return of `getNodeName` :)

Originally I had `getNodeName(/* ... */) || ''`, but since that's a runtime solution, it's expected to be tested for by branch coverage :joy:

src/rules/no-if.ts
G-Rath: Actually apparently I can leave it as `TSESTree.Node`, but yeah much better!

SimenB: no need to widen the type when we don't need it 🙂

G-Rath: Given the functions name (`isTestArrowFunction`), I think it should: either it should test that the type is `ArrowFunctionExpression`, or it should test if a given `ArrowFunctionExpression ` is a "test" arrow function.

PR #327, opened 2019-07-20
ts-migration/consistent-test-it

G-Rath:


src/rules/consistent-test-it.ts
```diff
@@ -26,7 +30,14 @@ export default {
         additionalProperties: false,
       },
     ],
+    type: 'suggestion',
   },
+  defaultOptions: [
+    { fn: 'test', withinDescribe: 'it' } as {
+      fn?: string;
```
SimenB: can we use the enum instead of `string`?

src/rules/consistent-test-it.ts
```diff
@@ -39,6 +50,10 @@ export default {
       CallExpression(node) {
         const nodeName = getNodeName(node.callee);
 
+        if (!nodeName) {
+          return;
```
SimenB: new, but seeing as tests pass I guess it doesn't matter.

G-Rath: It's required by typescript - `getNodeName` could always return null:

```
export const getNodeName = node => {
  function joinNames(a, b) {
    return a && b ? `${a}.${b}` : null;
  }

  switch (node && node.type) {
    case 'Identifier':
      return node.name;
    case 'MemberExpression':
      return joinNames(getNodeName(node.object), getNodeName(node.property));
  }

  return null;
};
```

Ah, the joys of complex ASTs :joy:

G-Rath: > seeing as tests pass I guess it doesn't matter.

I believe this is b/c of the `isDescribe` - the conditions for which `getNodeName` would return `null` are mutually exclusive w/ `isDescribe`, which is why this has never been a problem.

src/rules/consistent-test-it.ts
```diff
@@ -26,7 +30,14 @@ export default {
         additionalProperties: false,
       },
     ],
+    type: 'suggestion',
   },
+  defaultOptions: [
+    { fn: 'test', withinDescribe: 'it' } as {
+      fn?: 'it' | 'test';
```
SimenB: do we need `as const` on the schema?

SimenB: might not need the type info here then

G-Rath: Not as far as I'm aware - the `schema` is json schema, so it's not actually used by TS (it'd be cool if it was, but it's far too complex)

src/rules/consistent-test-it.ts
G-Rath: Just so it's written down somewhere, this is the code that can be used to make `getNodeName` to return `null`:

```
[1,2,3].forEach(() => { test("foo") })
```

It's b/c `[1,2,3]` is an `ArrayExpression`, which causes `getNodeName` to bail out.

I needed to know how to do this to implement a test for another rule to get coverage other the line :)

PR #329, opened 2019-07-21
ts-migration/prefer-strict-equals

G-Rath:
After a fair bit of mental headaches and a couple of rewrites, it's finally here - the beginning of converting the `expect<x>` methods!

Rather than try & re-implement the entire behaviour of utility functions in one go, I've opted to pretty much chucking them straight out, and implementing new guards as required by each rule.

This should result in a much smoother conversation process, as well as a more formal & official definition of parts of the AST tree by end. In particular, it'll make it much easier to test the boundaries of rules, as it's very hard to test changes to the conditional checks when the majority of the rules that use them are still in JS.

I've also decided to completely ignore `Literal` vs `Identifier`, opting for now to just say `Literal`s are not supported.

The rational is that it's faster, as once we have all the rules converted, we just need to change the types in two interfaces to be `TSESTree.Identifier | TSESTree.Literal`, and TypeScript will tell us where the implementation needs to be adjusted to support said change.


SimenB:
I like these utils! I don't think we need to port the older ones unless they make sense. While they're battle tested, they also show that they've been tweaked a lot, and a fresh (type safe) start is probably a good idea 🙂

Issue #312, opened 2019-07-19
[new rule] no-snapshots


cartogram:
Wondering if I could contribute [this `no-snapshots` rule](https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/jest/no-snapshots.md) that I wrote for Shopify

SimenB:
For sure, I bet more people would like to use it. Thanks!

benmonro:
this is already supported by just setting the maxSize to zero on no-large-snapshots

NullDivision:
What about a parameter that would still allow snapshots as long as other non-snapshot assertions are present? I'm not completely against snapshots as a guide to how stuff should look overall but I am against them if there are no assertions for actual elements.

markogresak:
> this is already supported by just setting the maxSize to zero on no-large-snapshots

@benmonro I was hoping for that, but it's not true.

Here are the problems:
1. We have to set up linting of `.snap` files. In multi-project workspaces, like one managed by [Nx](https://nx.dev/), it can be tedious to correctly configure each new project (e.g. a simple lib module). 
2. The feedback cycle is too long. I want the devs who use ESLint editor integration to see the error on `toMatchSnapshot` while they are writing the test, not after running the test to generate the `.snap` file and then run eslint.  
3. The error message is meaningless:
    The config:
    ```
    "rules": {
      "jest/no-large-snapshots": ["error", { "maxSize": 0 }]
    }
    ```
    Error reported by ESLint:
    ```
    3:1  error  `103`s should begin with lowercase  jest/no-large-snapshots
    ```
    Is that a bug in the code? The error is better with `{ "maxSize": 1 }`:
    ```
    3:1  error  Expected Jest snapshot to be smaller than 1 lines but was 103 lines long  jest/no-large-snapshots
    ```

I believe the `no-snapshots` rule serves a different rule than `no-large-snapshots`. It forbids the usage of `toMatchSnapshot`, while `no-large-snapshots` allows snapshot testing, but moderates the snapshot size. 

For reference, the URL to docs linked by @cartogram above is broken, it should be this: https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/jest/no-snapshots.md

[...]

PR #330, opened 2019-07-21
feat(rules): no-snapshots

cartogram:
This rule prevents all snapshots methods: `toMatchSnapshot()`, `toMatchInlineSnapshot()`, `toThrowErrorMatchingSnapshot()` and `toThrowErrorMatchingInlineSnapshot()`.

Closes https://github.com/jest-community/eslint-plugin-jest/issues/312

### DIFF
--- a/docs/rules/no-snapshots.md
+++ b/docs/rules/no-snapshots.md
@@ -1,0 +1,32 @@
+# Disallow snapshots (no-snapshots)
+
+Jest’s snapshot feature allows you to assert that a value has not changed from a
+stored value in a previous test. The matchers `toMatchSnapshot`,
+`toMatchInlineSnapshot`, `toThrowErrorMatchingSnapshot` and
+`toThrowErrorMatchingInlineSnapshot` will generate snapshots when used inside
+test blocks.
+
+## Rule Details
+
+Using snapshots will often result in poorly documented, difficult to debug tests
+that encourage writing a single test to cover a broad area of functionality when
+seperate, more specific tests would be preferred. This rule aims to prevent the
+use of jest snapshots.
+
+Example of **incorrect** code for this rule:
+
+```js
+expect(doWork()).toMatchSnapshot();
+```
+
+```js
+expect(doWork()).toMatchInlineSnapshot();
+```
+
+```js
+expect(doWork()).toThrowErrorMatchingSnapshot();
+```
+
+```js
+expect(doWork()).toThrowErrorMatchingInlineSnapshot();
+```
--- a/docs/rules/no-snapshots.md
+++ b/docs/rules/no-snapshots.md
@@ -8,8 +8,8 @@ test blocks.
 
 ## Rule Details
 
-Using snapshots will often result in poorly documented, difficult to debug tests
-that encourage writing a single test to cover a broad area of functionality when
+Using snapshots may result in poorly documented, difficult to debug tests that
+encourage writing a single test to cover a broad area of functionality when
 seperate, more specific tests would be preferred. This rule aims to prevent the
 use of jest snapshots.
 
--- a/src/rules/__tests__/no-snapshots.test.js
+++ b/src/rules/__tests__/no-snapshots.test.js
@@ -1,0 +1,32 @@
+import { RuleTester } from 'eslint';
+import rule from '../no-snapshots';
+
+const ruleTester = new RuleTester();
+
+const errors = [{ messageId: 'noSnapshots' }];
+
+ruleTester.run('no-snapshots', rule, {
+  valid: [
+    {
+      code: `expect(foo).toHaveProperty('foo');`,
+    },
+  ],
+  invalid: [
+    {
+      code: 'expect(foo).toMatchSnapshot();',
+      errors,
+    },
+    {
+      code: 'expect(foo).toMatchInlineSnapshot();',
+      errors,
+    },
+    {
+      code: 'expect(foo).toThrowErrorMatchingSnapshot();',
+      errors,
+    },
+    {
+      code: 'expect(foo).toThrowErrorMatchingInlineSnapshot();',
+      errors,
+    },
+  ],
+});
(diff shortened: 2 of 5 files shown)


src/rules/no-snapshots.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+};
+
+function isSnapshot(name) {
+  return [
+    'toMatchSnapshot',
+    'toMatchInlineSnapshot',
+    'toThrowErrorMatchingSnapshot',
+    'toThrowErrorMatchingInlineSnapshot',
+  ].some(method => method === name);
```
G-Rath: You can just use `includes` for this

src/rules/no-snapshots.js
```diff
@@ -0,0 +1,37 @@
...
+    messages: {
+      noSnapshots:
+        'Do not use {{method}} or related methods that generate snapshots.',
+    },
+  },
+
+  create(context) {
+    return {
+      CallExpression({ callee: { property } }) {
+        if (isSnapshot(property && property.name)) {
```
G-Rath: It'd be better to do the existence check outside of the method call, since that way you'd only do the check if the `property` is truthy.

src/rules/no-snapshots.js
```diff
@@ -0,0 +1,37 @@
...
+
+export default {
+  meta: {
+    docs: {
+      description: 'Disallow snapshots',
+      uri: getDocsUrl(__filename),
+    },
+    messages: {
+      noSnapshots:
+        'Do not use {{method}} or related methods that generate snapshots.',
```
G-Rath: Could this have inner spaces, to be consistent with the other rules? (i.e `{{ method }}`)

src/rules/__tests__/no-snapshots.test.js
```diff
@@ -0,0 +1,32 @@
...
+import rule from '../no-snapshots';
+
+const ruleTester = new RuleTester();
+
+const errors = [{ messageId: 'noSnapshots' }];
+
+ruleTester.run('no-snapshots', rule, {
+  valid: [
+    {
+      code: `expect(foo).toHaveProperty('foo');`,
```
G-Rath: Typically we omit the braces in these tests when they're not needed.

src/rules/__tests__/no-snapshots.test.js
```diff
@@ -0,0 +1,32 @@
+import { RuleTester } from 'eslint';
+import rule from '../no-snapshots';
+
+const ruleTester = new RuleTester();
+
+const errors = [{ messageId: 'noSnapshots' }];
```
G-Rath: Could I ask you to inline this like how the other rules do? 

I'm not against this, but when this rule is converted to TypeScript, I'll have explicitly type the array or inline the object so that TS can implicitly infer the type, so doing it now will save me a bit of extra work later :)

docs/rules/no-snapshots.md
```diff
@@ -0,0 +1,32 @@
...
+
+Jest’s snapshot feature allows you to assert that a value has not changed from a
+stored value in a previous test. The matchers `toMatchSnapshot`,
+`toMatchInlineSnapshot`, `toThrowErrorMatchingSnapshot` and
+`toThrowErrorMatchingInlineSnapshot` will generate snapshots when used inside
+test blocks.
+
+## Rule Details
+
+Using snapshots will often result in poorly documented, difficult to debug tests
```
SimenB:
```suggestion
Using snapshots may result in poorly documented, difficult to debug tests
```

src/rules/no-snapshots.js
```diff
@@ -0,0 +1,37 @@
...
+            node: property,
+            data: { method: property.name },
+          });
+        }
+      },
+    };
+  },
+};
+
+function isSnapshot(name) {
```
SimenB: can you keep this function outside of the export?

src/rules/__tests__/no-snapshots.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+ruleTester.run('no-snapshots', rule, {
+  valid: [
```
G-Rath: Could you chuck in a few more valid tests that might come up during coding?

I'm thinking just things like `expect(a);` and `toMatchSnapshot();`.

I know those are pretty plain and boring tests, but they're super useful in setting the boundaries on what is and isn't expected for a rule, which in turn makes future work easier.

src/rules/no-snapshots.js
cartogram: @SimenB Might be missing something, but isn't it outside the export?

jeysal:
> I wonder if, instead of `no-snapshots` we should have a `ban-matcher` rule which allows you to provide matchers as options?

Hell yeah, if we give it a map of alternatives to suggest it can replace prefer-to-be-null, prefer-to-strict-equal, etc. all at once.
(Just gotta make sure the suggested alternative isn't banned as well lol)

SimenB:
We could do something similar for `prefer-*`, yeah, that'd be good! Not all of those are about replacing matchers though, but also e.g. `expect(arr.length).toBe(2) -> expect(arr).toHaveLength(2)` which is not really the same thing. But getting that into a single rule might make sense

SimenB:
I've landed the PR with the TS migration on master, so regardless of how this rule ends up, it should be rewritten in TS so we don't have to migrate it later 🙂

benmonro:
@SimenB this is already supported by using `no-large-snapshots` with maxSize=0

SimenB:
That seems quite a roundabout way of achieving it. However, it makes me feel more strongly that the correct approach is a `ban-matchers` rule instead where you can pass in what you want to disallow rather than a specific matcher rule

cartogram:
@SimenB a `ban-matcher` rule makes sense to me. Closing this PR

benmonro:
@SimenB did the ban-matcher rule ever get added?

G-Rath:
@benmonro I'm implementing one as we speak :)

Issue #295, opened 2019-07-09
[new rule] prefer-to-throw / consistent-expect-error


chrisblossom:
I would like to see a new rule named `prefer-to-throw`, or `consistent-expect-error` that would _(preferably)_ default to the `.toThrow` syntax.

If `consistent-expect-error` is set to use the `try/catch` syntax, an error should be thrown if `expect.hasAssertions` (or a variation of) has not been set.

```js
const throwSyncError = () => {
    throw new Error('sync example');
};

const throwAsyncError = async () => {
    throw new Error('async example');
};

describe('prefer-to-throw - sync', () => {
    test('eslint pass', () => {
        expect(() => throwSyncError()).toThrow('async example');
    });

    test('eslint error', () => {
        expect.hasAssertions();
        try {
            throwSyncError();
        } catch (error) {
            expect(error.message).toEqual('sync example');
        }
    });
});

describe('prefer-to-throw - async', () => {
    test('eslint pass', async () => {
        await expect(throwAsyncError()).rejects.toThrow('async example');
    });

    test('eslint error', async () => {
        expect.hasAssertions();
        try {
            await throwAsyncError();
        } catch (error) {
            expect(error.message).toEqual('async example');
        }
    });
});
```

jeysal:
Matching any try-catch where the catch block contains an `expect` to recommend using `expect.assertions`/`expect.hasAssertions` seems doable, but I'm not sure about `toThrow` - I don't see a good way to avoid false positives where the assertions in the catch blocks do more than just check the error message

chrisblossom:
> I don't see a good way to avoid false positives where the assertions in the catch blocks do more than just check the error message

I don't know enough about eslint to respond to this concern, but I would accept this as a limitation and document it. I would just use a `// eslint-disable-next-line consistent-expect-error` if there is an actual need to use a `try / catch`.

This would also mean it should not be enabled in the recommended config (or it would enable to `try / catch` syntax, which I don't think is preferred. For example, [`typescript/no-explicit-any`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md) is a great rule to use, but sometimes you need `any`.


> Matching any try-catch where the catch block contains an expect to recommend using expect.assertions/expect.hasAssertions

I think this is a great additional rule and should be enabled as part of the recommended config. This should include a `.catch` for a promise.

jeysal:
[...]

PR #331, opened 2019-07-21
feat(rules): no-try-expect

cartogram:
This rule prevents the use of `expect` inside `catch` blocks.

The following patterns are warnings:

 ```js
it('foo', () => {
  try {
    foo(); // `foo` may be refactored to not throw exceptions, yet still appears to be tested here.
  } catch (err) {
    expect(err).toMatch(/foo error/);
  }
});
 it('bar', async () => {
  try {
    await foo();
  } catch (err) {
    expect(err).toMatch(/foo error/);
  }
});
```

The following patterns are not warnings:

 ```js
it('foo', () => {
  expect(() => foo()).toThrow(/foo error/);
});
 it('bar', async () => {
  await expect(fooPromise).rejects.toThrow(/foo error/);
});
```

Related https://github.com/jest-community/eslint-plugin-jest/issues/295

### DIFF
--- a/docs/rules/no-try-expect.md
+++ b/docs/rules/no-try-expect.md
@@ -49,7 +49,7 @@ it('bar', async () => {
 });
 
 it('baz', async () => {
-  expect(() => foo()).toThrow(
+  await expect(() => foo()).rejects.toThrow(
     expect.objectContaining({ code: 'MODULE_NOT_FOUND' }),
   );
 });
--- a/docs/rules/no-try-expect.md
+++ b/docs/rules/no-try-expect.md
@@ -1,0 +1,42 @@
+# Prevent catch assertions in tests (no-try-expect)
+
+This rule prevents the use of `expect` inside `catch` blocks.
+
+## Rule Details
+
+Expectations inside a `catch` block can be silently skipped. While Jest provides
+an `expect.assertions(number)` helper, Shopify rarely uses this feature. Using
+`toThrow` concisely guarantees that an exception was thrown, and that its
+contents match expectations.
+
+The following patterns are warnings:
+
+```js
+it('foo', () => {
+  try {
+    foo(); // `foo` may be refactored to not throw exceptions, yet still appears to be tested here.
+  } catch (err) {
+    expect(err).toMatch(/foo error/);
+  }
+});
+
+it('bar', async () => {
+  try {
+    await foo();
+  } catch (err) {
+    expect(err).toMatch(/foo error/);
+  }
+});
+```
+
+The following patterns are not warnings:
+
+```js
+it('foo', () => {
+  expect(() => foo()).toThrow(/foo error/);
+});
+
+it('bar', async () => {
+  await expect(fooPromise).rejects.toThrow(/foo error/);
+});
+```
--- a/docs/rules/no-try-expect.md
+++ b/docs/rules/no-try-expect.md
@@ -5,9 +5,9 @@ This rule prevents the use of `expect` inside `catch` blocks.
 ## Rule Details
 
 Expectations inside a `catch` block can be silently skipped. While Jest provides
-an `expect.assertions(number)` helper, Shopify rarely uses this feature. Using
-`toThrow` concisely guarantees that an exception was thrown, and that its
-contents match expectations.
+an `expect.assertions(number)` helper, it might be cumbersome to add this to
+every single test. Using `toThrow` concisely guarantees that an exception was
+thrown, and that its contents match expectations.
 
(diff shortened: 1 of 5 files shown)


src/rules/__tests__/no-try-expect.test.js
```diff
@@ -0,0 +1,66 @@
+import { RuleTester } from 'eslint';
+import rule from '../no-try-expect';
+
+const ruleTester = new RuleTester({
+  parser: require.resolve('babel-eslint'),
+});
+
+const errors = [
```
G-Rath: Could I ask you to inline this like how the other rules do? 

I'm not against this, but when this rule is converted to TypeScript, I'll have explicitly type the array or inline the object so that TS can implicitly infer the type, so doing it now will save me a bit of extra work later :)

src/rules/__tests__/no-try-expect.test.js
```diff
@@ -0,0 +1,66 @@
...
+  {
+    messageId: 'noTryExpect',
+  },
+];
+
+ruleTester.run('no-try-catch', rule, {
+  valid: [
+    {
+      code: `it('foo', () => {
```
G-Rath: Typically we omit the braces in these tests when they're not needed.

docs/rules/no-try-expect.md
```diff
@@ -0,0 +1,42 @@
+# Prevent catch assertions in tests (no-try-expect)
+
+This rule prevents the use of `expect` inside `catch` blocks.
+
+## Rule Details
+
+Expectations inside a `catch` block can be silently skipped. While Jest provides
+an `expect.assertions(number)` helper, Shopify rarely uses this feature. Using
```
SimenB:
```suggestion
an `expect.assertions(number)` helper, it might be cumbersome to add to every single test. Using
```

or something 🙂

docs/rules/no-try-expect.md
```diff
@@ -0,0 +1,42 @@
...
+`toThrow` concisely guarantees that an exception was thrown, and that its
+contents match expectations.
+
+The following patterns are warnings:
+
+```js
```
SimenB: can you add an example where you test `error.code === 'MODULE_NOT_FOUND'` or something? You're not always testing the message of the error.

Should be `expect(() => foo()).toThrow(expect.objectContaining({ code: 'MODULE_NOT_FOUND' }));` in the "not warning" section

src/rules/__tests__/no-try-expect.test.js
```diff
@@ -0,0 +1,66 @@
+import { RuleTester } from 'eslint';
+import rule from '../no-try-expect';
+
+const ruleTester = new RuleTester({
+  parser: require.resolve('babel-eslint'),
```
SimenB: can you use the default parser, but set `ecmaVersion` to something that supports `async-await` (which I assume is why you needed a different parser)

src/rules/__tests__/no-try-expect.test.js
cartogram: TIL!

docs/rules/no-try-expect.md
```diff
@@ -0,0 +1,56 @@
...
+it('foo', () => {
+  expect(() => foo()).toThrow(/foo error/);
+});
+
+it('bar', async () => {
+  await expect(fooPromise).rejects.toThrow(/foo error/);
+});
+
+it('baz', async () => {
+  expect(() => foo()).toThrow(
```
SimenB:
```suggestion
  await expect(() => foo()).rejects.toThrow(
```

or make it non-async

docs/rules/no-try-expect.md
```diff
@@ -0,0 +1,56 @@
+# Prevent catch assertions in tests (no-try-expect)
+
+This rule prevents the use of `expect` inside `catch` blocks.
+
+## Rule Details
+
+Expectations inside a `catch` block can be silently skipped. While Jest provides
+an `expect.assertions(number)` helper, it might be cumbersome to add this to
```
jeysal: Wondering if the rule should check for these and pass if the test uses it, for the cases where one might actually need complicated logic in a catch

SimenB: in that case I think an `eslint-disable-next-line` comments is better rather than try to make the rule too clever.

jeysal: Not if you `eslint-disable` without adding `expect.assertions` - the lint error could suggest adding `expect.assertions` to get rid of the warning and save someone from mindlessly disabling it

PR #335, opened 2019-07-22
ts-migration/prefer-todo

G-Rath:
...and for my next trick: I'll make it possible to call `getNodeName` in a typesafe way that means it'll not return `null`!


src/rules/prefer-todo.ts
```diff
@@ -0,0 +1,117 @@
...
+} from './tsUtils';
+
+function isOnlyTestTitle(node: TSESTree.CallExpression) {
+  return node.arguments.length === 1;
+}
+
+function isFunctionBodyEmpty(node: FunctionExpression) {
+  /* istanbul ignore next https://github.com/typescript-eslint/typescript-eslint/issues/734 */
+  if (!node.body) {
+    throw new Error();
```
SimenB: include some text here

Issue #336, opened 2019-07-22
valid-describe should support describe.each


G-Rath:
As per #334, `valid-describe` currently doesn't support `describe.each`, instead just bailing out.

G-Rath:
I'm wondering if this should be a separate rule? (`valid-describe-each`).

Just b/c the domains are nicely encapsulated - otherwise `valid-describe` is going to get pretty bloated & strung out.

SimenB:
I think conceptually it doesn't make sense to separate them, even though the implementation might be a bit big. But branching inside of it into what's essentially to separate rules can work? Without increasing implementation complexity

SimenB:
It doesn't bail out anymore, it has false positives

G-Rath:
Yeah huh I thought I closed this already 🤷‍♀

SimenB:
False positives is _worse_ than bailing 😛

G-Rath:
> False positives is worse than bailing 😛

Well then make an issue 😜 

I thought I fixed all the false positives?

Issue #334, opened 2019-07-22
jest/valid-describe false positive for descibe.each variant


topaxi:
The rule `jest/valid-describe` errors for the `describe.each` variant:

```javascript
describe.each([...data])('Something', data => {
  // ...
})
```

leads to:

```
  3:15  error  First argument must be name                    jest/valid-describe
  3:15  error  Describe requires name and callback arguments  jest/valid-describe
```

joshuadiablito:
Also have the same error, it's valid: https://jestjs.io/docs/en/api#describeeachtable-name-fn-timeout

G-Rath:
I suspect I know what's causing this, but I can't reproduce this - I managed to have it show up once in my IDE, but it didn't have a stacktrace and went away after a second.

I hate to be a pain, but if you could provide a small repo that reproduced the error, or even a stacktrace, that'd be a great help!

SimenB:
Pasting in the snippet in the OP fails for me.

Diff with failing test:

```diff
diff --git i/src/rules/__tests__/valid-describe.test.ts w/src/rules/__tests__/valid-describe.test.ts
index adfbad5..5f63b49 100644
--- i/src/rules/__tests__/valid-describe.test.ts
+++ w/src/rules/__tests__/valid-describe.test.ts
@@ -37,6 +37,11 @@ ruleTester.run('valid-describe', rule, {
       test('bar', () => {})
     )
     `,
+    `
+    describe.each([...data])('Something', data => {
+      // ...
+    })
+    `,
   ],
   invalid: [
     {
```

G-Rath:
eeekkkk but totally what I expected - I'll try and wipe up a fix asap.

SimenB:
The _real_ fix is to handle `describe.each`. The _quick_ fix is to not recognize it as a `describe` - `each` is pretty different, so I think really fixing it is a bit more work than needed

G-Rath:
One step ahead of you ;)

I'm going to make a quick fix to ignore `each`, and open a new issue to support `each`.

The rule readme actually excludes `each`:

>The following describe function aliases are also validated:
>
>describe
>describe.only
>describe.skip
>fdescribe
>xdescribe

G-Rath:
Sorry my quick fix ended up not being so quick as I got distracted finishing off the conversion for `prefer-todo` :joy:

topaxi:
Wow! Thanks for the quick fix! <3

topaxi:
[...]

PR #337, opened 2019-07-22
Fix 334

G-Rath:
Fxies #334 

I also inverted the `if` condition to reduce indentation.

Implementing support for `describe.each` is tracked by #336


SimenB:
This will release the TS migration as well - 🤞 nothing explodes 😀

PR #341, opened 2019-07-22
fix(util): identify only valid global properties

### LINKED ISSUE #340: False positives for `no-identical-title` rule
## The Problem
Given the following code (simplified version of the use-case): 

```js
it('data.foo should contain bar and baz', () => {
  const data = {
    foo: 'bar_baz',
  };

  const test = {
    content: key => data[key],
  };

  expect(test.content('foo')).to.contain('bar');
  expect(test.content('foo')).to.contain('baz');
});
```

We get the following error:

```error
  11:10  error  Test title is used multiple times in the same describe block  jest/no-identical-title
```

## Why it Happened?

As part of #316 & #317 we now treat `test.<anything>()` as a test case. While only `(test|it).skip`, `(test|it).only` & `(test|it).todo` are Jest valid options.

In a project that works with `it`s and `describe`s it's possible that a user will choose the name `test` for a utility object.

(Sadly our case 🙈 )

While in our project we can change those object names, it will create a better developer experience if `eslint-plugin-jest` will avoid raising false-positive errors in this case.
 
## Proposed Solution

In the following context https://github.com/jest-community/eslint-plugin-jest/pull/317/files#diff-fecc1b14fa96780438948ea994aa4b29R123

Verify that the `node.callee.property.name` is one of the following: `skip|only|todo`.

ranyitz:
This PR fixes #340 (well, only a part of it)

By improving the utilities `isTestCase` and `isDescribe` to return `true` only when used with valid properties (e.g. `test.skip`) in oppose to `test.foo` which is not a valid test case.

### DIFF
--- a/src/rules/__tests__/no-identical-title.test.js
+++ b/src/rules/__tests__/no-identical-title.test.js
@@ -113,6 +113,26 @@ ruleTester.run('no-identical-title', rule, {
         es6: true,
       },
     },
+    {
+      code: [
+        'const test = { content: () => "foo" }',
+        'test.content(`testing backticks with the same title`);',
+        'test.content(`testing backticks with the same title`);',
+      ].join('\n'),
+      env: {
+        es6: true,
+      },
+    },
+    {
+      code: [
+        'const describe = { content: () => "foo" }',
+        'describe.content(`testing backticks with the same title`);',
+        'describe.content(`testing backticks with the same title`);',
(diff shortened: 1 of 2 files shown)


SimenB:
@ranyitz wanna PR this to `tsUtils` as well?

ranyitz:
@SimenB No problems, I'll send a PR soon

PR #344, opened 2019-07-23
ts-migration/no-try-expect

G-Rath:
I had to alter the tests from using `catch` w/o a parameter.

For some reason it would cause a parsing error - I don't know if that's a TS or ESlint support problem, but since it's not critical to the rule (it just changes the `CatchClause#param` property from `Identifier` to `null`), I've omitted it for now.

### DIFF
--- a/src/rules/no-try-expect.ts
+++ b/src/rules/no-try-expect.ts
@@ -1,5 +1,5 @@
 import { TSESTree } from '@typescript-eslint/experimental-utils';
-import { createRule, isTestCase } from './tsUtils';
+import { createRule, isExpectCall, isTestCase } from './tsUtils';
 
 export default createRule({
   name: __filename,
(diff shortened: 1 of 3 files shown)


src/rules/no-try-expect.ts
```diff
@@ -13,13 +16,18 @@ export default {
...
+  defaultOptions: [],
   create(context) {
     let isTest = false;
     let catchDepth = 0;
 
-    function isThrowExpectCall(node) {
-      return catchDepth > 0 && node.callee.name === 'expect';
+    function isThrowExpectCall(node: TSESTree.CallExpression) {
+      return (
+        catchDepth > 0 && 'name' in node.callee && node.callee.name === 'expect'
```
SimenB: Can we use `isExpectCall`?

SimenB: we can 😀

G-Rath: I'll be sure to make sure that doesn't change when cleaning up the new guards

G-Rath:
@SimenB Yeah, I figured it was b/c of something like that, but wanted to keep rolling through the bigger changes first :P

Thanks for handling this one for me!

PR #348, opened 2019-07-24
ts-migration/no-identical-title

G-Rath:
This one doesn't require the new guards, so can be merged ahead of them.


src/rules/no-identical-title.ts
```diff
@@ -0,0 +1,77 @@
...
+          if (currentLayer.testTitles.indexOf(title) !== -1) {
+            context.report({ messageId: 'multipleTestTitle', node });
+          }
+          currentLayer.testTitles.push(title);
+        }
+
+        if (!isDescribe(node)) {
+          return;
+        }
+        if (currentLayer.describeTitles.indexOf(title) !== -1) {
```
SimenB: `includes`

Issue #342, opened 2019-07-22
valid-describe with expect


benmonro:
curious if we can/should modify valid-describe to report an error if it finds an expect in a describe.

for example:

```javascript
describe("some stuff", () => {
  expect("blah").toBe("foo");
}
```

jest won't report that as a test so wondering if it should be reported as an error.  if it should be reported should we put that in valid-describe or in a standalone rule called `no-standalone-expect`

thoughts?

SimenB:
I like this idea. A separate rule throwing if an `expect` is outside of a test case might make sense? Potentially with a flag that just disallows it in hooks and `describe`s (so people can use helper functions for shared assertions), maybe? I don't think it needs to be coupled to `valid-describe`. _Could_ stick it in `valid-expect`, but I think that's already getting super advanced and a separate rule makes more sense. Maybe not, though 🙂

PR #350, opened 2019-07-24
feat(rules): no-standalone-expect

rabelfish:
This rule prevents expects that are not inside of test block or helper function as these expects will not execute.


src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,66 @@
...
+  // functionDeclaration: function func() {}
+  if (func && func.type === 'FunctionDeclaration') {
+    return 'function';
+  } else if (func && isFunction(func) && func.parent) {
+    const expr = func.parent;
+    // arrowfunction or function expr
+    if (expr.type === 'VariableDeclarator') {
+      return 'function';
+      // if it's not a variable, it will be callExpr, we only care about describe
+    } else if (isDescribe(expr as TSESTree.CallExpression)) {
```
G-Rath: Is it possible to eliminate this cast?

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const getBlockType = (stmt: TSESTree.BlockStatement) => {
+  const func = stmt.parent;
+  // functionDeclaration: function func() {}
+  if (func && func.type === 'FunctionDeclaration') {
+    return 'function';
+  } else if (func && isFunction(func) && func.parent) {
+    const expr = func.parent;
+    // arrowfunction or function expr
+    if (expr.type === 'VariableDeclarator') {
```
G-Rath: You can use `AST_NODE_TYPES` for these constants.

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,66 @@
...
+      return 'function';
+      // if it's not a variable, it will be callExpr, we only care about describe
+    } else if (isDescribe(expr as TSESTree.CallExpression)) {
+      return 'describe';
+    } else {
+      return 'callExpr';
+    }
+  } else {
```
G-Rath: You don't need this else - instead you can just `return` :)

src/rules/__tests__/no-standalone-expect-test.js
```diff
@@ -0,0 +1,44 @@
+import { RuleTester } from 'eslint';
+import rule from '../no-standalone-expect';
+
+const ruleTester = new RuleTester({
```
G-Rath: This file should be in typescript too :)

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+    messages: {
+      unexpectedExpect: `Expect must be inside of a test block.`,
+    },
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const callStack: Array<String> = [];
```
G-Rath: It's not a big deal, but typically I've been using square brackets for simple array types (`string[]`).

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,66 @@
...
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Prevents expects that are outside of an it or test block.',
+      recommended: false,
+    },
+    messages: {
+      unexpectedExpect: `Expect must be inside of a test block.`,
```
SimenB: Single quotes, please :)

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,66 @@
+import { createRule, isDescribe, isExpectCall, isFunction } from './tsUtils';
+import { TSESTree } from '@typescript-eslint/experimental-utils';
```
SimenB: Could you have local imports below module imports?

src/rules/no-standalone-expect.ts
SimenB: We should probably lint for this

src/rules/no-standalone-expect.ts
SimenB: +1 for early `return`s rather than `else` or nested `if`

src/rules/no-standalone-expect.ts
SimenB: `(expr.type === 'CallExpression && isDescribe(expr))` should do the trick

src/rules/no-standalone-expect.ts
G-Rath: I was going to say exactly that, as it's been on my mind for a while now - I'll make a PR

src/rules/no-standalone-expect.ts
G-Rath: `if (expr.type === AST_NODE_TYPES .CallExpression && isDescribe(expr))` 😉 

(hmm someone should make an eslint plugin for that... 🤔)

rabelfish: yeah that works. Thanks, not super familiar with typescript

rabelfish:
Hey @SimenB @G-Rath Thanks for the feedback! I think I fixed everything you guys mentioned. Let me know how it looks now!

benmonro:
resolves https://github.com/jest-community/eslint-plugin-jest/issues/342

src/rules/__tests__/no-standalone-expect-test.ts
```diff
@@ -0,0 +1,44 @@
...
+    'describe("a test", () => { it("an it", () => { const func = () => { expect(1).toBe(1); }; }); });',
+    'describe("a test", () => { const func = () => { expect(1).toBe(1); }; });',
+    'describe("a test", () => { function func() { expect(1).toBe(1); }; });',
+    'describe("a test", () => { const func = function(){ expect(1).toBe(1); }; });',
+    'const func = function(){ expect(1).toBe(1); };',
+  ],
+  invalid: [
+    {
+      code: 'describe("a test", () => { expect(1).toBe(1); });',
+      parserOptions: { sourceType: 'module' },
```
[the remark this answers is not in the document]
rabelfish: It does indeed :)

src/rules/__tests__/no-standalone-expect-test.ts
```diff
@@ -0,0 +1,44 @@
...
+import rule from '../no-standalone-expect';
+
+const ruleTester = new TSESLint.RuleTester({
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-standalone-expect', rule, {
```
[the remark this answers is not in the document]
rabelfish: Sure thing. It looks like expect.hasAssertions() will still execute if it is outside of a test block so I don't think that should trigger this rule. expect(1).toBe probably should though.

(luckily for me this is how it already functions :P)

rabelfish:
@SimenB  I seem to be failing the ci because every time I make a commit I get this:

✖   message may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings

Can you explain to me what I need to do to resolve this?

SimenB:
It wants you to make conventional (semantic) commits. See https://www.conventionalcommits.org/

If you want them to be green, a quick `chore: *my thing*` works. Otherwise, don't worry about it as I'll squash merge and fix the commit message when doing so 🙂 As long as linting and tests are green on CI, it's fine

docs/rules/no-standalone-expect.md
```diff
@@ -0,0 +1,60 @@
...
+    expect(1).toBe(1);
+  };
+
+  it('an it', () => {
+    helper();
+  });
+});
+```
+
+\*Note that this rule will not trigger if the helper function is never used even
```
SimenB: why the `\*`?

src/rules/__tests__/no-standalone-expect-test.ts
```diff
@@ -0,0 +1,44 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+import rule from '../no-standalone-expect';
+
+const ruleTester = new TSESLint.RuleTester({
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module',
```
SimenB: do we need this default sourcetype?

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,67 @@
...
+  TSESTree,
+} from '@typescript-eslint/experimental-utils';
+import { createRule, isDescribe, isExpectCall, isFunction } from './tsUtils';
+
+const getBlockType = (stmt: TSESTree.BlockStatement) => {
+  const func = stmt.parent;
+  // functionDeclaration: function func() {}
+  if (func && func.type === AST_NODE_TYPES.FunctionDeclaration) {
+    return 'function';
+  } else if (func && isFunction(func) && func.parent) {
```
SimenB: could you clean up the else after return here?

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,67 @@
...
+          }
+        }
+      },
+      BlockStatement(stmt) {
+        const blockType = getBlockType(stmt);
+        if (blockType) {
+          callStack.push(blockType);
+        }
+      },
+      'BlockStatement:exit'(stmt) {
```
SimenB:
```suggestion
      'BlockStatement:exit'(stmt: TSESTree.BlockStatement) {
```

or is it correctly inferred? If it does, ignore me 🙂

G-Rath: [It won't be inferred](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/experimental-utils/src/ts-eslint/Rule.ts#L221) as it matches `[nodeSelector: string]: RuleFunction | undefined;`.

Only properties that map directly to a `AST_NODE_TYPES` constant will be inferred.

docs/rules/no-standalone-expect.md
rabelfish: Oh I didn't mean the \ just meant a * there

src/rules/no-standalone-expect.ts
SimenB: Merging in master will give lint errors here, btw

src/rules/__tests__/no-standalone-expect-test.ts
rabelfish: We do not, I definitely copied this from someone else... oops.

src/rules/no-standalone-expect.ts
rabelfish: Not sure what you mean, clean up in what way?

docs/rules/no-standalone-expect.md
SimenB: Still here. I guess prettier adds it? `*` is used to make things bold, and when you only have one, prettier makes sure to escape it

src/rules/no-standalone-expect.ts
rabelfish: Oh should I not pull master?

SimenB: No, please in master, and you'll get lint errors due to https://eslint.org/docs/rules/no-else-return

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,70 @@
...
+    return {
+      CallExpression(node) {
+        if (isExpectCall(node)) {
+          const parent = callStack[callStack.length - 1];
+          if (!parent || parent === 'describe') {
+            context.report({ node, messageId: 'unexpectedExpect' });
+          }
+        }
+      },
+      BlockStatement(stmt: TSESTree.BlockStatement) {
```
SimenB: You don't need the type declaration here as it'll be inferred correctly. It's just needed for the `:exit` one below. Doesn't hurt of course 🙂

docs/rules/no-standalone-expect.md
```diff
@@ -0,0 +1,67 @@
+# No standalone expect in a describe block (no-standalone-expect)
+
+Prevents expects outside of a test or it block. An expect within a helper
+function (but outside of a test or it block) will not trigger this rule.
```
SimenB: Could you wrap `test` and `it` in backticks? As well as `expect`

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,70 @@
+import {
+  AST_NODE_TYPES,
+  TSESTree,
+} from '@typescript-eslint/experimental-utils';
+import { createRule, isDescribe, isExpectCall, isFunction } from './tsUtils';
+
+const getBlockType = (stmt: TSESTree.BlockStatement) => {
```
SimenB: Could you type the return value here? A union type would be great :)

src/rules/no-standalone-expect.ts
rabelfish: good to know :)

src/rules/no-standalone-expect.ts
rabelfish: Ah, sorry, I misunderstood. I have master pulled and actually am not seeing errors. I am pretty sure I already removed all of the else returns :)

SimenB: huh, it seems you do! Weird... I fixed them myself, see the commit for what I meant 🙂

SimenB:
Oh, missing mention in the main readme

src/rules/no-standalone-expect.ts
```diff
@@ -4,21 +4,26 @@ import {
 } from '@typescript-eslint/experimental-utils';
 import { createRule, isDescribe, isExpectCall, isFunction } from './tsUtils';
 
-const getBlockType = (stmt: TSESTree.BlockStatement): String | false => {
+const getBlockType = (
+  stmt: TSESTree.BlockStatement,
+): 'function' | 'describe' | 'callExpr' | false => {
   const func = stmt.parent;
+
+  if (!func) {
```
rabelfish: @SimenB So it looks like this is the causing the test coverage to not be 100%, do you know of any way to test this? when would the BlockStatment not have a parent?

SimenB:
That covered one line, still missing the other 😬 

If it's just a type error and you're unable to reproduce it in a test, feel free to just throw an error and ask people who hit it to open up an issue. Like https://github.com/jest-community/eslint-plugin-jest/blob/master/src/rules/no-test-callback.ts#L40-L45

---

FWIW, I ran this against the Jest code base, and I found a bug. Every expect in this file is reported as a violation: https://github.com/facebook/jest/blob/cb322d19ade00134d5a16c3be7db0db41766a855/packages/jest-get-type/src/__tests__/getType.test.ts

rabelfish:
@SimenB Yeah, I know it only covered the one. I am not sure how to cover the other. Do you know of any BlockStatement that would not have a parent?

Also yes that would fail... I did not think of that case. Darn es6... let me see if I can handle that.

SimenB:
I don't know... Just throw and explicit error there and ask for issues 🙂

src/rules/no-standalone-expect.ts
G-Rath: You might be able to sneak around this by moving [line 19](https://github.com/jest-community/eslint-plugin-jest/pull/350/files/9656d908b7a2526ac01e52652948d2cb16b9f0c2..d24f4afeeb6955569081938ac5b5bf8364654829#diff-8a6284e5ebef9b04182b8b9972128c4bR19) up (you also might have to put the `func.parent` before the `isFunction` check).

SimenB: I moved it out explicitly - if we can't cover that condition I think we should be explicit about it rather than combining conditions and "fooling" coverage

G-Rath: Ah fair enough.

tbh I think I was thinking of methods anyway:

```
export const isExpectCall = (node: TSESTree.Node): node is ExpectCall =>
  node.type === AST_NODE_TYPES.CallExpression &&
  isExpectAccessor(node.callee) &&
  node.parent !== undefined;
```

I don't think `node.parent !== undefined` is ever not true, but coverage doesn't complain b/c it's not in an `if` branch.

> we should be explicit about it

Absolutely!

rabelfish:
okay, so, test coverage is 100% and I handled that bug!! However, it did prompt me to realize that a case like this should also pass (as a helper function) and it does not: 

```js 
const func = () => expect(1).toBe(1);
```

benmonro:
FYI, i'm enjoying watching all this unfold before my eyes (@rabelfish sits 1 row in front of me). 🍿 good times.

rabelfish:
@SimenB okay I did it actually!! However I added some more code so I am sure you will have issues with it again... 😛

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,103 @@
...
+  createRule,
+  isDescribe,
+  isExpectCall,
+  isFunction,
+  isTestCase,
+} from './tsUtils';
+
+const getBlockType = (
+  stmt: TSESTree.BlockStatement,
+): 'function' | 'describe' | 'callExpr' | false => {
```
G-Rath: I think I'd prefer returning `null` instead of `false`; in particular, TS treats `null` & `undefined` a bit differently that can be useful vs `false`.

Also would be nice to have this as type alias :)

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,103 @@
...
+    const expr = func.parent;
+    // arrowfunction or function expr
+    if (expr.type === AST_NODE_TYPES.VariableDeclarator) {
+      return 'function';
+    }
+    // if it's not a variable, it will be callExpr, we only care about describe
+    if (expr.type === AST_NODE_TYPES.CallExpression && isDescribe(expr)) {
+      return 'describe';
+    }
+    return false;
```
G-Rath: You don't need this `return`, since it's the same as the one outside the `if` :)

SimenB:
@G-Rath I fixed both of your comments.

---

@rabelfish this is really starting to look good. Funnily, the last couple of fixes you made ended up reporting 62 violations in jest's code base! 😅 Some of the are legitimate, but I also found a couple of false for `test.each`: https://github.com/facebook/jest/blob/cb322d19ade00134d5a16c3be7db0db41766a855/packages/jest-util/src/__tests__/isPromise.test.ts#L10-L17

rabelfish:
@SimenB Ah, I think I assumed isTestCase in utils would catch an each as well.

SimenB:
Yeah it _should_, but we (by that I mean @G-Rath 😀) haven't gotten to dealing with `.each` yet (#336 is about `describe`, but the same could be said about `test/it.each`).

rabelfish:
Hmm so it actually looks like the utility functions work fine but the way the ast tree traverses the each, it actually exits the each call expression before getting to the expect. So the each is added onto my call stack, it is just popped off before it traverses into the callback. which is unexpected. The it.only and .skip work just fine

rabelfish:
@SimenB @G-Rath okay so it the it.each is working but I am having some typescript trouble. If one of you could help me with that I'd super appreciate it 😃

G-Rath:
I'm currently away from my computer right now, so while I can still help Ill be doing so from memory.

In regards to the utils: not entirely sure what you mean, but also haven't been following the last couple of bumps on this PR.

If it helps, the ast structure is such that the expect is always at the bottom: the topmost node is typically the right most "thingy".

At some point I want to have a talk about what we consider what, in terms of matching for rules - i.e ".toBe" (member expression) vs .toBe() (call expression)" - some rules you definitely want to match the CE, but others we could consider either legal but that's a talk for after we've finished converting.

I am in the process of rewriting the expect utility matchers, so that they're now doing a parsing thing that returns a handy object like this:

```
interface Expectation<
  Modifier extends ExpectModifier = ExpectModifier,
  Matcher extends string = string,
>{
  expect: ExpectCall;
  modifier?: ExpectMember<Modifier>;
  matcher?: ExpectMember<Matcher>;
}
```

I've actually finished these, but still in the process of converting the remaining rules - if I get the time I should have them done by this weekend.

But that won't help you now, so I'll take a look at your PR 😂

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,119 @@
...
+    },
+    messages: {
+      unexpectedExpect: 'Expect must be inside of a test block.',
+    },
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const callStack: Array<'test' | 'function' | 'describe' | 'arrowFunc'> = [];
```
G-Rath: I think let's make this an enum 😂

src/rules/no-standalone-expect.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+  return null;
+};
+
+const isEach = (node: TSESTree.CallExpression): boolean => {
+  if (
+    node &&
+    node.callee &&
+    node.callee.callee &&
```
G-Rath: You need to check the type of the node before you do this, as only `CallExpression` nodes have `callee`

rabelfish: Ah, makes sense. I just thought checking if they existed would be enough, will address these two things tomorrow 😄

G-Rath: That's actually a "is truthy" check :)

You can check property existence via `in`: `'callee' in node.callee`, which can be useful when there are multiple expressions that could have that property (and that we don't care about the type specifically).

However since in this case `callee` is only found on `CallExpression` nodes, it's best to check for that type specifically.

SimenB:
@rabelfish I fixed the type error. Tested again on Jest's code base, and found another false positive. I added a failing test 🙂 We're getting really close!

rabelfish:
just kidding ignore that push.....

benmonro:
@SimenB FYI looks like @rabelfish fixed it, looks like build fails are just git related, we good to merge?  we're doing a dog and pony show later today and would love to have @rabelfish show this off. :)

SimenB:
>we good to merge?

yeah, I was literally just verifying there were no false positives in Jest's code base

benmonro:
cool, thanks!

SimenB:
Thank you _so_ much @rabelfish, this is a fantastic new rule. Thanks for being patient with me and working through the review!

rabelfish:
@SimenB Thank you so much!! I appreciate the thoroughness 😄

swissspidy:
Not sure if I'm wrong, but this doesn't seem to work well with `jest-each`.

I'm using it like this as described in the project's documentation:

```js
each([[1, 1, 2], [1, 2, 3], [2, 1, 3]]).test(
  'returns the result of adding %d to %d',
  (a, b, expected) => {
    expect(a + b).toBe(expected);
  },
);
```

And now get an error caused by this rule.

The tests in this PR seem to cover the `it.each()` case, but not the `each().it()` case.

Please let me know if I'm doing something wrong, thanks.

SimenB:
Huh, I didn't consider extensions when testing this. Could you open up a new issue?

Also, any reason why you use jest-each? With jest 23 it's built-in via test.each etc

G-Rath:
@swissspidy @SimenB I've made #354 to cover this - I think this rule should follow the pattern of `expect-expect` with it's `assertFunctionNames`, as much as possible (and within reason) for the simple fact that this is Javascript - there's any number of interesting patterns people could use.

swissspidy:
Thanks!

In the meantime I‘ll see whether I can drop jest-each :-)

SimenB:
Unless you're stuck on Jest 22 there's no reason to keep using jest-each directly - the module is part of the jest monorepo and is baked into the globals jest exposes

Mouvedia:
I wish this rule could detect `(process.platform === 'linux' ? it : it.skip)('should test linux code only', () => { […]` as passing.

SimenB:
Can you file a bug report?

PR #351, opened 2019-07-24
Update linting rules

G-Rath:
Did this via editor to prevent having to shelf changes locally just to checkout master :joy:

Feel free to push up prettier or other formatting changes if required.

These are the rules I recommend having, feel free to ask questions :)


G-Rath:
Ugh except I just realised that travis will fail b/c of eslint now, so I'll have to check it out locally to fix.

I'll do that sometime soon.

G-Rath:
I've removed `@typescript-eslint/no-unnecessary-type-assertion` & `@typescript-eslint/no-unnecessary-qualifier` b/c while very cool & useful, they require type information, which means linting doubles in time.

It's not the end of the world, but something I figured you guys could decide since you're paying for the pipeline :P

.eslintrc.js
```diff
@@ -25,6 +25,10 @@ module.exports = {
     es6: true,
   },
   rules: {
+    '@typescript-eslint/array-type': ['error', 'array-simple'],
```
jeysal: In Jest we have this set to `generic` for consistency

SimenB: consistency with what? (I don't care either way, and since this is fixable it really doesn't matter what we go with, as long as we lint for one or the other 🙂)

G-Rath: My two-cents: I've found `array-simple` to look rather nice, and gives you the best of both worlds in terms of readability vs extra typing/characters/line length.

Just for the record, here's the difference:

```
// 'generic' - always use "Array", never "[]"
declare const arr: Array<string | number>;
declare const arr: Array<MyObj<string>>;
declare const arr: Array<string>;

// 'array-simple' - always use '[]', unless the generic type is "complex" (uses generics, union, etc)
declare const arr: Array<string | number>;
declare const arr: Array<MyObj<string>>;
declare const arr: string[];

// 'simple' - always use `[]`, never `Array`
declare const arr: (string | number)[];
declare const arr: MyObj<string>[];
declare const arr: string[];
```

I use to use `generic`, but got sick of having to go to the start of the type everytime I wanted to make thing like a `string` an array - w/ `array-simple` you can just keep typing when doing quick speedy type definitions.

jeysal: Just thought it'd be nice to have a code base that is easy to get into if you've already worked on Jest itself. Don't care much either, I always write it in `array` and let it autofix anyway 😄

PR #352, opened 2019-07-25
valid-expect does not work with then - Closes #347

### LINKED ISSUE #347: valid-expect does not work with then
Hello, I have a problem with `jest/valid-expect` rule, it seems to not work when I put a `.then` after a `.rejects` or a `.resolves` like this:

```
return expect(functionReturningAPromise()).resolves.toEqual(1).then(() => {
    expect(someMock).toHaveBeenCalledTimes(1);
});
```

When I disable the rule `jest/valid-expect` or I remove the `.then` it works well.

yatki:
resolves #347


yatki:
@SimenB actually I added some comments with https://github.com/yatki/eslint-plugin-jest/commit/50dcd02290fa053241217fb3ebfb0eb5ad3b4585 would you like to pull them maybe ?

SimenB:
For sure! Mind opening up a new PR? Some comments explaining what's going on in this rule seems like a great idea

PR #355, opened 2019-07-29
Update .gitignore to include .idea/

G-Rath:
b/c I'm sick of it not being ignored.

I figured I'd take the safe route of a PR rather than committing directly to master 😬

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 node_modules/
 coverage/
 lib/


SimenB:
I recommend adding it to a global gitignore file, that way repos you work in don't need it. Doesn't hurt, of course!

G-Rath:
Problem with that it that I commit it in some repos - Some of the files are designed to be committed, but it's not of much use if not everyone is using that IDE.

But yes I should have a clean of my global .gitignore :joy:

PR #358, opened 2019-07-31
methodName: ensure method(node) returns truthy before trying to use return

### LINKED ISSUE #357: Crashes when `expect` with no following statement.
If you have code like `expect()` without anything after, like `toBe`, then it crashes the linter. It shouldn't crash the linter. I imagine Jest would do enough to notify you that something is wrong, or at least the linter should indicate that the following statement is missing.

Sample code:

```js
test('test', () => {
  expect('a');
})
```

Which results in the following error when linting:

```
TypeError: Cannot read property 'name' of undefined
Occurred while linting asdf.test.js:2
    at methodName (/usr/local/lib/node_modules/eslint-plugin-jest/lib/rules/util.js:66:40)
    at expectToBeCase (/usr/local/lib/node_modules/eslint-plugin-jest/lib/rules/util.js:42:140)
    at CallExpression (/usr/local/lib/node_modules/eslint-plugin-jest/lib/rules/prefer-to-be-null.js:25:45)
    at listeners.(anonymous function).forEach.listener (/usr/local/lib/node_modules/eslint/lib/util/safe-emitter.js:45:58)
    at Object.emit (/usr/local/lib/node_modules/eslint/lib/util/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/usr/local/lib/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/usr/local/lib/node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (/usr/local/lib/node_modules/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (/usr/local/lib/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:632:23)
```

G-Rath:
Fixes #357 

It's a simple quick fix, as the underlying problem is fixed by the upcoming ts conversation stuff.

### DIFF
--- a/src/rules/util.js
+++ b/src/rules/util.js
@@ -89,7 +89,7 @@ export const method2 = node => node.parent.parent.property;
 
 const methodName = node => method(node) && method(node).name;
 
-const methodName2 = node => method2(node).name;
+const methodName2 = node => method2(node) && method2(node).name;
 
 export const argument = node =>
   node.parent.parent.arguments && node.parent.parent.arguments[0];
--- a/src/rules/util.js
+++ b/src/rules/util.js
@@ -87,7 +87,7 @@ export const method = node => node.parent.property;
 
 export const method2 = node => node.parent.parent.property;
 
-const methodName = node => method(node).name;
+const methodName = node => method(node) && method(node).name;
 
 const methodName2 = node => method2(node).name;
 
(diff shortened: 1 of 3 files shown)


src/rules/util.js
```diff
@@ -87,7 +87,7 @@ export const method = node => node.parent.property;
 
 export const method2 = node => node.parent.parent.property;
 
-const methodName = node => method(node).name;
+const methodName = node => method(node) && method(node).name;
 
 const methodName2 = node => method2(node).name;
```
SimenB: Same here? `expect(something).not;` or something like that?

G-Rath: I could do - I don't think it's needed per say b/c all the rules that use it check using `method` first, but I'll add it in anyway :)

PR #214, opened 2019-01-18
chore: add changelog to repo

SimenB:
They require node 8, which makes it a bit of a pain to manage.

The diff here is essentially added `@semantic-release/changelog` in the middle of the default plugin configuration. No idea if it works 📿

PR #360, opened 2019-08-01
chore: remove misleading changelog

MiniGod:
The `CHANGELOG.md` file is outdated and serves no purpose as far as I can tell.

Up to date changelog available on the [releases page](https://github.com/jest-community/eslint-plugin-jest/releases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,0 @@
-# [22.2.0](https://github.com/jest-community/eslint-plugin-jest/compare/v22.1.3...v22.2.0) (2019-01-29)
-
-
-### Features
-
-* **rules:** add prefer-todo rule ([#218](https://github.com/jest-community/eslint-plugin-jest/issues/218)) ([0933d82](https://github.com/jest-community/eslint-plugin-jest/commit/0933d82)), closes [#217](https://github.com/jest-community/eslint-plugin-jest/issues/217)


jeysal:
Unfortunately it seems like the 22.2.0 entry is [missing in the releases](https://github.com/jest-community/eslint-plugin-jest/releases?after=v22.2.1). Not sure if there's a way to add it there now

G-Rath:
Personally I'm a big fan of changelogs, as they're git-host-agnostic.

Would it be possible for semantic bot to generate the changelog?

SimenB:
It is, and is why it exists. See #214, 924daedc7bdc00a757bb534138bd4ca65b5a281f and d912a70286fa13dd65f121719c91ed21ca0da080. I'd prefer to fix it :)

G-Rath:
Ah right - so if I understand correctly, we've waiting until we drop support for node 6 before we can fix it?

SimenB:
I don't remember why I reverted it tbh. I think it might've been just bad config, and a revert was the quickest fix

G-Rath:
Fair enough - I've made, and in closing in favor of, #361. I probably won't touch it for now as I've not used semantic-release before 😬

@MiniGod thanks for bring this to our attention - feel free to have a crack at it if you'd like, but I suspect it might require release privileges :/

PR #364, opened 2019-08-05
feat(rules): no-expect-resolves

eranshabi:
This rule prevents the use of `await expect(promise).resolves` in favor of `expect(await promise)`,
for consistency and readability.


src/rules/__tests__/no-expect-resolves.test.ts
```diff
@@ -0,0 +1,24 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+import rule from '../no-expect-resolves';
+
+const ruleTester = new TSESLint.RuleTester({
+  parserOptions: {
+    ecmaVersion: 2017,
+  },
+});
+
+ruleTester.run('no-expect-resolves', rule, {
```
SimenB: this rule should probably have more tests, but I ran it against Jest and found no errors (at least no false positives and no crashes)

SimenB:
Thanks for the contribution @eranshabi!

mrtnzlml:
@SimenB Hi, I have just a silly question - what is the recommended style? Is it `await expect(promise).resolves` or `expect(await promise)`? Just wondering why does this rule exist with such defaults. Maybe it's because `resolves` is going to be deprecated? Thanks! :)

G-Rath:
@mrtnzlml I'm no Simen, but can weight in :)

While I'm not part of the jest team, I don't think they'll be deprecating `resolves`, unless Promises go away.

I believe the main reason for this rule is that Promise testing can be a bit of a mess, and it's easy to write tests that pass, but don't actually test the promise correctly (i.e by not calling `done`, or returning a promise).

`await` removes this problem - while it still suffers from similar mess to `Promises` of that there's a couple of different ways that you can use it to get the desired results, its usage means a promise will *always* be returned and so your tests will at least not silently pass.

So the idea behind this rule is that it's typically best to try to get back to testing a synchronous value, as it tends to be less edge-casey, which is why `expect(await <value>)` is prefered over `await expect(<value>).resolves`.

The former is testing a synchronous value, in the same as:

```
Promise.resolves(<value>)
             .then(value => expect(value).<matcher>)
```

The latter is testing an async value, that `expect`/`jest` has to internally `await` on.

Remember that `await` was only added in Node 7.5, and it's still something a lot of people are learning to use :)

Hope that clears things up :D

SimenB:
We won't be removing `resolves` as its sibling `rejects` is not ergonomic to use with `await`. You could do something like `expect(await promise.then(res => Promise.reject(res), res => res)).toEqual(someRejectedValue)`, but it's not really readable compared to `await expect(promise).rejects.toEqual(someRejectedValue)`. Could of course extract that into a helper, but you might not want abstractions like that in your tests. (the alternative is a try-catch, which has its own caveats, like asserting that it actually rejects).

As to what you should use - up to you! I _generally_ prefer `expect(await promise)` but if I'm testing a bunch of resolving and rejecting promises at once it might be easier to read if the assertions are kept in the same style. I merged the rule since I personally prefer less library boilerplate, but this rule won't be made part of the recommended rules as that's purely a personal preference.

---

Note that after `resolves/rejects` was added to Jest (https://github.com/facebook/jest/pull/3068), we've added async matchers (https://github.com/facebook/jest/pull/5919), so you could add a custom `toRejectAndEqual` matcher or whatever you want. However, `resolves` and `rejects` won't go away as they allow you to make _any_ matcher, custom or bundled, work with promises instead of having to reimplement existing sync matchers.

Hope that clears things up 🙂

PR #365, opened 2019-08-05
fix(prefer-spy-on): only check for prefer-spy-on within `test` blocks

### LINKED ISSUE #209: `prefer-spy-on` should ignore `jest.fn` outside of `test` blocks
Per [this](https://github.com/jest-community/eslint-plugin-jest/issues/185#issuecomment-439673024), the new rule should not recommend replacing `jest.fn` with `jest.spyOn` when used outside of `test` blocks, since `jest.spyOn` can't even be used outside of test blocks.

garyking:
Fixes #209.

### DIFF
--- a/src/rules/prefer-spy-on.ts
+++ b/src/rules/prefer-spy-on.ts
@@ -69,6 +69,17 @@ export default createRule({
 
         if (!jestFnCall) return;
 
+        // The `jest.fn()` must be in a `test`/`it` block in order for us to
+        // report it.
+        const ancestors = context.getAncestors();
+        const isInTestBlock = ancestors.some(
+          ancestor =>
+            ancestor.type === AST_NODE_TYPES.CallExpression &&
+            ['it', 'test'].includes(ancestor.callee.name),
+        );
+
+        if (!isInTestBlock) return;
+
         context.report({
           node,
           messageId: 'useJestSpyOn',
--- a/src/rules/prefer-spy-on.ts
+++ b/src/rules/prefer-spy-on.ts
@@ -2,12 +2,7 @@ import {
   AST_NODE_TYPES,
   TSESTree,
 } from '@typescript-eslint/experimental-utils';
-import {
-  DescribeAlias,
-  TestCaseName,
-  createRule,
-  getNodeName,
-} from './tsUtils';
+import { createRule, getNodeName, isDescribe, isTestCase } from './tsUtils';
 
 const findNodeObject = (
   node: TSESTree.CallExpression | TSESTree.MemberExpression,
@@ -81,8 +76,7 @@ export default createRule({
         const isInTestBlock = ancestors.some(
           ancestor =>
             ancestor.type === AST_NODE_TYPES.CallExpression &&
-            (ancestor.callee.name in DescribeAlias ||
-              ancestor.callee.name in TestCaseName),
+            (isDescribe(ancestor) || isTestCase(ancestor)),
         );
 
         if (!isInTestBlock) return;
--- a/src/rules/prefer-spy-on.ts
+++ b/src/rules/prefer-spy-on.ts
@@ -2,7 +2,12 @@ import {
   AST_NODE_TYPES,
   TSESTree,
 } from '@typescript-eslint/experimental-utils';
-import { createRule, getNodeName } from './tsUtils';
+import {
+  DescribeAlias,
+  TestCaseName,
+  createRule,
+  getNodeName,
+} from './tsUtils';
 
 const findNodeObject = (
   node: TSESTree.CallExpression | TSESTree.MemberExpression,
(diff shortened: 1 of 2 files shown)


G-Rath:
Thanks for the PR!

However you can use `spyOn` outside of `test` blocks:

```
describe('when the current date is before techweek', () => {
  beforeEach(() => jest.spyOn(Date, 'now').mockReturnValue(TECHWEEK_END_TIME - 10000));
   /* ... */
});
```

I don't think it's quite correct to say that it can only be used inside a `test` block.

I'd be happy w/ this being added if it was an opt-in behaviour, but I don't think it should be the default 😬 

A quick glance of the related material (comments & co) it seems to me it's not valid to use `jest.spyOn` outside of Jest, rather than outside of `test` blocks.

Another very easy way to solve this would be to use a nested eslint config, such as I do [in this repo](https://github.com/G-Rath/strongly-typed-event-emitter).

src/rules/prefer-spy-on.ts
```diff
@@ -69,6 +69,17 @@ export default createRule({
 
         if (!jestFnCall) return;
 
+        // The `jest.fn()` must be in a `test`/`it` block in order for us to
+        // report it.
+        const ancestors = context.getAncestors();
+        const isInTestBlock = ancestors.some(
+          ancestor =>
+            ancestor.type === AST_NODE_TYPES.CallExpression &&
+            ['it', 'test'].includes(ancestor.callee.name),
```
G-Rath: I think this should use the `utils` enum, so that it also considers other forms such as `xtest` & `xit` as test blocks.
(Can't remember the enum name off the top of my head :joy:)

garyking: How do I match multiple enums with a string? `[...DescribeAlias, ...TestCaseName].includes(ancestor.callee.name)` throws `TypeError: _tsUtils.DescribeAlias is not iterable`.

G-Rath: Actually, if you use `in` it'll match `hasOwnProperty` (which is what you should use instead of `in` in this case).

But actually you should use the guard we have: `isTestCase`, as it's a more accurate test :)

garyking:
@G-Rath 	Thanks, I made some changes. It checks if it's within `describe` or `test` blocks now.

G-Rath:
@garyking cheers, but sadly that still isn't quite accurate; check out my second example:

```
// utils.ts
/**
 * Due to the way `jest` mocking works, we have to spy on each function, to have its own implementation.
 *
 * It's weird - just roll with it.
 */
const mockIcons = () => (
  Object.keys(actualIcons) // icons end with "Icon"
        .filter((icon): icon is jest.FunctionPropertyNames<typeof icons> => icon.endsWith('Icon'))
        .forEach(icon => jest.spyOn(icons, icon).mockImplementation(actualIcons[icon]))
);
```

This works completely fine, and isn't in a `describe` or `test` block.

src/rules/prefer-spy-on.ts
```diff
@@ -69,6 +74,19 @@ export default createRule({
...
         if (!jestFnCall) return;
 
+        // The `jest.fn()` must be in a `describe` block (so this includes all
+        // hook blocks) or `test` block (since they can be outside `describe`
+        // blocks) in order for us to report it.
+        const ancestors = context.getAncestors();
+        const isInTestBlock = ancestors.some(
+          ancestor =>
+            ancestor.type === AST_NODE_TYPES.CallExpression &&
+            (ancestor.callee.name in DescribeAlias ||
```
G-Rath: You might has missed my previous comment - sorry for the double up if you did see it before pushing!

`in` will match `hasOwnProperty`, which is why we switched from `in` to use `hasOwnProperty`.

But you can actually just replace this w/ the `isTestCase` & `isDescribe` guards we have in `utils` :)

garyking:
Okay, so the only check we can make is if we're in a file whose path includes `__mocks__` then, I suppose. But that's not possible with ESLint, I don't think?

G-Rath:
Even then, just b/c you use `__mocks__` doesn't mean others do. I for example actually like to have my test & src separate, and so don't use that pattern.

Hence my point that I don't think this should be something the rule should be caring about.
As I said in [my comment on the issue](https://github.com/jest-community/eslint-plugin-jest/issues/209#issuecomment-518385197):

> This plugin is targeted at linting files that are considered "jest tests", and so assumes anything it's given is based within a test suite.
> If we can't assume that, we've have to have some way of figuring out if a file is a test suite or not, which would mean making an arbitrary choice that I'm sure won't work for every possible variation you could have.

`eslint` is more than equipped to handle disabling rules on a per-file/path basis, but it's designed to do so at the config level, rather than at the rule level:

```
overrides: {
  {
    files: ['*/__mocks__/*'],
    rules: {
      'jest/prefer-spy-on': 'off'
    }
  }
}
```

PR #368, opened 2019-08-06
convert dangerfile to typescript

G-Rath:
`chore(dangerfile): convert to typescript`


SimenB:
Running this locally fails.

image

---

@orta maybe try to clean up that error? Seems like a relatively simple syntax error that completely blows up the terminal with stacks and weird state

G-Rath:
Interesting - I had that before w/ `const`, and hoped it was just b/c it wanted `var` for some reason.

This was one of the last files I converted, and it was very late at night, so...

I'm happy to put this on hold until we drop node6 and then revisit it.

SimenB:
SGTM

G-Rath:
Umm actually this failed on all versions of danger...

Maybe it's b/c of `npx`?

orta:
So I think this is two things: 

- `danger local` doesn't have access to GitHub - that's intentional, you're looking for `danger pr` which aims to replicate running on a PR on your computer. It'd be nice for danger to catch that error and raise a message specifically warning you about that.

- Looks like using `npx` doesn't get TypeScript support, this is because TypeScript isn't a dependency of danger (so NPX has no idea how to include it, and so it can't transpile) - so node bails when it sees typescript syntax 

Agree on the stack trace though, and looks like source mapping is off too

SimenB:
We'll ditch the `npx` approach in favour of a dependency in package.json for the next major, which _should_ fix this then, IIUC

G-Rath:
> @SimenB SimenB changed the base branch from master to next 11 minutes ago

How did you manage to do that? I couldn't find that option anywhere..?

SimenB:
image
image

G-Rath:
Welllll I feel silly for having missed that. Cheers!

PR #371, opened 2019-08-06
ts-migration/snapshot-processor

G-Rath:
Figured I might as well break this out

### DIFF
--- a/src/processors/snapshot-processor.ts
+++ b/src/processors/snapshot-processor.ts
@@ -1,0 +1,10 @@
+// https://eslint.org/docs/developer-guide/working-with-plugins#processors-in-plugins
+
+type PostprocessMessage = { ruleId: string };
+
+export const snapshotProcessor = {
+  preprocess: (source: string): string[] => [source],
+  postprocess: (messages: PostprocessMessage[][]) =>
+    // snapshot files should only be linted with snapshot specific rules
+    messages[0].filter(message => message.ruleId === 'jest/no-large-snapshots'),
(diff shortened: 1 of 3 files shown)


src/processors/snapshot-processor.ts
```diff
@@ -0,0 +1,10 @@
+// https://eslint.org/docs/developer-guide/working-with-plugins#processors-in-plugins
+
+type PostprocessMessage = { ruleId: string };
+
+export const snapshotProcessor = {
```
SimenB: `export default`?

G-Rath: I've sort of started to [be against them](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/). Not strongly, but like it's a think I'm trying and have started to kind of liked it and now encourage?

But I'm not married to it, so happy to make it a default export if you'd prefer :)

SimenB: In this case I'd ether go for 2 separate named exports (sa it currently is) or a single `default` one

src/processors/snapshot-processor.ts
```diff
@@ -0,0 +1,10 @@
+// https://eslint.org/docs/developer-guide/working-with-plugins#processors-in-plugins
+
+type PostprocessMessage = { ruleId: string };
```
SimenB: link to https://github.com/typescript-eslint/typescript-eslint/issues/808?

PR #363, opened 2019-08-04
ts-migration/everything-else

G-Rath:
It's finally here!

This PR converts the remaining JS files to TypeScript (where possible) - All tests are passing, and coverage is 100%; as such this PR is good to go as is.

For now I'm going to do some light cleanup, which I'll commit & push to this branch, but can also be in the next PR - let me know if you'd like me to cherry pick anything out into separate PRs & what have you.

*I am trying to avoid changing the already-converted rules in this PR*, which is why there is definitely some redundancy, and forms a nice bridge to the second half of what is quickly becoming one of my classic overly detailed text dumps:

My immediate next step is going over every rule w/ a fine-tooth comb to make sure they're all consistent, and to fix a number of nitpicks, including:
  * property order (i.e `valid` above `invalid` in rule tests)
  * using the new `parseExpectCall` where possible, & other such new methods
  * ensure the same property name is used in `data` (i.e `name` vs `propertyName` vs `matcherName`)
  * checking all the meta for each rule is correct (recommended, description, etc)
  * merging redundant types & code
  * use template strings instead of `.join('\n')` for multi-line strings

The big massive inconsistency in the room that I'd like to get addressed is our stance on bracket accessors:

```
expect(files['length']).toBe(1);
expect(/* args */)['not'].toBe(/* args */);
describe['skip'](/* args */);
```

A few tests have explicit test cases showing support (i.e `describe['skip']`), but most other rules either won't work w/ that type of accessor, or - in the case of fixers - will error.

The former is not a problem, as the `AccessorNode` & co (`getAccessorValue`, `isSpecificAccessor`, etc) types & methods are for facilitating the normalisation required for rules to support both `Identifier` (`expect.<blah>`) & literal node types (`expect['blah']`).

However, all the fixers currently will just error out - this is b/c the ones that this matter for are typically searching for a `propertyDot` value.

It's fairly straightforward to implement support, and am happy to do it - but it's something that other maintainers should be aware of for when reviewing new rules, as we should try and ensure all rules have tests for both types of accessor.

In light of that, I'm going look to create a series of standard tests that all rules should have to be tested against, w/ things like `expect();`, `expect`, `expect.toBe()`, etc to make sure that all our rules can handle the same basic edge cases.

I also want to review the stuff for test, describe, etc, & `getNodeName` as they're somewhat duck-taped together. Ideally it'd be cool if I could create a "parse" method for those too, but that might be a bit overkill - either way, there are some open issues relating to them so let's see what I can come up w/ :D

[...]

PR #372, opened 2019-08-06
ts-migration/index

G-Rath:
chore(index): convert to typescript.

Might as well break this out too :joy:

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 import { readdirSync } from 'fs';
-import { basename, join } from 'path';
+import { join, parse } from 'path';
 import globals from './globals.json';
 import * as snapshotProcessor from './processors/snapshot-processor';
 
@@ -22,9 +22,7 @@ const rules = readdirSync(rulesDir)
       rule !== 'tsUtils.ts' &&
       rule !== 'tsUtils.js',
   )
-  .map(rule =>
-    rule.endsWith('js') ? basename(rule, '.js') : basename(rule, '.ts'),
-  )
+  .map(rule => parse(rule).name)
   .reduce(
     (acc, curr) =>
       Object.assign(acc, { [curr]: importDefault(join(rulesDir, curr)) }),
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,15 +5,16 @@ import * as snapshotProcessor from './processors/snapshot-processor';
 
 // copied from https://github.com/babel/babel/blob/d8da63c929f2d28c401571e2a43166678c555bc4/packages/babel-helpers/src/helpers.js#L602-L606
 /* istanbul ignore next */
-function interopRequireDefault(obj) {
-  return obj && obj.__esModule ? obj : { default: obj };
-}
+const interopRequireDefault = (obj: any): { default: any } =>
+  obj && obj.__esModule ? obj : { default: obj };
 
-function importDefault(moduleName) {
-  return interopRequireDefault(require(moduleName)).default;
-}
+const importDefault = (moduleName: string) =>
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  interopRequireDefault(require(moduleName)).default;
 
-const rules = readdirSync(join(__dirname, 'rules'))
+const rulesDir = join(__dirname, 'rules');
+
+const rules = readdirSync(rulesDir)
   .filter(
     rule =>
       rule !== '__tests__' &&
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -26,16 +27,13 @@ const rules = readdirSync(join(__dirname, 'rules'))
...
-let allRules = {};
-Object.keys(rules).forEach(function(key) {
-  allRules[`jest/${key}`] = 'error';
-});
+const allRules: Record<string, string> = {};
+Object.keys(rules).forEach(key => (allRules[`jest/${key}`] = 'error'));
 
-// eslint-disable-next-line import/no-commonjs
-module.exports = {
+export default {
```
SimenB: as mentioned in the other review, I don't think this can be `export default` as eslint just does `require('plugin')` and expect that object to have `rules` etc. I don't think it deals with `.default` etc

G-Rath: Maybe, but the `lib` output looks the same:

`exports.default`:
```
Object.keys(rules).forEach(key => allRules[`jest/${key}`] = 'error');
exports.default = {
  configs: {
    all: {
  ....
```

`export default`:
```
Object.keys(rules).forEach(key => allRules[`jest/${key}`] = 'error');
var _default = {
  configs: {
   .... 
}
exports.default = _default;
```

SimenB: it was `module.exports`, not `exports.default `previously

G-Rath: No idea what you mean - totally didn't blunder it and use `exports.default` by mistake ;P 

~But yeah still the same output as before:~

Yeah you're right. I'll revert that :joy:

SimenB: 😀

SimenB:
This fails locally due to #371 not being merged. Can land after that's done 🙂

G-Rath:
haha yeah I saw that - I'm doing that right now.

[btw do you want `parse` in this](https://github.com/jest-community/eslint-plugin-jest/pull/363#discussion_r310985205)?

SimenB:
Ah yeah, `parse` looks like a nice improvement over `basename`

G-Rath:
> `.filter(rule => !excludedFiles.includes(rule))`

Woops... 😅

I was about to amend the "bad" commit message, but now that you've pushed, I'll leave it to you if that's ok.

Once this has been merged would you mind rebasing #363 for me? I did it locally, but something weird happened and it decided to not apply the change to `snapshot-processor.test.ts` so it was using the old `const { snapshotProcessor }` import 😬

I can do it if you're not got the time :)

src/index.ts
```diff
@@ -25,11 +25,7 @@ const rules = readdirSync(rulesDir)
   );
 
 const allRules = Object.keys(rules).reduce<Record<string, string>>(
-  (rules, key) => {
-    rules[`jest/${key}`] = 'error';
-
-    return rules;
-  },
+  (rules, key) => ({ ...rules, [`jest/${key}`]: 'error' }),
```
G-Rath: This is actually surprisingly "slow", b/c everytime you're creating a new object & copying all the keys.

It's by far not a big deal, and even less so since we're doing it the once (I hope? I assume the way eslint runs it typically reuses the same node process), but it's an interesting little fact.

(That might have changed in later versions of node - iirc I last tested this on node8)

SimenB: doing it once during `require`, yeah - probably not even measurable except in micro benchmarks. _Nothing_ compared to the FS operations in the same file 🙂

SimenB:
Rebased #363 now. I _think_ I did it 😛

src/index.ts
SimenB: and yeah, object spread has gotten way quicker (although it's currently transpiled down by babel until we drop node 6)

https://twitter.com/bmeurer/status/1015105293301280768?lang=en

G-Rath:
@SimenB 

> SyntaxError: /c/Users/G-Rath/workspace/projects-personal/eslint-plugin-jest/src/index.ts: `export =` is not supported by @babel/plugin-transform-typescript
> Please consider using `export <value>;`.
>  31 |
>  32 | // eslint-disable-next-line import/no-commonjs
>\> 33 | export = {

That's the syntax for `modules.export` 😭

SimenB:
I wrote a custom babel plugin for it in jest: https://github.com/facebook/jest/blob/master/scripts/babel-plugin-jest-replace-ts-export-assignment.js

G-Rath:
oohhhohhh exciting!

My work does [investment time](https://thoughtbot.com/blog/investment-time) every Friday (which is tomorrow), so I'll explore that plugin, and look about adding it at some point :D

SimenB:
Feel free to rip it out, publish it, and send a PR to jest migrating to using the published version 🙂

G-Rath:
Sure thing :D

PR #373, opened 2019-08-09
Use export assignment

G-Rath:
@SimenB I probably spent more time than I should, but hey least I know a lot more about babel :D

I'll make a PR into `jest` soon, once I've convinced my laptop to play nice (and once I've figured out how to install the package...).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -26,7 +26,7 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.4.4":
+"@babel/core@^7.1.0", "@babel/core@^7.4.4", "@babel/core@^7.5.5":
   version "7.5.5"
   resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.5.5.tgz#17b2686ef0d6bc58f963dddd68ab669755582c30"
   integrity sha512-i4qoSr2KTtce0DmkuuQBV4AuQgGPUcPXMr9L5MyYAtk06z068lQ10a4O009fe5OB/DfNV+h+qqT7ddNV8UnRjg==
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -1421,6 +1421,15 @@ babel-plugin-jest-hoist@^24.6.0:
   dependencies:
     "@types/babel__traverse" "^7.0.6"
 
+babel-plugin-replace-ts-export-assignment@^0.0.1:
+  version "0.0.1"
+  resolved "https://registry.yarnpkg.com/babel-plugin-replace-ts-export-assignment/-/babel-plugin-replace-ts-export-assignment-0.0.1.tgz#78305d7e15fe1cbfff7bbbfdf7f0cf92f21a15d6"
+  integrity sha512-YBhWpHLAveoXQSDiiU0lk1pPDHKeBpDTcwjUTdzFOTe40kAGxcSM+XvD7Vp6SP9D79eBKNHgG5Ji3cigPjAcfw==
+  dependencies:
+    "@babel/core" "^7.5.5"
```
SimenB: `@babel/core` should be a peer dep, not a dep (if you need it).

---

Why do you need any deps at all, btw?

G-Rath: Yeah that was something I debated over, and was a little unsure of, as they're needed for the types.

But looking over it you're completely right - for some reason I thought I imported one thing from `babel` that meant I need a dep instead of a dev (i.e something like AST_TREE_NODE, where it was just referencing a constant) but no I don't.

I'll make them dev deps & publish a new version :)

G-Rath: I've released a new version w/o any dependencies :)

G-Rath:
@SimenB would you be able to recommend a good look to learn how to do mono-repos to make a PR for this into `jest`? 

I've gotten the repo forked & setup locally, w/ all but two tests passing, but I've never worked w/ a mono-repo before, & jest is a big one 😬 

I'm keen to make the PR myself, as I'm wanting to take a crack at [8330](https://github.com/facebook/jest/issues/8330) after I've finished the TS conversation here (as a bit of a "break" before jumping into the open issues here).

SimenB:
A monorepo isn't anything super special, and contributing.md should have the details on how to work with the code base. I don't know of any guides or blog posts, so I guess my googling would be as good as yours 🙂 Any particular questions?

>w/ all but two tests passing

If those are for mercurial (`hg`), don't worry about them. They fail for me as well. I _think_ you need some specific version of mercurial installed, but I haven't bothered to look into it

G-Rath:
> A monorepo isn't anything super special, and contributing.md should have the details on how to work with the code base

Yeah, I've got it all setup & working, but the one thing contributing.md doesn't covert is adding new packages:

> g-rath@CINDY:/c/Users/G-Rath/workspace/projects-oss/jest$ yarn add -D babel-plugin-replace-ts-export-assignment
> yarn add v1.17.3
> error Running this command will add the dependency to the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.

I saw that and decided to bail for now, in favor of reading up on how dependencies are handled in monorepos.

So that's pretty much my main question: how do I add new dependencies?

> If those are for mercurial (hg), don't worry about them

They were, which is exactly why I ignored them 😄 I figured if they were really failing outside my environment, someone would already be handling it.

SimenB:
Aha! Yeah, just do what the error suggests, add the `-W` flag 🙂

G-Rath:
Cheers, worked as expected - I'm now running the 8min+ test suite 😬 

I'll carry on working on this stuff while pondering if I should have called it `babel-plugin-typescript-replace-export-assignment` 😕 

----

Just to complete my knowledge: Since packages don't have their own lock files, am I correct in thinking that adding a dep in a "package" works by the normal `npm i`/`yarn add`, which `lerna` intercepts & redirects/applies/magically-teleports into the root lock file?

SimenB:
You must use `yarn`, but yeah, it's added to the root lockfile. This is because we use yarn workspaces: https://yarnpkg.com/lang/en/docs/workspaces/. Jest uses lerna for publishing, not for dependency management 🙂

PR #374, opened 2019-08-10
ts-migration/improve-no-empty-title

G-Rath:
I was converting `no-large-snapshots`, as that pulls in all the `getAccessorValue` & related stuff w/o needing the expect parsing, when I realised that `no-empty-title` can be easily simplified :D


G-Rath:
tut tut - silly "Invalid revision range"; you're just so silly XD

PR #375, opened 2019-08-10
Improve no identical title

G-Rath:
This is based off #374, which should be reviewed & merged first.

I had to simplify `getAccessorValue` & `AccessorNode` to not handle `Identifier` nodes, as nothing covers that yet.

Otherwise it's a direct port.

### DIFF
--- a/src/rules/tsUtils.ts
+++ b/src/rules/tsUtils.ts
@@ -109,41 +109,6 @@ export const getStringValue = <S extends string>(node: StringNode<S>): S =>
   isTemplateLiteral(node) ? node.quasis[0].value.raw : node.value;
 
 /**
- * Represents a `MemberExpression` with a "known" `property`.
- */
-interface KnownMemberExpression<Name extends string = string>
-  extends TSESTree.MemberExpression {
-  property: AccessorNode<Name>;
-}
-
-/**
- * Represents a `CallExpression` with a "known" `property` accessor.
- *
- * i.e `KnownCallExpression<'includes'>` represents `.includes()`.
- */
-export interface KnownCallExpression<Name extends string = string>
-  extends TSESTree.CallExpression {
-  callee: CalledKnownMemberExpression<Name>;
-}
-
-/**
- * Represents a `MemberExpression` with a "known" `property`, that is called.
- *
- * This is `KnownCallExpression` from the perspective of the `MemberExpression` node.
- */
-export interface CalledKnownMemberExpression<Name extends string = string>
-  extends KnownMemberExpression<Name> {
-  parent: KnownCallExpression<Name>;
-}
-
-/**
- * An `Identifier` with a known `name` value - i.e `expect`.
- */
-interface KnownIdentifier<Name extends string> extends TSESTree.Identifier {
-  name: Name;
-}
-
-/**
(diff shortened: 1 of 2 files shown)


src/rules/tsUtils.ts
```diff
@@ -108,6 +108,57 @@ export const isStringNode = <V extends string>(
...
+  extends TSESTree.MemberExpression {
+  property: AccessorNode<Name>;
+}
+
+/**
+ * Represents a `CallExpression` with a "known" `property` accessor.
+ *
+ * i.e `KnownCallExpression<'includes'>` represents `.includes()`.
+ */
+export interface KnownCallExpression<Name extends string = string>
```
SimenB: no need to export, but I guess we can clean that up later?

SimenB: doesn't seem to be used either?

G-Rath: Yeah, exactly - it does need to be exported later for at least one rule.

I don't think it's worth worrying about exports too much right now.
Same with possible use of varargs, as you're about to see in my next PR for `no-large-snapshots` :P

G-Rath: They're circular: `KnownCallExpression` is used by `CalledKnownMemberExpression`, which is used by `KnownCallExpression`, and extends `KnownMemberExpression`.

You're right that they can be removed.

SimenB: `KnownCallExpression` is only used in `CalledKnownMemberExpression`, which is only used in `KnownCallExpression`. It seems we only need `getAccessorValue` here? Test coverage won't complain about unused types as they're stripped out (is there some tsc option we can add that yells about it?)

SimenB: hah, right. GH didn't feel like telling me you had commented...

G-Rath: > Test coverage won't complain about unused types as they're stripped out (is there some tsc option we can add that yells about it?)

Yeah I've been wondering that myself - as far as I know tsc doesn't have such an option, but I think it'd be a pretty big can of worms.

I'm happy to investigate - IntelliJ has an inspection that'll pick them up (which they've now exposed via API, so technically you could use that in CI.... 😜)

I've just checked `@typescript-eslint`, and they don't have a rule for it.

Right now I think it'd be an annoying can of worms - I'm already running into trouble w/ the fact that an `AccessorNode` can be an `Identifier`, but no rule calls `getAccessorValue` on an `Identifier` (it'll make sense when you see the PR in a sec).

G-Rath: ugh yeah it seems to be a slow day in the tech world - I'm trying to push a new commit that removes those interfaces, but it's being sllooooww (might have something to do w/ having the jest monorepo open in another project, but pfft)

G-Rath:
@SimenB Can I merge, since you've approved this?

SimenB:
Mind resolving the conflict?

G-Rath:
Sweet as. I'm about to make a couple of similar PRs that'll likely conflict, and then head to bed: so if you approve them, I'm happy to resolve conflicts & merge them when I'm up tomorrow :)

brb - someone over on jest wants someone who knows TS to eye up his PR 😉

PR #376, opened 2019-08-10
ts-migration/no-large-snapshots

G-Rath:
Based off #375

### DIFF
--- a/src/rules/no-large-snapshots.ts
+++ b/src/rules/no-large-snapshots.ts
@@ -1,20 +1,35 @@
+import {
+  AST_NODE_TYPES,
+  TSESLint,
+  TSESTree,
+} from '@typescript-eslint/experimental-utils';
 import { isAbsolute } from 'path';
-import { getDocsUrl, getStringValue } from './util';
+import {
+  createRule,
+  getAccessorValue,
+  isExpectMember,
+  isSupportedAccessor,
+} from './tsUtils';
 
-const reportOnViolation = (context, node) => {
-  const lineLimit =
-    context.options[0] && Number.isFinite(context.options[0].maxSize)
-      ? context.options[0].maxSize
-      : 50;
+interface RuleOptions {
+  maxSize?: number;
+  whitelistedSnapshots?: Record<string, Array<string | RegExp>>;
+}
+
+type MessageId = 'noSnapshot' | 'tooLongSnapshots';
+
+type RuleContext = TSESLint.RuleContext<MessageId, [RuleOptions]>;
+
+const reportOnViolation = (
+  context: RuleContext,
+  node: TSESTree.CallExpression | TSESTree.ExpressionStatement,
+  { maxSize: lineLimit = 50, whitelistedSnapshots = {} }: RuleOptions,
+) => {
   const startLine = node.loc.start.line;
   const endLine = node.loc.end.line;
   const lineCount = endLine - startLine;
-  const whitelistedSnapshots =
-    context.options &&
-    context.options[0] &&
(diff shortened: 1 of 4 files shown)


src/rules/no-large-snapshots.ts
```diff
@@ -78,28 +103,34 @@ export default {
...
       };
     } else if (context.getFilename().endsWith('.js')) {
       return {
         CallExpression(node) {
-          const propertyName =
-            node.callee.property && node.callee.property.name;
           if (
-            propertyName === 'toMatchInlineSnapshot' ||
-            propertyName === 'toThrowErrorMatchingInlineSnapshot'
+            'property' in node.callee &&
```
G-Rath: @SimenB I know what you're going to say - I'm going to implement varargs in it's own PR, as it's a waterfall change that while relatively straightforward, does require a bit of its own work due to subtyping.

The main reason I replaced the array here is this way `isSupportedAccessor` has full coverage; otherwise I'd have left it as `includes` & done the varargs thing in another PR.

src/rules/tsUtils.ts
```diff
@@ -108,6 +108,128 @@ export const isStringNode = <V extends string>(
...
+ * @param {AccessorNode<S>} accessor
+ *
+ * @return {S}
+ *
+ * @template S
+ */
+export const getAccessorValue = <S extends string = string>(
+  accessor: AccessorNode<S>,
+): S =>
+  /* istanbul ignore next */
```
G-Rath: This is b/c we have no rules that use this on an `Identifier` so coverage complains, *but* `no-large-snapshots` uses `isSupportedAccessor` on an `Identifier`, meaning `AccessorNode` needs to be a union, and so TypeScript requires us to handle both branches.

It's a catch-22 :)

G-Rath:
FYI I'm in the middle of rebasing this one :)

src/rules/tsUtils.ts
SimenB: you solved this another way? Seems to be gone

SimenB:
>FYI I'm in the middle of rebasing this one :)

Heh, new conflict 😅

G-Rath:
>Heh, new conflict 😅

 hehe yeah I'm going to finish the second round of rebasing, and then go to bed, cause I've got work in 5 hours.

I'll look to respond to your other review stuff "tomorrow", but in general from what I've read yes to everything :P

src/rules/tsUtils.ts
G-Rath: I solved it by using it - this was never needed in this branch, but slipped in via the rebase

G-Rath:
Provided the tests push, I'm going to push up the rebased version.

@SimenB if you've got the time, feel free to rebase the commits to make future rebasing easier while I'm sleeping.

Right now it's actually pretty horrible, I think in part b/c there's a number of now redundant commits that are very steppy i.e rename method, move method, remove method -> end result is method that now already exists in master.

Otherwise, I can try and remove them via rebase when I get to work.

Cheers

SimenB:
I can rebase #363 after landing this and see if I can clear out a few of the "chore" commits

src/rules/tsUtils.ts
G-Rath: Sorry thought this was #363 - my tests locally still require this for this PR?

G-Rath:
@SimenB should be good to go - I've id'd a couple of changes that can be rebased away in #363 so hopefully that helps.

src/rules/tsUtils.ts
SimenB: this line doesn't exist anymore

PR #382, opened 2019-08-11
ts-migration/valid-expect-in-promise

G-Rath:
This'll probably conflict a little when #376 is merged, but should be fine.


G-Rath:
Ugh WebStorm gets confused when you switch from js -> ts, b/c it marks the `js` as a generated file, so it told me `getStringValue` wasn't being used, when its needed by `no-large-snapshots` :joy:

src/rules/__tests__/valid-expect-in-promise.test.ts
```diff
@@ -325,12 +196,128 @@ ruleTester.run('valid-expect-in-promise', rule, {
...
     `,
-
     `
       it('name of done param does not matter', (nameDoesNotMatter) => {
         const promise = getPromise();
         promise.then(() => expect(someThing).toEqual(true));
       });
     `,
   ],
+  invalid: [
```
SimenB: these are just moved right? no actual change?

G-Rath: Yup - every other test has `valid` at the top, & I think the consistency is nice.

G-Rath:
Want me to handle the conflict?

SimenB:
yes please 🙂

SimenB:
This is great! (too bad git gets so confused by renames :( reviewing locally works way better)

PR #383, opened 2019-08-12
ts-migration/no-alias-methods

G-Rath:
`/* simenb ignore next it's that or radically restructure the types just to revert them in the next PR*/`

### DIFF
--- a/src/rules/tsUtils.ts
+++ b/src/rules/tsUtils.ts
@@ -336,14 +336,6 @@ export interface ParsedExpectMatcher<
   Node extends ExpectMember<Matcher> = ExpectMember<Matcher>
 > extends ParsedExpectMember<Matcher, Node> {
   arguments: TSESTree.CallExpression['arguments'] | null;
-  // isNamed<PossibleName extends Matcher>(
-  //   name: PossibleName,
-  //   ...or: PossibleName[]
-  // ): this is ParsedExpectMatcher<PossibleName>;
-  // isNamed<PossibleName extends Matcher>(
-  //   // name: PossibleName,
-  //   ...names: [PossibleName, ...PossibleName[]]
-  // ): this is ParsedExpectMatcher<PossibleName, Node>;
 }
 
 type BaseParsedModifier<
@@ -403,9 +395,6 @@ const reparseAsMatcher = (
     parsedMember.node.parent.type === AST_NODE_TYPES.CallExpression
       ? parsedMember.node.parent.arguments
       : null,
-  // isNamed(...names: string[]) {
-  //   return names.includes(this.name);
-  // },
 });
 
 /**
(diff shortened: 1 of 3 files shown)


src/rules/tsUtils.ts
```diff
@@ -320,6 +320,185 @@ export const isExpectCallWithParent = (
...
+  }
+
+  /* istanbul ignore if */
+  if (
+    !isSpecificMember(parsedMember, ModifierName.resolves) &&
+    !isSpecificMember(parsedMember, ModifierName.rejects)
+  ) {
+    throw new Error(
+      `modifier name must be either "${ModifierName.resolves}" or "${ModifierName.rejects}" (got "${parsedMember.name}")`,
+    ); // ts doesn't think that the ModifierName.not check is the direct inverse as the above two checks
```
SimenB: can you move these comments above the `throw` rather than inline on the closing braces?

src/rules/tsUtils.ts
```diff
@@ -320,6 +320,185 @@ export const isExpectCallWithParent = (
...
+
+    return expectation;
+  }
+
+  const modifier = (expectation.modifier = reparseMemberAsModifier(
+    parsedMember,
+  ));
+
+  const memberNode =
+    ('negation' in modifier && modifier.negation) || modifier.node;
```
SimenB:
```suggestion
    modifier.negation || modifier.node;
```

src/rules/tsUtils.ts
```diff
@@ -320,6 +320,185 @@ export const isExpectCallWithParent = (
...
+
+/**
+ * Represents a parsed expect matcher, such as `toBe`, `toContain`, and so on.
+ */
+export interface ParsedExpectMatcher<
+  Matcher extends MatcherName = MatcherName,
+  Node extends ExpectMember<Matcher> = ExpectMember<Matcher>
+> extends ParsedExpectMember<Matcher, Node> {
+  arguments: TSESTree.CallExpression['arguments'] | null;
+  // isNamed<PossibleName extends Matcher>(
```
SimenB: what are these comments for?

src/rules/tsUtils.ts
```diff
@@ -320,6 +320,185 @@ export const isExpectCallWithParent = (
...
+   *
+   * If this matcher isn't called, this will be `null`.
+   */
+  arguments:
+    /* istanbul ignore next */
+    parsedMember.node.parent &&
+    parsedMember.node.parent.type === AST_NODE_TYPES.CallExpression
+      ? parsedMember.node.parent.arguments
+      : null,
+  // isNamed(...names: string[]) {
```
SimenB: ?

src/rules/tsUtils.ts
G-Rath: They're something I was toying around w/ that seem to work, but I didn't want to switch to until I'd gotten everything else landed.

I wanted them committed so that they were in the history for when I wanted to play around w/ them again.

They can go now probably.

(the idea is that you can do `matcher.isNamed('toBe')`)

src/rules/tsUtils.ts
G-Rath: Good catch - that's not required anymore as it's become type safe :D

PR #386, opened 2019-08-12
ts-migration/require-tothrow-message

G-Rath:
chore(require-tothrow-message): convert to typescript


SimenB:
Test failures

src/rules/__tests__/require-tothrow-message.test.js
```diff
@@ -1,112 +0,0 @@
-import { RuleTester } from 'eslint';
```
SimenB: Why  isn't this a rename?

src/rules/require-tothrow-message.ts
```diff
@@ -0,0 +1,50 @@
...
+      CallExpression(node) {
+        if (!isExpectCall(node)) {
+          return;
+        }
+
+        const { matcher, modifier } = parseExpectCall(node);
+
+        if (
+          matcher &&
+          matcher.arguments &&
```
SimenB: These first two checks should imo be unnecessary after the `isExpectCall` guard

G-Rath:
ugh crud that's b/c of indenting

src/rules/require-tothrow-message.ts
```diff
@@ -0,0 +1,50 @@
...
+          matcher.arguments.length === 0 &&
+          ['toThrow', 'toThrowError'].includes(matcher.name) &&
+          (!modifier ||
+            !(modifier.name === ModifierName.not || modifier.negation))
```
SimenB: Why is both `not` and `negation` check needed?

src/rules/__tests__/require-tothrow-message.test.js
G-Rath: You'll have to ask git that 🤷‍♂

src/rules/require-tothrow-message.ts
G-Rath: By what logic?

An expect call is `expect()` - that's it. It has never had a requirement for there to be a matcher, not for that matcher to be called.

src/rules/require-tothrow-message.ts
G-Rath: They're not, for the same reason `'negation' in matcher` isn't.

src/rules/__tests__/require-tothrow-message.test.js
G-Rath: > If indentation changed, please revert it.

Sure, but I honestly have no idea what it is:
```
import { RuleTester } from 'eslint';
// vs
```

Maybe line ends? I'll check it out

src/rules/require-tothrow-message.ts
SimenB: Then `isExpectCall` should change (although we might wanna do that in the future)). The only rule that should ever have to care "was there a matcher?" should be `valid-expect` - _all_ other rules should just bail from some sort of type guard earlier if there's no invoked matcher on the `expect` call

G-Rath: Sure, that's fair, and why I originally had `isValidExpect` call.

I pegged it to discuss later, since while you're technically, it'll mean that simply having an extra parameter means you only get told about `valid-expect`, which can be annoying - if you make a typo, you have to run the linter twice to know about all the flaws.

Do you want me to update all the rules in this PR, or afterwards?

If I do it at the end, it's a simple PR: I just restore `isValidExpectCall`, and "rename" `isExpectCall` to that method name.

SimenB: > it'll mean that simply having an extra parameter means you only get told about `valid-expect`

argument where? if it's `expect(one, two)` we need to make that configurable so we don't break https://www.npmjs.com/package/jest-expect-message

> if you make a typo, you have to run the linter twice to know about all the flaws.

You should get that feedback in your editor so I don't think that's something to worry about.

>Do you want me to update all the rules in this PR, or afterwards?

A separate PR sounds ideal 🙂

G-Rath: > argument where? if it's expect(one, two) we need to make that configurable so we don't break https://www.npmjs.com/package/jest-expect-message

It's not configurable: 
> multipleArgs: 'More than one argument was passed to expect().',

> You should get that feedback in your editor so I don't think that's something to worry about.

If I'm using typescript.

My point is, say you push to CI, w/ something like:

```
expect(Promise.reject(2)).not.resolves.toEqual(3)
                           \ -- valid-expect fail!
```

So you say "oh drat", fix it, and push it back up:

```
expect(Promise.reject(2)).not.resolves.toEqual(3)
                                          \ -- prefer-strict-equal fail!
```

You've just had two CI runs to find that out.

I personally don't mind either way, b/c I use TypeScript - so yes, I can't have an invalid expect.

However, while you can just say "oh well your editor should be running the linter anyway", which I'd say to, I work at a company that recently added hooks to their projects b/c this sort of stuff was slowing them down as their CI would run for 10 minutes before hitting a linting error and reject the build.

I'm just saying that there are rules that can rule w/ "invalid" expects, as the expect is not the focus.

> A separate PR sounds ideal 🙂

Awesome 😄

src/rules/require-tothrow-message.ts
G-Rath: >   ● require-tothrow-message › valid › expect(() => { throw new Error('a'); }).not.toThrowError();

That's why the `not` & negation is required.

`negation` is for `resolves` & `rejects`.

src/rules/require-tothrow-message.ts
SimenB: >It's not configurable

`valid-expect` should be configurable here to allow it, but we can fix that later. 

>You've just had two CI runs to find that out.

That's true for unit tests as well. We should optimize for the people who run stuff locally and don't count on CI to fix things they should fix before pushing. IDE integration, precommit hooks etc..

>I'm just saying that there are rules that can rule w/ "invalid" expects, as the expect is not the focus.

I totally agree.

It would be nice if syntactically valid `expect`s that `valid-expect` reports on still went through the `isValidExpect` check (or whatever it's called), as most rules almost always only care about the matcher. And if `valid-expect` is _not_ used, the other rules should still warn. So sorta what you wanted up top, but I think we can find a middle ground between the two options.

I'd prefer `isValidExpect` to return true if it's a _syntactically_ valid `expect`, even if it's not _semantically_ valid (which means `expect(something, maybeMoreThings).someFunction(with, orWithoutArgs)` should be counted as an `expect`)

G-Rath: > valid-expect should be configurable here to allow it, but we can fix that later.

Tracking via #391 🙂

----

All-in-all I'm definitely keen to have a chat about this, b/c it's somewhat dependent on what each rule wants, and would drastically change the current typings & introduce a bit of duplication depending on what we went w/.

RIght now the parser was done w/ the speed of existence checks in mind - while it's an extra line of code, pound-for-pound the cost of it is "almost 0", especially given that V8 optimises a lot of that sort of stuff.

> I'd prefer isValidExpect to return true if it's a syntactically valid expect, even if it's not semantically valid (which means expect(something, maybeMoreThings).someFunction(with, orWithoutArgs) should be counted as an expect) 

That sounds like what I had w/ `isExpect` & `isValidExpect` - and is typable as well.

The only question I have w/ it is what do you mean by `(with, orWithoutArgs)`, b/c remember that the `matcher` is not the same as the "call".

i.e the `()` are a *CallExpression*, while `someFunction` is a `MemberExpression` (of which the `CallExpression` is it's parent).

Overall, everything I'm hearing is sounds like a positive discussion 😄

SimenB: for everything but `valid-expect`, a matcher should always be a callexpression imo, otherwise it's not interesting as it shouldn't be considered a matcher. Just a random member

PR #333, opened 2019-07-22
WIP: ts-migration/valid-expect

yatki:

PR #387, opened 2019-08-12
ts-migration/valid-expect

G-Rath:
chore(valid-expect): convert to typescript

Closes #333


src/rules/__tests__/valid-expect.test.ts
```diff
@@ -15,23 +15,65 @@ ruleTester.run('valid-expect', rule, {
...
         'test("valid-expect", () => expect(Promise.resolve(2)).resolves.toBeDefined());',
       options: [{ alwaysAwait: true }],
     },
     'test("valid-expect", () => expect(Promise.resolve(2)).resolves.toBeDefined());',
     'test("valid-expect", () => expect(Promise.reject(2)).rejects.toBeDefined());',
-    'test("valid-expect", () => expect(Promise.reject(2)).not.resolves.toBeDefined());',
-    'test("valid-expect", () => expect(Promise.reject(2)).not.rejects.toBeDefined());',
     'test("valid-expect", () => expect(Promise.reject(2)).resolves.not.toBeDefined());',
     'test("valid-expect", () => expect(Promise.reject(2)).rejects.not.toBeDefined());',
+    // 'test("valid-expect", () => expect(Promise.reject(2)).not.resolves.toBeDefined());',
```
SimenB: ?

src/rules/__tests__/valid-expect.test.ts
```diff
@@ -15,23 +15,65 @@ ruleTester.run('valid-expect', rule, {
     'expect(undefined).not.toBeDefined();',
     'test("valid-expect", () => { return expect(Promise.resolve(2)).resolves.toBeDefined(); });',
     'test("valid-expect", () => { return expect(Promise.reject(2)).rejects.toBeDefined(); });',
-    'test("valid-expect", () => { return expect(Promise.resolve(2)).not.resolves.toBeDefined(); });',
-    'test("valid-expect", () => { return expect(Promise.resolve(2)).not.rejects.toBeDefined(); });',
-    'test("valid-expect", function () { return expect(Promise.resolve(2)).not.resolves.toBeDefined(); });',
-    'test("valid-expect", function () { return expect(Promise.resolve(2)).not.rejects.toBeDefined(); });',
-    'test("valid-expect", function () { return Promise.resolve(expect(Promise.resolve(2)).not.resolves.toBeDefined()); });',
-    'test("valid-expect", function () { return Promise.resolve(expect(Promise.resolve(2)).not.rejects.toBeDefined()); });',
+    'test("valid-expect", () => { return expect(Promise.resolve(2)).resolves.not.toBeDefined(); });',
```
SimenB: would be cool to error on these _not_ being valid. can do that later, though

G-Rath: Agreed

src/rules/__tests__/valid-expect.test.ts
G-Rath: I left there here commented out b/c the whole thing was a mess, and I didn't know what the & wasn't valid :)

Case & point, up until now we were testing the the exact opposite of the correct (`not.resolve`) was correct.

Feel free to remove them if you want, but I'll def be circling back, so they won't be there for ever :)

SimenB:
This turned out pretty clean, I love it! 😀

G-Rath:
Just wait until we've got all the rules converted, and I do my second pass over everything ;)

PR #388, opened 2019-08-12
improve-prefer-strict-equal

G-Rath:
chore(prefer-strict-equal): use `parseExpectCall`


SimenB:
Lowers test coverage - can we delete some stuff from `tsUtils`?

Issue #389, opened 2019-08-12
prefer-spy-on fix breaks/changes tests


chrisblossom:
Currently the autofixer for `jest/prefer-spy-on` breaks/changes the behavior of the test by re-implementing the original behavior of the function (which may or may not be wanted).

For example, `Date.now = jest.fn()` becomes `jest.spyOn(Date, 'now')`, but it should be `jest.spyOn(Date, 'now').mockImplementation()`. If a function is provided to `jest.fn()` it works as expected.

PR #390, opened 2019-08-12
fix(prefer-spy-on): do not change behavior of fixed instances

chrisblossom:
If this is expected behavior, I'd suggest removing the autofix when `mockImplementation` is empty because it can (and did in my case) break people's code.

Fixes: #389


SimenB:
When this releases, it'll include a bunch of TS migration stuff. 🤞 nothing breaks as I'm heading to bed 😛

PR #393, opened 2019-08-13
ts-migration/prefer-to-contain

G-Rath:
This brings over the majority of what's left for `tsUtils`.


src/rules/prefer-to-contain.ts
```diff
@@ -0,0 +1,248 @@
...
+  const [containArg] = node.arguments;
+  const negationPropertyDot = findPropertyDotToken(modifier.node, sourceCode);
+
+  const toContainFunc = buildToContainFuncExpectation(
+    matcher.arguments[0].value,
+  );
+  // ? 'not.toContain'
+  // : 'toContain';
+
+  /* istanbul ignore next */
```
SimenB:
```suggestion
  /* istanbul ignore if */
```

src/rules/prefer-to-contain.ts
```diff
@@ -0,0 +1,248 @@
...
+) => {
+  const [containArg] = node.arguments;
+  const includesCallee = node.callee;
+
+  const propertyDot = findPropertyDotToken(includesCallee, sourceCode);
+
+  const closingParenthesis = sourceCode.getTokenAfter(containArg);
+  const openParenthesis = sourceCode.getTokenBefore(containArg);
+
+  /* istanbul ignore next */
```
SimenB:
```suggestion
  /* istanbul ignore if */
```

src/rules/prefer-to-contain.ts
```diff
@@ -0,0 +1,248 @@
...
+    fixer.remove(modifier.node.property),
+    fixer.replaceText(matcher.node.property, toContainFunc),
+    fixer.replaceText(matcher.arguments[0], sourceCode.getText(containArg)),
+  ];
+};
+
+const getCommonFixes = (
+  node: FixableIncludesCallExpression,
+  sourceCode: TSESLint.SourceCode,
+  fixer: TSESLint.RuleFixer,
```
SimenB: instead of passing in the fixer, could we just return the array and loop through them in the `fix` method itself?

G-Rath: I'd prefer that too - however it'll make it harder to replace w/ `removeRange`.

Which depends on the answer to this question: do we care about persisting things like comments & spaces in fixers?

Currently, this will pass:

```
    {
      code: 'expect(/**/a/**/./**/includes/**/(/**/b/**/)/**/).toEqual(true);',
      errors: [{ messageId: 'useToContain', line: 1 }],
      output: 'expect(/**/a/**//**//**//**//**//**/).toContain(b);',
    },
    {
      code: 'expect(    a    .    includes    (    b    )    ).toEqual(true);',
      errors: [{ messageId: 'useToContain', line: 1 }],
      output: 'expect(    a                        ).toContain(b);',
    },
    {
      code:
        'expect(/**/a/**/[/**/"includes"/**/]/**/(/**/b/**/)/**/).toEqual(true);',
      errors: [{ messageId: 'useToContain', line: 1 }],
      output: 'expect(/**/a/**//**//**//**//**//**/).toContain(b);',
    },
    {
      code:
        'expect(    a    [    "includes"    ]    (    b    )    ).toEqual(true);',
      errors: [{ messageId: 'useToContain', line: 1 }],
      output: 'expect(    a                        ).toContain(b);',
    },
```

but if we use `removeRange`, it won't. In contrast however, the fixer becomes something like `removeRange(node.start, node.end)`, and makes it a lot easier for supporting square brackets.

SimenB: We should definitely preserver comments, otherwise it's not a safe fix. Whitespace is more meh, but preferably it should be kept

PR #395, opened 2019-08-15
ts-migration/prefer-to-be-undefined

G-Rath:
chore(prefer-to-be-undefined): convert to typescript


src/rules/prefer-to-be-undefined.ts
```diff
@@ -0,0 +1,75 @@
...
+} from './tsUtils';
+
+interface UndefinedIdentifier extends TSESTree.Identifier {
+  name: 'undefined';
+}
+
+const isUndefinedIdentifier = (
+  node: TSESTree.Node,
+): node is UndefinedIdentifier =>
+  node.type === AST_NODE_TYPES.Identifier && node.name === `${undefined}`;
```
SimenB:
```suggestion
  node.type === AST_NODE_TYPES.Identifier && node.name === 'undefined';
```

PR #396, opened 2019-08-15
ts-migration/prefer-to-have-length

G-Rath:
chore(prefer-to-have-length): convert to typescript


SimenB:
@G-Rath seems this lowered test coverage :( mind fixing that? Might be good enough to just ignore coverage in the js utils file as removing guards seems risky as not all rule tests test every weird combination and relies on the guard just working

PR #398, opened 2019-08-15
rename `tsUtils` to `utils`

G-Rath:
chore(tsutils): rename to `utils`


G-Rath:
@SimenB I'll probably merge this in half an hour once I'm back from lunch, rather than wait for you to wake up :)

This might be a good interlude to mention your feelings about me merging stuff, as the timezones means it can be a long wait between me being able to work on this, vs your being able to review :/

Personally, I'm confident enough to merge "small" things like this, & requesting your review when I think it's needed, and happy w/ being asked to make post-merge-review changes, if that works for you.

From what I've seen, you tend to be stretched pretty thin across a number of projects, so I'd also be happy to take up the mantle of "primary contributor/implementer" as well if you'd like, so you can just tag or assign issues to me and then just wait for the PRs :P

Let me know what you think :)

PR #403, opened 2019-08-21
Fix ts as expression type error

### LINKED ISSUE #332: prefer-to-be-null TypeError with TSAsExpression
ESLint: **6.1.0**
@typescript-eslint/parser: **1.12.0**
eslint-plugin-jest: **22.12.0**

Code:
```ts
(expect('Model must be bound to an array if the multiple property is true') as any).toHaveBeenTipped()
```

Error in `methodName`:
```
TypeError: Cannot read property 'name' of undefined
Occurred while linting /home/kael/Documents/vuetifyjs/vuetify/packages/vuetify/src/components/VItemGroup/__tests__/VItemGroup.spec.ts:53
    at methodName (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint-plugin-jest/lib/rules/util.js:70:12)
    at expectNotCase (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint-plugin-jest/lib/rules/util.js:23:101)
    at expectToBeCase (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint-plugin-jest/lib/rules/util.js:43:41)
    at CallExpression (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint-plugin-jest/lib/rules/prefer-to-be-null.js:25:45)
    at /home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Object.emit (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint/lib/linter/node-event-generator.js:253:26)
    at NodeEventGenerator.applySelectors (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint/lib/linter/node-event-generator.js:282:22)
    at NodeEventGenerator.enterNode (/home/kael/Documents/vuetifyjs/vuetify/node_modules/eslint/lib/linter/node-event-generator.js:296:14)
```

image
`node.parent.property` is undefined

G-Rath:
I think our *default* stance on casts should be to just ignore them where possible, instead checking the node they're casting, for similar reason to why we don't bother w/ `Boolean(<value>)`

There could of course be code where it's logical to take casts into account, but I think this is a sensible default.

As such, I added `expect(a.includes(b)).toEqual(0 as boolean);` to reflect that.

fixes #332

### DIFF
--- a/src/rules/utils.ts
+++ b/src/rules/utils.ts
@@ -15,6 +15,35 @@ export const createRule = ESLintUtils.RuleCreator(name => {
   return `${REPO_URL}/blob/v${version}/docs/rules/${ruleName}.md`;
 });
 
+export type MaybeTypeCast<Expression extends TSESTree.Expression> =
+  | TSTypeCastExpression<Expression>
+  | Expression;
+
+export type TSTypeCastExpression<
+  Expression extends TSESTree.Expression = TSESTree.Expression
+> = AsExpressionChain<Expression> | TypeAssertionChain<Expression>;
+
+interface AsExpressionChain<
+  Expression extends TSESTree.Expression = TSESTree.Expression
+> extends TSESTree.TSAsExpression {
+  expression: AsExpressionChain<Expression> | Expression;
+}
+
+interface TypeAssertionChain<
+  Expression extends TSESTree.Expression = TSESTree.Expression
+> extends TSESTree.TSTypeAssertion {
+  // expression: TypeAssertionChain<Expression> | Expression;
+  expression: any; // https://github.com/typescript-eslint/typescript-eslint/issues/802
+}
+
+export const followTypeAssertionChain = (
+  expression: TSESTree.Expression | TSTypeCastExpression,
+): TSESTree.Expression =>
+  expression.type === AST_NODE_TYPES.TSAsExpression ||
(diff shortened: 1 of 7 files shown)


src/rules/utils.ts
```diff
@@ -15,6 +15,35 @@ export const createRule = ESLintUtils.RuleCreator(name => {
...
+  Expression extends TSESTree.Expression = TSESTree.Expression
+> extends TSESTree.TSAsExpression {
+  expression: AsExpressionChain<Expression> | Expression;
+}
+
+interface TypeAssertionChain<
+  Expression extends TSESTree.Expression = TSESTree.Expression
+> extends TSESTree.TSTypeAssertion {
+  // expression: TypeAssertionChain<Expression> | Expression;
+  expression: any; // https://github.com/typescript-eslint/typescript-eslint/issues/802
```
G-Rath: This is fixed in `2.0.0`, so can be changed when we update.

SimenB: Cool. Just needs us to drop node 6 then.

PR #404, opened 2019-08-21
chore(utils): improve `followTypeAssertionChain` typing

G-Rath:
@SimenB I figured since you liked the original code so much, you'd enjoy some more.

Totally wasn't me forgetting to update the types in `prefer-to-contain` :joy:

### DIFF
--- a/src/rules/prefer-to-contain.ts
+++ b/src/rules/prefer-to-contain.ts
@@ -6,6 +6,7 @@ import {
 import {
   CallExpressionWithSingleArgument,
   KnownCallExpression,
+  MaybeTypeCast,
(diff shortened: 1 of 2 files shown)


SimenB:
This is just a type change, and not a behavioural one?

G-Rath:
In theory it's a minor behavioural change, as `matcher.arguments[0].value` should be `undefined` if it's a cast, but I can't make it fail w/ any of my test code, so I don't think it's worth a whole new release.

Worse case we can just release it if someone opens an issue :)

SimenB:
Perfect 🙂

PR #406, opened 2019-08-29
fix-405

### LINKED ISSUE #405: [valid-expect-in-promise] Problem chaining .then's
using eslint-plugin-jest v22.15.2

This reports an error:
```javascript
const fetch = url =>
    Promise.resolve({
        url,
        text: () => Promise.resolve('text'),
    });

// Error: Promise should be returned to test its fulfillment or rejection
test('valid-expect-in-promise', async () => {
    const text = await fetch('url')
        .then(res => res.text()) 
        .then(text => text);     

    expect(text).toBe('text');
});
```
If I remove the last `.then`, no error is reported:
```javascript
// No eslint errors
test('valid-expect-in-promise', async () => {
    const text = await fetch('url')
        .then(res => res.text()) 

    expect(text).toBe('text');
});
```
The following doesn't report error, so perhaps the problem is not in chaining:
```javascript
// No eslint errors
test('valid-expect-in-promise', async () => {
    const text = await fetch('url')
        .then(text => text)
        .then(text => text);

    expect(text).toBe('text');
});
```

G-Rath:
I'm not entirely sure why this is needed, as it doesn't seem to be different from `espree`. 

Still, all the tests are now passing, so I call it a win.

Fixes #405


G-Rath:
😬 my apologies - for some reason I thought we had `async` function tests.

I'veimplementedd the proper solution :)

SimenB:
Could we verify all non-`async` test cases have an `async` version and vice versa? Also add in the example from the issue as `valid`

G-Rath:
More than happy to tomorrow morning (it's 12am here) - also happy if you want to push to this branch :)

SimenB:
will do. thanks!

src/rules/valid-expect-in-promise.ts
```diff
@@ -154,6 +154,67 @@ const isHavingAsyncCallBackParam = (testFunction: FunctionExpression) =>
...
+ */
+const isAwaited = (functionExpression: FunctionExpression) => {
+  if (!functionExpression.async) {
+    return false;
+  }
+
+  const body = getFunctionBody(functionExpression);
+
+  if (body.type !== AST_NODE_TYPES.BlockStatement) {
+    return false;
```
SimenB: not covered

src/rules/valid-expect-in-promise.ts
```diff
@@ -154,6 +154,67 @@ const isHavingAsyncCallBackParam = (testFunction: FunctionExpression) =>
...
+          declaration.init &&
+          declaration.init.type === AST_NODE_TYPES.AwaitExpression,
+      );
+    }
+
+    if (value.type === AST_NODE_TYPES.ExpressionStatement) {
+      return value.expression.type === AST_NODE_TYPES.AwaitExpression;
+    }
+
+    return false;
```
SimenB: not covered

G-Rath:
@SimenB The uncovered lines are not actually coverable w/o some more refactoring.

We can "cover" the second `return false` by just removing it, but the first one will have to be ignored.

Having re-read the material, I think the problem here is that this rule is actually meant to be about expects *inside* promises; so `await` isn't something it should care about - that's instead handled by `valid-expect`.

The problem in this case is that we bail out if the `parent` is an `await` expression, which means the parents parent isn't checked :/

G-Rath:
@SimenB Have you had a chance to look over this yet? (just in case the email got lost again :joy:)

SimenB:
I haven't - super crunch at work for the last month... 

> The problem in this case is that we bail out if the `parent` is an `await` expression, which means the parents parent isn't checked :/

Can we recurse until we hit the `test`?

(I haven't re-read this or the issue, so context is missing atm)

SimenB:
what's the current status here?

G-Rath:
😬 

I believe the status is that I think we're got a rule that's overreaching and so coming up a bit short:

> Having re-read the material, I think the problem here is that this rule is actually meant to be about expects inside promises; so await isn't something it should care about - that's instead handled by valid-expect.

Looking at the tests & the issue, I think it should be that `valid-expect-in-promise` should only care if an expect is actually *in a promise*.

I'm a little bit of two minds on this, but I think that's because it's a bit of a confusing one to follow what with `valid-expect` vs `valid-expect-in-promise` - my reasoning is that if you take the example in the issue:

```
test('valid-expect-in-promise', async () => {
    const text = await fetch('url')
        .then(res => res.text()) 
        .then(text => text);     

    expect(text).toBe('text');
});
```

If you remove the `await` that's technically wrong but none of our rules will catch, but *jest* will because the test itself will fail; and that's where I think we should leave it, because otherwise `valid-expect-in-promise` would have to somehow account for these:

```
test('valid-expect-in-promise', async () => {
    const text = fetch('url')
        .then(res => res.text()) 
        .then(text => text);     

    expect(text).toBe('text');
    expect(await text).toBe('text');
    expect(text).resolves.toBe('text');
    expect(text).toBe(Promise.resolve); // or something...?
    expect(text).toBeInstanceOf(MyCustomPromiseClass);
});
```

G-Rath:
@SimenB have you had a chance to look over my last comment?

PR #407, opened 2019-08-29
feat(rules): add require-top-level-describe rule

### LINKED ISSUE #401: [new-rule] Enforce top level describe
Hello everyone,

Jest, and most testing library, let you write and organise your tests as you want, which is good but it can become a mess with large testing codebase.
(Somewhat similar to #362 and `no-standalone-expect` ) 
It would be super cool to have a rule to enforce a top-level describe and make sure no `test`, `it`, `beforeAll` etc... can live outside this describe.
Maybe it can be something like:
- `require-top-level-describe` 
or 
- merged with `no-standalone-expect` 
   => `no-standalone: ['error', ['test', 'hook', 'expect']]`

What do you think?

thomlom:
This PR intends to add a new ESLint rule which enforces the user to add a top-level `describe` in their test files.

Closes #401

### DIFF
--- a/src/rules/require-top-level-describe.ts
+++ b/src/rules/require-top-level-describe.ts
@@ -12,7 +12,8 @@ export default createRule({
       recommended: false,
     },
     messages: {
-      unexpectedMethod: 'All test cases must be wrapped in a describe block.',
+      unexpectedTestCase: 'All test cases must be wrapped in a describe block.',
+      unexpectedHook: 'All hooks must be wrapped in a describe block.',
     },
     type: 'suggestion',
     schema: [],
@@ -27,11 +28,16 @@ export default createRule({
           return;
         }
 
-        if (
-          (numberOfDescribeBlocks === 0 && isTestCase(node)) ||
-          isHook(node)
-        ) {
-          context.report({ node, messageId: 'unexpectedMethod' });
+        if (numberOfDescribeBlocks === 0) {
+          if (isTestCase(node)) {
+            context.report({ node, messageId: 'unexpectedTestCase' });
+            return;
+          }
+
+          if (isHook(node)) {
+            context.report({ node, messageId: 'unexpectedHook' });
+            return;
+          }
         }
       },
       'CallExpression:exit'(node: TSESTree.CallExpression) {
--- a/src/rules/require-top-level-describe.ts
+++ b/src/rules/require-top-level-describe.ts
@@ -1,0 +1,37 @@
+import { TSESTree } from '@typescript-eslint/experimental-utils';
+
+import { createRule, isDescribe, isHook, isTestCase } from './utils';
+
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description:
+        'Prevents test cases and hooks to be outside of a describe block',
+      recommended: false,
+    },
+    messages: {
+      unexpectedMethod: 'There must be a top-level describe block.',
+    },
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    let numberOfDescribeBlock = 0;
+    return {
+      CallExpression(node) {
+        isDescribe(node) && numberOfDescribeBlock++;
+        if ((isTestCase(node) || isHook(node)) && numberOfDescribeBlock === 0) {
+          context.report({ node, messageId: 'unexpectedMethod' });
+        }
+      },
+      'CallExpression:exit'(node: TSESTree.CallExpression) {
+        if (isDescribe(node)) {
+          numberOfDescribeBlock--;
+        }
+      },
+    };
+  },
+});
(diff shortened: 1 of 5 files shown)


SimenB:
Thanks! Ignore CI, the commitlint thing is somewhat broken. I should look into it at some point...

src/rules/require-top-level-describe.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    let numberOfDescribeBlock = 0;
+    return {
+      CallExpression(node) {
+        isDescribe(node) && numberOfDescribeBlock++;
```
SimenB:
```suggestion
        if (isDescribe(node)) {
          numberOfDescribeBlock++;

          return;
        }
```

src/rules/require-top-level-describe.ts
```diff
@@ -0,0 +1,37 @@
...
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description:
+        'Prevents test cases and hooks to be outside of a describe block',
+      recommended: false,
+    },
+    messages: {
+      unexpectedMethod: 'There must be a top-level describe block.',
```
SimenB:
```suggestion
      unexpectedMethod: 'All test cases must be wrapped in a describe block.',
```
?

src/rules/require-top-level-describe.ts
```diff
@@ -0,0 +1,37 @@
...
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    let numberOfDescribeBlock = 0;
+    return {
+      CallExpression(node) {
+        isDescribe(node) && numberOfDescribeBlock++;
+        if ((isTestCase(node) || isHook(node)) && numberOfDescribeBlock === 0) {
```
SimenB:
```suggestion
        if (numberOfDescribeBlock === 0 && (isTestCase(node) || isHook(node))) {
```

cheapest check, so might as well bail early if possible

src/rules/require-top-level-describe.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+    messages: {
+      unexpectedMethod: 'There must be a top-level describe block.',
+    },
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    let numberOfDescribeBlock = 0;
```
SimenB:
```suggestion
    let numberOfDescribeBlocks = 0;
```

src/rules/require-top-level-describe.ts
```diff
@@ -0,0 +1,44 @@
...
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description:
+        'Prevents test cases and hooks to be outside of a describe block',
+      recommended: false,
+    },
+    messages: {
+      unexpectedMethod: 'All test cases must be wrapped in a describe block.',
```
SimenB: should we have separate test case and hook error message?

src/rules/require-top-level-describe.ts
```diff
@@ -0,0 +1,44 @@
...
+    let numberOfDescribeBlocks = 0;
+    return {
+      CallExpression(node) {
+        if (isDescribe(node)) {
+          numberOfDescribeBlocks++;
+          return;
+        }
+
+        if (
+          (numberOfDescribeBlocks === 0 && isTestCase(node)) ||
```
SimenB: this is wrong, right?
```suggestion
          (numberOfDescribeBlocks === 0) ||
           (isTestCase(node) || sHook(node))
```
looks more correct

src/rules/require-top-level-describe.ts
thomlom: We could go with `'All test cases and hooks must be wrapped in a describe block.'` but it feels long to me.
Having separate error messages make sense to me. I'll add a test in that case!

src/rules/require-top-level-describe.ts
thomlom: Oops, my mistake 🙂

src/rules/require-top-level-describe.ts
```diff
@@ -0,0 +1,50 @@
...
+          return;
+        }
+
+        if (numberOfDescribeBlocks === 0) {
+          if (isTestCase(node)) {
+            context.report({ node, messageId: 'unexpectedTestCase' });
+            return;
+          }
+
+          if (isHook(node)) {
```
SimenB: CI complains that this is not covered. Maybe the `else` branch? A test with a `CallExpression` inside a `describe` that's neither a test nor a hook should probably hit it?

thomlom: Indeed! I added a dummy test to fix it: 

```js
'foo()',
```

But thinking of it, should we trigger a warning if the user uses a `CallExpression` that's neither a test or a hook or can we leave it as is?

SimenB: no, that's fine, I think. Jest itself uses it to e.g. skip tests on windows, or run some other sort of helper 🙂

PR #412, opened 2019-09-05
feat(matcher): add toStrictEqual as equality matcher

JrSchild:
This PR adds `toStrictEqual` as equality matcher. This fixes the following rules:
- `prefer-to-be-null`
- `prefer-to-be-undefined`
- `prefer-to-contain`
- `prefer-to-have-length`

The tests and documentation have also been updated.

### DIFF
--- a/src/rules/utils.ts
+++ b/src/rules/utils.ts
@@ -341,6 +341,7 @@ export enum ModifierName {
 enum EqualityMatcher {
   toBe = 'toBe',
   toEqual = 'toEqual',
+  toStrictEqual = 'toStrictEqual',
 }
 
 export const isParsedEqualityMatcherCall = (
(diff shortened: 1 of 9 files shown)


src/rules/utils.ts
```diff
@@ -341,6 +341,7 @@ export enum ModifierName {
 enum EqualityMatcher {
   toBe = 'toBe',
   toEqual = 'toEqual',
+  toStrictEqual = 'toStrictEqual',
```
SimenB: hah, only code change. awesome, thanks for verifying by adding some tests, though!

JrSchild:
Thanks @SimenB!

src/rules/utils.ts
G-Rath: Well, that's one question I can cross off my list :joy:

I had this pegged to ask about at some point, but never got around to it.
Cheers to all for putting a bow on this :)

PR #419, opened 2019-09-29
feat: support `allow` option in `lowercase-name` rule

G-Rath:
This makes things a bit nicer for writing tests for `supertest`.

When I finally get time to finish off my test prefix rule, it might be worth merging into that, but that'd require the whole rule to be merged in.

### DIFF
--- a/src/rules/lowercase-name.ts
+++ b/src/rules/lowercase-name.ts
@@ -54,9 +54,13 @@ const testDescription = (argument: ArgumentLiteral): string | null => {
 
 const jestFunctionName = (
   node: CallExpressionWithCorrectCalleeAndArguments,
+  allowedPrefixes: readonly string[],
 ) => {
   const description = testDescription(node.arguments[0]);
-  if (description === null) {
+  if (
+    description === null ||
+    allowedPrefixes.some(name => description.startsWith(name))
+  ) {
     return null;
   }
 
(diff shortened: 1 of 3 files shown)


G-Rath:
@SimenB I've got a few minor improvement PRs incoming, which'll probably have some merge conflicts on them.

This is the main one that needs reviewing - If you review & approve, I'm happy to merge after resolving any conflicts that come up (i.e instead of having you re-review each PR as they're merged).

src/rules/lowercase-name.ts
```diff
@@ -96,13 +103,18 @@ export default createRule({
             items: { enum: ['describe', 'test', 'it'] },
             additionalItems: false,
           },
+          allow: {
```
jeysal: Name could be clearer IMO, but e.g. `allowedStarts` sound terrible. Can you come up with anything else perhaps?

src/rules/lowercase-name.ts
```diff
@@ -96,13 +103,18 @@ export default createRule({
...
+          },
         },
         additionalProperties: false,
       },
     ],
   } as const,
-  defaultOptions: [{ ignore: [] } as { ignore: readonly JestFunctionName[] }],
-  create(context, [{ ignore }]) {
+  defaultOptions: [{ ignore: [], allow: [] }],
+  create(context, [{ ignore = [], allow = [] }]) {
```
SimenB:
```suggestion
  create(context, [{ ignore, allow }]) {
```

`defaultOptions` should do it for you

G-Rath: Agreed, but TypeScript doesn't know that :(

src/rules/lowercase-name.ts
G-Rath: I agree, but really couldn't :joy:

jeysal: Yeah I guess it's slightly better than `allow` IMO.
(Should also rename the variables further at the top of the file then)

PR #420, opened 2019-09-29
Improve types

G-Rath:
Removes most of our type casts, making things a bit nicer.

This will cause a merge conflict w/ #419

### DIFF
--- a/src/rules/__tests__/consistent-test-it.test.ts
+++ b/src/rules/__tests__/consistent-test-it.test.ts
@@ -1,5 +1,6 @@
 import { TSESLint } from '@typescript-eslint/experimental-utils';
 import rule from '../consistent-test-it';
+import { TestCaseName } from '../utils';
 
 const ruleTester = new TSESLint.RuleTester({
   parserOptions: {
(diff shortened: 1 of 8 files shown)


src/rules/__tests__/lowercase-name.test.ts
```diff
@@ -46,6 +46,7 @@ ruleTester.run('lowercase-name', rule, {
     'describe(``)',
     'describe("")',
     'describe(42)',
+    { code: 'describe(42)', options: [{ ignore: undefined }] },
```
jeysal: Does `defaultOptions` not work if the property is `undefined` instead of absent?

src/rules/consistent-test-it.ts
```diff
@@ -1,7 +1,21 @@
...
 
-export default createRule({
+export default createRule<
+  [
+    Partial<{
+      fn: TestCaseName.it | TestCaseName.test;
+      withinDescribe: TestCaseName.it | TestCaseName.test;
+    }>,
+  ],
+  'consistentMethod' | 'consistentMethodWithingDescribe'
```
jeysal: "Withing" typo

src/rules/__tests__/consistent-test-it.test.ts
```diff
@@ -11,90 +12,105 @@ ruleTester.run('consistent-test-it with fn=test', rule, {
   valid: [
     {
       code: 'test("foo")',
-      options: [{ fn: 'test' }],
+      options: [{ fn: TestCaseName.test }],
```
jeysal: Not a fan of these enums making everything so verbose (vs string union types) :|
But I guess that might just be a matter of preference

src/rules/__tests__/lowercase-name.test.ts
SimenB: if so, I think that should be reported as a bug

G-Rath: So the problem here is coverage - You could argue this as a bug, but if we do, then we lose the ability to make coverage happy, as TS doesn't know about `defaultOptions`, and so will flag the properties as possibly undefined if we don't have the fallback value.

src/rules/__tests__/consistent-test-it.test.ts
G-Rath: We have to do it in the tests b/c they're in the rule.

I'm in the same boat as @SimenB - they're extra verbose, but I find the gain of being able to have them all tied back to a single thing that can be usage queried vs saving a few bytes worth it. 

You can have the best of both worlds by making the eum const.

jeysal: The TS lang server can actually show references (identical string literals)! :smile: 
But if there's a string literal `'test'` somewhere that is not used as a `TestCaseName` it'll still show up because it's still assignable to the type, so it might show too much.

But yeah as I said this is probably preference. Except for `const enum`, they're horrible because they cannot be compiled within the scope of a single source file ;D

G-Rath:
I'm going to merge this as I think the remaining unresolved comment isn't worth blocking on: if it is a problem, it'll mean rethinking our coverage, and/or waiting for a bug fix on another package.

It's pretty easy to address in post if we want :)

src/rules/__tests__/lowercase-name.test.ts
G-Rath: I'm resolving this per my [comment here](https://github.com/jest-community/eslint-plugin-jest/pull/420#issuecomment-546645055).

PR #421, opened 2019-09-29
chore(no-commented-out-tests): remove unneeded escape and improve regex

G-Rath:
Super minor optimisations - `[xf]` is slightly faster than `x|f`, and lets us drop `()`, and `]` doesn't need to be escaped.

Saves us a few bytes all round.

### DIFF
--- a/src/rules/no-commented-out-tests.ts
+++ b/src/rules/no-commented-out-tests.ts
@@ -2,7 +2,7 @@ import { TSESTree } from '@typescript-eslint/experimental-utils';
 import { createRule } from './utils';
 
 function hasTests(node: TSESTree.Comment) {
-  return /^\s*(x|f)?(test|it|describe)(\.\w+|\[['"]\w+['"]\])?\s*\(/m.test(
+  return /^\s*[xf]?(test|it|describe)(\.\w+|\[['"]\w+['"]\])?\s*\(/m.test(
     node.value,
   );
 }


src/rules/no-commented-out-tests.ts
```diff
@@ -2,7 +2,7 @@ import { TSESTree } from '@typescript-eslint/experimental-utils';
 import { createRule } from './utils';
 
 function hasTests(node: TSESTree.Comment) {
-  return /^\s*(x|f)?(test|it|describe)(\.\w+|\[['"]\w+['"]\])?\s*\(/m.test(
+  return /^\s*[xf]?(test|it|describe)(\.\w+|\[['"]\w+['"]])?\s*\(/m.test(
```
jeysal: Missed the backslash in front of the opening bracket I think
```suggestion
  return /^\s*[xf]?(test|it|describe)(\.\w+|[['"]\w+['"]])?\s*\(/m.test(
```

G-Rath: I don't think so, as that change causes the tests to fail :)

jeysal: Oh I see, the one at the closing bracket can be removed because there's no matching opening bracket for it to close so it is treated as a normal character instead, even without an escape.
But I think it makes the regex hard for humans to parse if one bracket is escaped but its counterpart is not (like how I was confused what was going on :smile:). I think we should leave it as is and just change the part at the front.

PR #422, opened 2019-09-29
Use enum values

G-Rath:
Using the enum values as much as possible means it's easier to track where we depend on the name of jest functions; in turn makes it easier for me to create parsers (among other things) 😄


SimenB:
`commitlint` is broken when there's multiple commits in a PR, IDK why 🙁

PR #423, opened 2019-09-29
Improve prefer strict equal

G-Rath:
Allows `EqualityMatcher` to be more strictly typed to a specifically named matcher.

### DIFF
--- a/src/rules/utils.ts
+++ b/src/rules/utils.ts
@@ -339,7 +339,7 @@ export enum ModifierName {
   resolves = 'resolves',
 }
 
-enum EqualityMatcher {
+export enum EqualityMatcher {
   toBe = 'toBe',
   toEqual = 'toEqual',
   toStrictEqual = 'toStrictEqual',
--- a/src/rules/utils.ts
+++ b/src/rules/utils.ts
@@ -326,8 +326,9 @@ type MatcherName = string /* & not ModifierName */;
 type ExpectPropertyName = ModifierName | MatcherName;
 
 export type ParsedEqualityMatcherCall<
-  Argument extends TSESTree.Expression = TSESTree.Expression
-> = Omit<ParsedExpectMatcher<EqualityMatcher>, 'arguments'> & {
+  Argument extends TSESTree.Expression = TSESTree.Expression,
+  Matcher extends EqualityMatcher = EqualityMatcher
+> = Omit<ParsedExpectMatcher<Matcher>, 'arguments'> & {
   // todo: probs should also type node parent as CallExpression
   arguments: [Argument];
 };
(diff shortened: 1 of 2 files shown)


src/rules/utils.ts
```diff
@@ -338,16 +339,21 @@ export enum ModifierName {
...
-enum EqualityMatcher {
+export enum EqualityMatcher {
   toBe = 'toBe',
   toEqual = 'toEqual',
   toStrictEqual = 'toStrictEqual',
 }
 
-export const isParsedEqualityMatcherCall = (
+export const isParsedEqualityMatcherCall = <
+  MatcherName extends EqualityMatcher = EqualityMatcher
```
jeysal: Why generic with `extends EqualityMatcher` instead of just `EqualityMatcher` as the parameter type? Doesn't seem like we could ever use a subtype?

G-Rath: Not quite sure what you mean? We use `extends EqualityMatcher` b/c otherwise we'd be able to type the matcher parameter to things other than equality matchers, which'll cause runtime errors b/c TS won't know it could be those values (since `ParsedEqualityMatcher` can only matchers of type `EqualityMatcher`).

i.e You could do `isParsedEqualityMatcherCall(matcher, 'toThrowMessage')`.

jeysal: I mean why do this function and `ParsedEqualityMatcherCall` need the type parameter at all, why don't we just substitute the usages of the type parameter with `EqualityMatcher`? It looks like the type parameter at usages is always exactly `EqualityMatcher`, not a subtype.

G-Rath: We actually are using [it w/ a subtype](https://github.com/jest-community/eslint-plugin-jest/pull/423/files#diff-032ef4c5f083206966108daa47382007R36), but it's not explicit b/c TypeScript can infer it :)

In this case, TypeScript correctly infers that `matcher` is type `ParsedEqualityMatcherCall<TSESTree.Expression, EqualityMatcher.toEqual>`:

```
const anotherEqualityMatcher: ParsedEqualityMatcherCall<any, EqualityMatcher.toBe> = matcher;

Error:(38, 17) TS2322: Type 'ParsedEqualityMatcherCall<Expression, EqualityMatcher.toEqual>' is not assignable to type 'ParsedEqualityMatcherCall<any, EqualityMatcher.toBe>'.
  Type 'ParsedEqualityMatcherCall<Expression, EqualityMatcher.toEqual>' is not assignable to type 'Pick<ParsedExpectMatcher<EqualityMatcher.toBe, ExpectMember<EqualityMatcher.toBe, CallExpression | ArrayExpression | ArrayPattern | ArrowFunctionExpression | AssignmentExpression | AssignmentPattern | ... 151 more ... | undefined>>, "name" | "node">'.
    Types of property 'name' are incompatible.
      Type 'EqualityMatcher.toEqual' is not assignable to type 'EqualityMatcher.toBe'.
```

jeysal: Take a look at https://github.com/jest-community/eslint-plugin-jest/commit/4b3bbbc0f954301038ebed63e97d1e97e019a7a4, it still passes that way so the additional parameter in the type signature seems unnecessary.

G-Rath: Yes, it'll still pass in this instance, but it means it's not as accurately typed, as the code example in my last comment is no longer possible, and won't error.

Given the complexity of AST parsing, and having converted most of the rules to TypeScript, I know firsthand how these small type improvements can help greatly down the line :)

jeysal: Alright, I'll leave the decision what is over-engineered and what is adequate to you, just wanted to raise the question and I think it's good we at least talked about the reasoning :+1:

G-Rath: I do appreciate the check, so thanks for talking it out.

The main thing is that this way it'll "just work" for others writing rules, who might not know TS as well, and so not think/know they could improve things by adding these couple of lines.

This is about where I draw the line on possible over-engineering.

PR #425, opened 2019-09-30
feat(rules): prefer-hooks-on-top

### LINKED ISSUE #356: New rule suggestion: Place all hooks at the top of the block.
I would like to suggest a new rule, to place all hook blocks before any other blocks.

So, this is good:

```js
describe('module', () => {
  let a
  
  beforeEach(() => {
    a = new A();
  })
  
  test('function', () => {
    expect(a).not.toBeUndefined();
  })
})
```

And this is bad:

```js
describe('module', () => {
  let a
  
  test('function', () => {
    expect(a).not.toBeUndefined();
  })
  
  beforeEach(() => {
    a = new A();
  })
})
```

M4rk9696:
Fixes #356 

Approach is to maintain a stack of `boolean` which gets marked as `true` when a hook is not before a test

Also an edge case
```js
describe("foo" () => {
  beforeEach(() => {
    //some hook code
  });
  
  // Not affected by rule
  someSetup();

  afterEach(() => {
    //some hook code
  });
  test("bar" () => {
    some_fn();
  });
});
```

### DIFF
--- a/src/rules/__tests__/prefer-hooks-on-top.test.ts
+++ b/src/rules/__tests__/prefer-hooks-on-top.test.ts
@@ -9,13 +9,13 @@ const ruleTester = new TSESLint.RuleTester({
 
 ruleTester.run('basic describe block', rule, {
   valid: [
-    `describe("foo" () => {
+    `describe("foo", () => {
         beforeEach(() => {
         });
         someSetupFn();
         afterEach(() => {
         });
-        test("bar" () => {
+        test("bar", () => {
           some_fn();
         });
       });`,
@@ -37,7 +37,7 @@ ruleTester.run('basic describe block', rule, {
       errors: [
         {
           messageId: 'noHookOnTop',
-          column: 3,
+          column: 11,
           line: 7,
         },
       ],
@@ -47,19 +47,19 @@ ruleTester.run('basic describe block', rule, {
 
 ruleTester.run('multiple describe blocks', rule, {
   valid: [
-    `describe.skip("foo" () => {
+    `describe.skip("foo", () => {
         beforeEach(() => {
         });
         beforeAll(() => {
         });
-        test("bar" () => {
+        test("bar", () => {
           some_fn();
         });
       });
-      describe("foo" () => {
+      describe("foo", () => {
         beforeEach(() => {
         });
-        test("bar" () => {
+        test("bar", () => {
           some_fn();
         });
       });`,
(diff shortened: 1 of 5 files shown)


src/rules/__tests__/prefer-hooks-on-top.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+const ruleTester = new TSESLint.RuleTester({
+  parserOptions: {
+    ecmaVersion: 6,
+  },
+});
+
+ruleTester.run('basic describe block', rule, {
+  valid: [
+    [
```
G-Rath: I tend to prefer using template strings, as they make it easier to modify things.

My main gripe however is that we've not got consistency across our tests 😭

M4rk9696:
CI is a lint issue, I'll push a commit along with the review changes

src/rules/__tests__/prefer-hooks-on-top.test.ts
M4rk9696: Yes, a common standard would be good. I didn't know some tests used template strings.

This is from `no-test-return-statement`, it would seem that there is extra indentation added to it. Is that ok? because that would require modifying the location of the error as well
```js
    `
    it("one", function () {
      expect(1).toBe(1);
    });
    `
  ]
```

G-Rath: And there you've hit the nail on moving to templates: to what level do we indent?

The snippet you've pasted is what I'd go w/ - the indentation is fine, as it's still completely valid test wise. It's either recalculate or call` .trim`.

Either way, the test still represent the same thing 🙂

src/rules/__tests__/prefer-hooks-on-top.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        '  });',
+        '  test("bar", () => {',
+        '    some_fn();',
+        '  });',
+        '});',
+      ].join('\n'),
+      errors: [
+        {
+          messageId: 'noHookOnTop',
+          column: 3,
```
M4rk9696: I tried with changing this test to template string. The column changed from `3` to `11` because it had to take into account the extra indentation.

I'm ok with both the approaches.

A question though, for a first time contributor to this project, would it be ok to expect them to know about the extra indent. It sound like a gotcha they would face when writing the test

G-Rath: Personally I think it's fine - You can always de-indent the block, and modify it like that, or like I said call trim, then once you know you've gotten it correct, add the indent amount or copy them from the test error.

While still not optimal, I think it's better than `join` as it makes it very speedy to add new test cases.

M4rk9696:
@G-Rath Changed it to string templates

PR #428, opened 2019-10-05
feat(rules): option for  rule to allow uppercase for top level describes

gayathrigs27:


src/rules/lowercase-name.ts
```diff
@@ -93,7 +120,7 @@ export default createRule({
         properties: {
           ignore: {
             type: 'array',
-            items: { enum: ['describe', 'test', 'it'] },
+            items: { enum: ['describe', 'test', 'it', 'top'] },
```
G-Rath: I think if you're going to do it this way, it would be better to use a boolean option since `top` isn't a `jest` keyword and so doesn't really make sense having it implemented in the same manner of `isTestCase` & `isDescribe`.

That would also be a good position to later extend it to be an array accepting the same properties as `ignore`.

.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules/
 coverage/
 lib/
 *.log
+package-lock.json
```
G-Rath: I assume you're adding this b/c you've used `npm`? I would advise against that, as it'll mean you're getting different versions of packages than CI, since `npm` won't consume (nor update!) the `yarn.lock`.

Eitherway, I'd prefer not to have this in the `.gitignore`, as I think it's confusing.

gayathrigs27: okay

src/rules/lowercase-name.ts
gayathrigs27: My understanding is that if I gave ignore=top, the lowercase rule will not apply to the first parent element(describe/test/it) the inner describe/test should have this rule.
If I do an early return as you suggested the inner elements will also be skipped off the rule. Which is 
```
describe("Xyz abc"){
  test("Cde"){}
}
``` will be a valid one

Comment (G-Rath) [reply src/rules/lowercase-name.ts]: Just to be clear, this is the sort of early return I had in mind:

```
if(ignoreTopLevel && isTopMostJestFunction(node)) {
  return;
}
```

Since `CallExpression` is called for *every* `CallExpression` that ESLint finds, the early return won't interfere with the inner jest function calls.

File: src/rules/lowercase-name.ts [hunk 3]
```diff
@@ -103,14 +130,25 @@ export default createRule({
...
   create(context, [{ ignore }]) {
+    let counter: number = 0
     return {
+
       CallExpression(node) {
         if (!isJestFunctionWithLiteralArg(node)) {
           return;
         }
+
+        if (isTestCase || isDescribe) {
```
G-Rath: This conditional will always be true, as you're not actually calling the functions 😬 

([TypeScript can actually detect this in the upcoming 3.7 release](https://github.com/microsoft/TypeScript/pull/33178) 🎉)

src/rules/__tests__/lowercase-name.test.ts
```diff
@@ -225,3 +225,37 @@ ruleTester.run('lowercase-name with ignore=it', rule, {
...
+    {
+      code: 'test(`Foo`, function () {})',
+      options: [{ ignore: [TopCase.top] }],
+    },
+    {
+      code: 'it(`Foo`, function () {})',
+      options: [{ ignore: [TopCase.top] }],
+    },
+  ],
+  invalid: [],
```
G-Rath: Would you be able to add some `invalid` test cases as well please?

Would be great if you could include some multi-top level ones like in my above comment; and ideally mixing the casing of the inner functions 😄

src/rules/__tests__/lowercase-name.test.ts
```diff
@@ -225,3 +225,37 @@ ruleTester.run('lowercase-name with ignore=it', rule, {
   ],
   invalid: [],
 });
+
+ruleTester.run('uppercase-name with ignore=top', rule, {
+  valid: [
```
G-Rath: Could you add a few muti-top level test cases, i.e:

```
describe('MyFirstClass', () => {
  ...
});

describe('MySecondClass', () => {
  ...
});
```

Ideally a couple would be nice, w/ a mix of single & multiple inner methods as well :)

src/rules/lowercase-name.ts
```diff
@@ -52,6 +53,32 @@ const testDescription = (argument: ArgumentLiteral): string | null => {
...
+const jestTopFunctioName = (
+  node: CallExpressionWithCorrectCalleeAndArguments,
+  counter: number,
+) => {
+  const description = testDescription(node.arguments[0]);
+  if (description === null) {
+    return null;
+  }
+
+  if (counter !== 1) {
```
G-Rath: You could do this check first, before anything else to reduce the work being done slightly; alternatively this could be merged w/ the `description === null` check.

src/rules/lowercase-name.ts
```diff
@@ -52,6 +53,32 @@ const testDescription = (argument: ArgumentLiteral): string | null => {
...
+    return null;
+  }
+
+  const firstCharacter = description.charAt(0);
+
+  if (!firstCharacter) {
```
G-Rath: Given that `charAt` is typed as returning a `string`, rather than `string | null`, I'd say it would be better to just check that `description` is truthy, rather than do this.

i.e change the earlier `description === null` condition to just `!description`, and remove this check.

src/rules/lowercase-name.ts
```diff
@@ -103,14 +130,25 @@ export default createRule({
...
           return;
         }
+
+        if (isTestCase || isDescribe) {
+          counter += 1
+        }
+
+        const topMethod = jestTopFunctioName(node, counter);
+
+        const topMethodIgnore = topMethod && ignore.includes(TopCase.top);
```
G-Rath: The only value being used in this conditional that is computed as part of the `CallExpression` function is `topMethod`, meaning you could move `ignore.includes(TopCase.top)` out of this function, and assign it to a variable like you do w/ `counter`.

That way we won't be iterating over `ignores` for every `CallExpression` that makes it this far - a micro optimisation for sure, but one that's zero-effort, so might as well.

src/rules/__tests__/lowercase-name.test.ts
G-Rath: Could you add a few muti-top level test cases, i.e:

```
describe('MyFirstClass', () => {
  ...
});

describe('MySecondClass', () => {
  ...
});
```

Ideally a couple would be nice, w/ a mix of single & multiple inner methods as well :)

src/rules/__tests__/lowercase-name.test.ts
G-Rath: Would you be able to add some `invalid` test cases as well please?

Would be great if you could include some multi-top level ones like in my above comment; and ideally mixing the casing of the inner functions 😄

gayathrigs27: added the testcases

gayathrigs27:
@G-Rath made changes based on your comments. Can you review it again and help me out

package.json
```diff
@@ -37,7 +37,8 @@
     "typecheck": "tsc -p ."
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^1.13.0"
+    "@typescript-eslint/experimental-utils": "^1.13.0",
+    "build": "^0.1.4"
```
G-Rath: Why do we need `build`?

src/rules/utils.ts
```diff
@@ -652,6 +660,13 @@ export const isDescribe = (
   );
 };
 
+export function isTopMostJestFunction(node: TSESTree.CallExpression): boolean {
```
G-Rath: I'd prefer that this be inside the rule, rather than in `utils` - iirc there is another rule doing something similar, but I'm going to be doing a spring-cleaning of the codebase this weekend, so I can refactor this into a utility method in that PR.

For now, it'll just be cleaner to have it in the rule :)

src/rules/utils.ts
```diff
@@ -550,7 +550,15 @@ export enum TestCaseProperty {
...
-export type JestFunctionName = DescribeAlias | TestCaseName | HookName;
+export enum TopCase {
+  'top' = 'top',
+}
+
+export type JestFunctionName =
+  | DescribeAlias
+  | TestCaseName
+  | HookName
+  | TopCase;
```
G-Rath: `top` is not a valid function provided by `Jest`, so it shouldn't be in this union.

src/rules/lowercase-name.ts
```diff
@@ -118,13 +141,20 @@ export default createRule<
...
   defaultOptions: [{ ignore: [], allowedPrefixes: [] }],
   create(context, [{ ignore = [], allowedPrefixes = [] }]) {
     return {
       CallExpression(node) {
         if (!isJestFunctionWithLiteralArg(node)) {
           return;
         }
+
+        const topMethod = jestTopFunctioName(node, isTopMostJestFunction(node));
+        if (topMethod && ignore.includes(TopCase.top)) {
```
G-Rath: Could you replace `TopCase.top` w/ an `ignoreTopLevel` option, [as](https://github.com/jest-community/eslint-plugin-jest/pull/428/files#r331757372) previously [discussed](https://github.com/jest-community/eslint-plugin-jest/pull/428/files#r332680009)?

You can also inline `topMethod`, since it's not used elsewhere :)

G-Rath:
@gayathrigs27 Let me know if you're interested in championing this PR, as it would be cool to get this across the line!

I'm happy to pick it up & finish it of if you'd prefer 😄

PR #429, opened 2019-10-07
Add `jest/no-mocks-import` to recommended rule set

unindented:
@SimenB suggested adding this at a later point: https://github.com/jest-community/eslint-plugin-jest/pull/246#discussion_r277167611

Could that be now?


jeysal:
I think @SimenB wants to do one pass of adding things to recommended that should be recommended, right before a major release

Issue #273, opened 2019-05-31
New rule: avoid-prefixing-tests-with


G-Rath:
Basically, this would catch things like `it('it should throw an error', () => { ...`

I figure it would be best as a more generic rule, that took a list of strings that should be considered as undesired to have at the start of the messages for test calls, with "it" being the default.

(The rule name could be improved :/)

ranyitz:
It would also help if you want to remove an accidental space:

```js
it(' should remove space in the beginning', () => { ...
```

I think that adding a rule like that which is fixable (removes the unwanted string) would be nice.

SimenB:
Sure, PR welcome! 👍

suganya-sangith:
Shall I start working on it?

G-Rath:
@suganya-sangith if you've got the time, go for it! I'd love to see this rule, but haven't had the time to finish off my implementation.

Feel free to tag me in the PR, and I'll review it 🙂

PR #433, opened 2019-10-17
feat(rules): prefer-valid-title (#273)

suganya-sangith:
As part of this, added a valid-title rule which checks for accidental space and duplicate prefix

@G-Rath Kindly review the PR and let me know if there are any changes

### DIFF
--- a/docs/rules/valid-title.md
+++ b/docs/rules/valid-title.md
@@ -1,0 +1,65 @@
+# Disallow duplicate setup and teardown hooks (no-duplicate-hooks)
+
+A describe/ test block should not contain accidentalSpace or duplicatePrefix.
+
+## Rule Details
+
+**duplicatePrefix**
+
+A describe/ test block should not start with duplicatePrefix
+
+Examples of **incorrect** code for this rule
+
+```js
+test('test foo', () => {});
+it('it foo', () => {});
+
+describe('foo', () => {
+  test('test bar', () => {});
+});
+
+describe('describe foo', () => {
+  test('bar', () => {});
+});
+```
+
+Examples of **correct** code for this rule
+
+```js
+test('foo', () => {});
+it('foo', () => {});
+
+describe('foo', () => {
+  test('bar', () => {});
+});
+```
+
+**accidentalSpace**
+
+A describe/ test block should not contain accidentalSpace
+
+Examples of **incorrect** code for this rule
+
+```js
+test(' foo', () => {});
+it(' foo', () => {});
+
+describe('foo', () => {
+  test(' bar', () => {});
+});
+
+describe(' foo', () => {
+  test('bar', () => {});
+});
+```
+
+Examples of **correct** code for this rule
+
+```js
+test('foo', () => {});
+it('foo', () => {});
+
+describe('foo', () => {
+  test('bar', () => {});
+});
+```
--- a/src/rules/valid-title.ts
+++ b/src/rules/valid-title.ts
@@ -1,0 +1,64 @@
+import {
+  createRule,
+  getNodeName,
+  getStringValue,
+  isDescribe,
+  isStringNode,
+  isTestCase,
+} from './utils';
+
+import { TSESTree } from '@typescript-eslint/experimental-utils';
+
+const trimFXprefix = (word: string) =>
+  ['f', 'x'].includes(word.charAt(0)) ? word.substr(1) : word;
+
+const getNodeTitle = (node: TSESTree.CallExpression): string | null => {
+  const [argument] = node.arguments;
+
+  return isStringNode(argument) ? getStringValue(argument) : null;
+};
+
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Enforce valid titles',
+      recommended: false,
+    },
+    messages: {
+      duplicatePrefix: 'should not have duplicate prefix',
+      accidentalSpace: 'should not have space in the beginning',
+    },
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (!isDescribe(node) && !isTestCase(node)) return;
+
+        const title = getNodeTitle(node);
+        if (!title) return;
+
+        if (title.trimLeft().length !== title.length) {
+          context.report({
+            messageId: 'accidentalSpace',
+            node,
+          });
+        }
+
+        const nodeName = trimFXprefix(getNodeName(node.callee));
+        const [firstWord] = title.split(' ');
+
+        if (firstWord.toLowerCase() === nodeName) {
+          context.report({
+            messageId: 'duplicatePrefix',
+            node,
+          });
+        }
+      },
+    };
+  },
+});
--- a/src/__tests__/rules.test.ts
+++ b/src/__tests__/rules.test.ts
@@ -3,7 +3,7 @@ import { resolve } from 'path';
 import plugin from '../';
 
 const ruleNames = Object.keys(plugin.rules);
-const numberOfRules = 40;
+const numberOfRules = 41;
 
 describe('rules', () => {
   it('should have a corresponding doc for each rule', () => {
--- a/src/rules/__tests__/valid-title.test.ts
+++ b/src/rules/__tests__/valid-title.test.ts
@@ -1,0 +1,161 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+import rule from '../valid-title';
+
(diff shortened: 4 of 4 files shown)


src/rules/utils.ts
```diff
@@ -652,6 +652,16 @@ export const isDescribe = (
   );
 };
 
+export const getNodeTitle = (node: TSESTree.CallExpression): string | null => {
```
G-Rath: ~Since this is only used by `valid-title`, I think it'd be best to have it in that rule for now, just to keep things in scope.~

~We can always move it into utils if it turns out useful in other rules :)~

After seeing it's usage in `valid-title`, I think it would make sense to just in-line this method for now, as it's doing a pretty standard node guard check :)

src/rules/utils.ts
```diff
@@ -652,6 +652,16 @@ export const isDescribe = (
...
+  const [argument] = node.arguments;
+
+  if (!isStringNode(argument) || AST_NODE_TYPES.Literal !== argument.type) {
```
G-Rath: iirc I think this can be replaced w/ `isValidAccessor` (might not be the exact name) - `getAccessorValue` definitely has a complementary guard that should cover all the expected conditions for string parameters.

src/rules/valid-title.ts
```diff
@@ -0,0 +1,56 @@
...
+  defaultOptions: [],
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (!isDescribe(node) && !isTestCase(node)) return;
+
+        const title = getNodeTitle(node);
+        if (!title) return;
+
+        if (title.charAt(0) === ' ') {
```
G-Rath: You could use `trimStart` here to detect any & all spaces - I can't imagine any title that'd you want to start w/ multiple spaces, a la how you'd not want one space.

suganya-sangith:
@G-Rath Thanks for taking time to review it, will work on the suggested changes 👍

src/rules/utils.ts
suganya-sangith: I saw similar checks are made in rules like lower-case, no-identical-title etc. Hence I added this method thinking I would do necessary changes in other rules to use this method. 

If you feel this should be an inline thing will move to valid-title.

G-Rath: I think for now I'd prefer it to be inlined, b/c for the same of avoiding extra functions & export, but thanks for raising this - I'll give the code base a sweep when I have a chance, and see if there are any duplications we could refactor into a utility function :)

suganya-sangith:
@G-Rath Made all the changes, expect getAccessorValue for which I could not find an alternate function to get the title.

src/rules/valid-title.ts
```diff
@@ -0,0 +1,73 @@
...
+const trimStart = (title: string) => /^\s/.test(title);
+
+const getNodeTitle = (node: TSESTree.CallExpression): string | null => {
+  const [argument] = node.arguments;
+
+  if (!isStringNode(argument) || AST_NODE_TYPES.Literal !== argument.type) {
+    return null;
+  }
+
+  return getAccessorValue(argument);
```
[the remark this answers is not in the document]
suganya-sangith: Thanks for the explanation 👍 Will spend time to understand the types available 😁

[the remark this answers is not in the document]
G-Rath: No problem - I wrote all of the types and guards, and still get confused.

It's actually why I wrote them so grainaulay: The nature of AST parsing means it's best to not try and do fancy things, and instead layer everything to build on the previous type.

As such, the types are something like:

A `StringLiteral` is a `Literal` whose value is of type String (i.e `'hello world'`).
A `TemplateLiteral` is a Template Literal, that has only one quasis (i.e `` `hello world` ``)
  * `quasis` is a property on `TemplateLiteral`s, that is an array of `TemplateElement`s.
    * No idea why it's `quasis` - it's called that in the spec, but I can't find a definition of the word.
  * In theory this should mean only "simple" template literals that don't have any expressions, as a `TemplateElement` sort of represents the "stringy" bits of a template literal i.e `` `hello ${world}` `` has 2 `TemplateElement`s, the first being `hello ${` (w/ a value of `hello `) and the second being `}` (w/ a value of `''`).
  * In practice, this does work, but I think it might be better to check the length of `expression`, to see if it's 0
    * I know `no-identical-title` checks both `isStringNode` & the expression length if it's a template, and fails if I remove the expression length check, which iirc means `isStringNode` isn't doing its job right.

A `StringNode` is a `node` that's either a `StringLiteral` OR a `TemplateLiteral`.

-----

That ends the "string" side of nodes. You then have `Identifiers`, which are pretty much what they say on the tin: they're something that "identifies" as something - best way to think of them is as plaintext "names" or "words" - anything else is another expression, i.e `[]` is an `ArrayExpression`.

Identifiers are what you use to access properties, methods, fields, etc via dot notation. They're also how variables are used. 

A `KnownIdentifier` is an `Identifier` whose name we "known"
  * This is actually a bit of a cheeky one - when we say "known", the "least known" `Identifier` has a `name` of type `string`; that's it. Which the case for *all* `Identifier`s anyway, and is the default. The name was more about reducing confusion between *our* generically typed `Identifier` and `TSESTree.Identifier`, of which ours extends.

--------
An `AccessorNode` is then either a `StringNode` OR a `KnownIdentifier`.

The whole backstory behind `AccessorNode` is that you can access things in two ways in JS: Dot notation, and Bracket notation.

Dot accessors are `Identifiers` - nice and easy. Bracket accessors however are done using "a string" in square brackets; so they're `StringNode`s (b/c they could be string literals or template expressions - they can also be `Identifier`s themselves, b/c variables but we don't support that b/c we're not trying to be TS 😉).

As such, we have `isSupportedAccessor`, which guards on if a `node` is an `AccessorNode` that we "support". Then for when we want to get the value of an `AccessorNode`, we have `getAccessorValue`, which normalises getting the value of the `node` you're working w/.

------

And there you go, that's pretty much it in a nutshell 😄 

While on the surface it can look very complex, it's actually pretty simple to the point of being elegant (IMO at least 😉) - it's done as just a composition chain, which means you just have to follow the type chain down each step to get an idea of what the previous step is made up of.

In saying that, once you throw the actual AST tree into the mix, it's *very* easy to get confused & mentally tongue tied. The above is covering only the property accessor side; theres also the `expect`/modifier/matcher utils, which are also hopefully as elegant, but still their own kettle of fish.

Finally, I'm meant to be writing a whole new set of parsers/guards for the jest globals (`describe`, `test`, `it`, etc), but that's slow going for now (plus what we have is working, so it's less needed).

-----

This is pretty much the best write up I've done on this so far, and something I've been meaning to do for a while. Cheers!

src/rules/valid-title.ts
```diff
@@ -0,0 +1,64 @@
...
+  defaultOptions: [],
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (!isDescribe(node) && !isTestCase(node)) return;
+
+        const title = getNodeTitle(node);
+        if (!title) return;
+
+        if (title.trimLeft().length !== title.length) {
```
G-Rath: You can actually just compare the two strings - by comparing `length`, *in theory* if the string somehow changes it could still pass:

```
String.prototype.trimLeft = function() { return this.toUpperCase(); }

const myString = ' hello world';

console.log('are strings equal?', myString.trimLeft().length === myString.length ? 'yes' : 'no');
```

This is o/c a really horrible thing that no one should ever be doing, and it saves us a few characters, but just wanted to point it out as it's useful to keep in mind for other conditionals 🙂

jeysal:
Yeah seems like a recommended candidate for next major version to me as well

src/rules/valid-title.ts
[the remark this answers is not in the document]
suganya-sangith: Wow, that's so much of information, Thank you very much for taking time to write this up 😊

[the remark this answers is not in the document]
G-Rath: No problem! I've tried to write this up a few times before, but didn't get very far, so when I started writing this comment I just let things snowball so that as much info as possible is on paper :joy:

docs/rules/valid-title.md
```diff
@@ -0,0 +1,65 @@
+# Disallow duplicate setup and teardown hooks (no-duplicate-hooks)
```
mrtnzlml: This title seems to be invalid. 😊 (probably copy-paste mistake)

mrtnzlml: Addressed here: https://github.com/jest-community/eslint-plugin-jest/pull/434 🎉

suganya-sangith: @mrtnzlml Thanks for catching it. I have used the template from other read me to maintain the consistency, but missed to change the title 😬

PR #438, opened 2019-10-26
chore: refresh lockfile

SimenB:
image

None of these matter, but annoying warning. hopefully this is enough to shut it up

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,9 +3,9 @@
 
 
 "@babel/cli@^7.4.4":
-  version "7.5.5"
-  resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.5.5.tgz#bdb6d9169e93e241a08f5f7b0265195bf38ef5ec"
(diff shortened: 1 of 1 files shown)


SimenB:
> Do you have any opinions on updating `package.json` w/ the latest versions?

no opinion beyond it creating a more noisy diff. so no opinion 😅

G-Rath:
> no opinion beyond it creating a more noisy diff

Fair enough. Speaking of dependencies, do you have any roadmap on dropping `node@6`, or is it just "when jest does"?

SimenB:
jest has dropped it on master, but I don't think we depend on that over here. I'm down with making a release whenever, just need to go over and decide on new recommended rules

SimenB:
>Let me know when/how you'd like to do that (and if you'd like my input).

now is fine with me 👍 wanna open up a PR for discussion?

G-Rath:
> now is fine with me

It's 11:40pm on a Saturday of a 3 day weekend here in New Zealand, so I've nothing stopping me from sleeping in tomorrow - lets do it!

> wanna open up a PR for discussion

Sure thing - I'll make one removing `node@6` where I can find, and go from there :)

PR #439, opened 2019-10-26
chore(ci): add GH Actions CI

SimenB:
might be interesting to add e.g. https://github.com/iCrawl/action-tsc for inline comments, but lets see if this works _at all_ first


SimenB:
@G-Rath hopefully this'll still release, only using the action instead. Feel free to nag me if that doesn't happen.

(hopefully danger goes away from future PRs - I've deleted the action)

G-Rath:
@SimenB cheers, I'll keep an eye on it and let you know if it causes trouble 😄

PR #440, opened 2019-10-26
drop `node@6` support

G-Rath:
:tada:

@SimenB how much do you want to include in this PR? 

(It's my first time doing an OSS major, so some light hand-holding might be required 😬)

Recommended rule-wise, I think we're decent, but should add:

* valid-title
* no-empty-title
* no-identical-title

Also might be worthing adding:
* no-standalone-expect
* no-try-expect
* no-if

In that order.

On a separate note, I have been wondering what you think about `valid-title` & `no-empty-title`, as we could either:
  * Merge `no-empty-title` into `valid-title`
  * Split `valid-title` up into seperate rules (`no-untrimmed-title`, `no-prefixed-title`, etc)
  * Leave them seperate, and continue on as is (but that still sort of leaves the question on if we should make new rules or expand `valid-title`).

I don't have a big opinion either way, but wondered if there was any general eslint convention or recommendation on if it's nicer to have lots of options or lots of rules?

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [6.x, 8.x, 10.x, 12.x, 13.x]
+        node-version: [8.x, 10.x, 12.x, 13.x]
     runs-on: ubuntu-latest
 
     steps:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ node_js:
 - 12
 - 10
 - 8
-- 6
 - node
 before_install:
 - curl -o- -L https://yarnpkg.com/install.sh | bash
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,6 +5,6 @@ module.exports = {
   plugins: ['replace-ts-export-assignment'],
   presets: [
     '@babel/preset-typescript',
-    ['@babel/preset-env', { targets: { node: 6 } }],
+    ['@babel/preset-env', { targets: { node: 8 } }],
   ],
 };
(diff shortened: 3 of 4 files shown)


SimenB:
the commit needs a `BREAKING CHANGE` in the body in order for it to trigger a new major when included in a release. we can bump dependencies as well in this PR, although we might wanna target a `next` branch until we're ready to make a release.

---

>Merge `no-empty-title` into `valid-title`

yeah, that makes sense to me. personally I prefer options as they're more discoverable than specific rule (e.g. like in this case - title validation)

---

I think we should remove `no-alias-methods` from the recommended rules, and add `expect-expect`, `no-commented-out-tests`, `no-export`, and `no-try-expect`.

I'd also like to consider making `no-test-callback` recommended, to move people over to a promise interface.

thoughts?

@jeysal @thymikee thoughts on recommended rules (or other stuff)?

G-Rath:
That all sounds reasonable. 

I'll make a next branch when I get in up the morning (~4 hours).

I agree w/ those rule changes for the most part - I'm on the fence about `no-commented-out-tests`, just b/c I know that sometimes it's useful to comment out a test that's known to be failing or in prep for some change, but it's easy to disable so 🤷‍♂ 

> I think we should remove no-alias-methods from the recommended rules

Any particular reason?

> I'd also like to consider making no-test-callback recommended, to move people over to a promise interface.

Completely down for this.

SimenB:
> I'm on the fence about `no-commented-out-tests`, just b/c I know that sometimes it's useful to comment out a test that's known to be failing or in prep for some change, but it's easy to disable so 🤷‍♂

you can ignore the lint failure while working on tests, but you shouldn't pass e.g. CI with commented out tests (and easy enough to disable if you really need to keep it in the source code).

we might even consider merging that rule into `no-disabled-tests` since it's essentially the same thing, only through syntax instead of api

SimenB:
see #443

PR #445, opened 2019-10-26
Merge no empty title into valid title

G-Rath:
I'm happy to backport this to minor via deprecation if needed.


markmsmith:
Just as a heads-up, because the commits were prefixed with `chore` rather than `feat`, this breaking change didn't show up in the change log :(  
Also, this commit has a typo in the spelling of breaking change, with `BREAKING CHANEG` (which would also foil the changelog + SemVer), though that seems to be on the merge to `master` rather than `next`.  
https://github.com/jest-community/eslint-plugin-jest/commit/bb2d09ea2084c12d6dc0b6fdb88c1fa8fa897273
I had to track this down when my linting started failing.  
Just thought you should know

G-Rath:
@markmsmith well that's embarrassing 😬 thanks for letting me know.

Chore is listed in the github release page: https://github.com/jest-community/eslint-plugin-jest/releases/tag/v23.0.0

I assume these are done automatically, but @SimenB might have done some manual work on it as he manages the release flow :)

@SimenB we also seem to be missing a CHANGELOG entry for 23.0.0 in general?

markmsmith:
Yeah, if it mentioned the rule `no-empty-title` by name, and that it was being folded in to `valid-title`, it would be a lot easier to find for others that run in to the same error.

G-Rath:
Worst case I can just recreate the rule as a wrapper that re-exports `valid-title` w/ the `deprecated` property set to `valid-title`.

Worst-worst-case I can create an empty rule w/ the `deprecated` set, and maybe that throws saying "use valid-title".

PR #451, opened 2019-10-27
chore: test on both eslint 5 and 6

SimenB:


src/rules/__tests__/consistent-test-it.test.ts
```diff
@@ -1,9 +1,10 @@
 import { TSESLint } from '@typescript-eslint/experimental-utils';
+import resolveFrom from 'resolve-from';
 import rule from '../consistent-test-it';
 import { TestCaseName } from '../utils';
 
 const ruleTester = new TSESLint.RuleTester({
-  parser: 'espree',
+  parser: resolveFrom(require.resolve('eslint'), 'espree'),
```
SimenB: this allows us to not depend on `espree`. eslint 5 & 6 uses different versions, so it'd be painful

G-Rath: What's the background for supporting 5 & 6? (Just for my personal info)

SimenB: no reason not to support 5

G-Rath: Oh, the pain you're referring to is w/ if we have a dependency on a specific version 5 or 6, for consumers of the library.

Sweet as :)

SimenB: it'd be painful for us to install both v4 and v5 of espree depending on which version of eslint we were testing against

PR #453, opened 2019-10-27
chore: update recommended config

SimenB:
@G-Rath @jeysal @thymikee thoughts? I'd like to add `no-test-callback` as well, but it might be too much?


G-Rath:
Just FYI `valid-title` is marked as recommended in `next`, b/c I didn't get around to doing the others. You might like to rebase so that that's done here for completeness :)

I'm also good w/ adding `no-test-callback`

SimenB:
cool, let's do it then. simple enough for people to disable if they don't want it anyways 👍

PR #460, opened 2019-10-28
Better support valid each

### LINKED ISSUE #458: `valid-title`: false positives when using `it.each`
With version 23, if I'm using something like this:

```js

describe( 'myFunction', () => {
	it.each( [
		[ 123, true ],
		[ 456, true ],
		[ 789, false ],
		[ 0, false ],
	] )( 'should do something',
		( number, myBool ) => {
			expect( muFunction( number ) ).toBe( myBool );
		},
	);
} );
```

The `jest/valid-title` complains about the array not being a valid title.

I see that there have been a few changes to `valid-title` in the last releases. So perhaps an issue like #334 is now appearing again, or at least this one here is very similar.

`src/rules/__tests__/valid-title.test.ts` doesn't have any test cases for `it.each` (except a single one from #436). Adding some would be good I think.

G-Rath:
Fixes #458

### DIFF
--- a/src/rules/__tests__/valid-describe.test.ts
+++ b/src/rules/__tests__/valid-describe.test.ts
@@ -11,16 +11,11 @@ const ruleTester = new TSESLint.RuleTester({
 
 ruleTester.run('valid-describe', rule, {
   valid: [
-    'describe.each()()',
-    'describe.each(() => {})()',
     'describe["each"]()()',
     'describe["each"](() => {})()',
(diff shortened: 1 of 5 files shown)


G-Rath:
This'll likely be less than perfect, but too tired to be bothered trying to weed every last use-case out.

Much rather have them reported by users w/ IRL code to test against.
Either way, it's better than what we had before, which was no support :D

PR #464, opened 2019-10-31
Fix 463

G-Rath:
We can later refactor `isDescribe` to be overloaded w/ a `property` argument, but for now this'll fix the specific issue w/ `each`.


SimenB:
The commit message on the merge was wrong, so this didn't teigger a release

G-Rath:
Oh yeah bugger - thanks for catching that.

I can reword it on master if that's a thing we do..? Otherwise, I'll make another commit w/ the right message for a release.

SimenB:
force pushing the latest seems reasonable

PR #469, opened 2019-11-07
fix(no-test-callback): don't provide fix for `async` functions

### LINKED ISSUE #466: no-test-callback autofix can produce invalid code
Hello team,

With the release of `v23` our team had issues with the new recommended rules, and in particular `no-test-callback`.

```js
test('my test', async (done) => {
  await myAsyncTask();
  expect(true).toBe(false);
  done();
});
```

This code is marked as `Illegal usage of test callback` which is a fair warning; I would say the `done` usage is more useless than illegal (but just a matter of wording here).

This code will be transformed to:
```js
test('my test', async () => {await new Promise(done => {
  await myAsyncTask();
  expect(true).toBe(false);
  done();
})});
```

Which is:
- not very pretty 
- invalid
   - as await now lives in non `async` method
   - if you put the `async` keyword you can encounter an other linter error `no-async-promise-executor`

---

This is not a bug nor an urgent matter, but I would be in favor of letting the user change his/her code instead of modifying the behaviour of the test without the user paying attention.

Have a great day,

G-Rath:
Fixes #466 

@SimenB could you check that the commit message is correct for triggering a new release?

### DIFF
--- a/src/rules/__tests__/no-test-callback.test.ts
+++ b/src/rules/__tests__/no-test-callback.test.ts
@@ -54,20 +54,38 @@ ruleTester.run('no-test-callback', rule, {
     },
     {
       code: 'test("something", async done => {done();})',
-      errors: [{ messageId: 'illegalTestCallback', line: 1, column: 25 }],
-      output:
-        'test("something", async () => {await new Promise(done => {done();})})',
+      errors: [{ messageId: 'useAwaitInsteadOfCallback', line: 1, column: 25 }],
     },
     {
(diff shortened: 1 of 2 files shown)


SimenB:
Itæd trigger a release yeah, just missing coverage 👍

G-Rath:
> just missing coverage 👍

:(

Forgot to remove the `callback.async` check, and then IntelliJ decided to truncate the schemas for all my shelved patches....

PR #480, opened 2019-11-23
chore: minor housekeeping

G-Rath:
General house cleaning - removed some comments that are no longer needed, and chucked in `eslint-plugin-eslint-comments` b/c why not.

Also refactored `no-alias-methods` to use a map instead of an array, which is technically slightly faster, but also just generally nicer logic-wise.

Finally, fixed the message of `require-to-throw-message` to use the right data property :D
That also made me realise it could be a good candidate for the new suggestions api? provided it lets you take in user input.

@SimenB I adjusted the `eslint` config to use `eslint-config-prettier` - let me know if I've just refactored away a work around for something?

### DIFF
--- a/src/rules/no-alias-methods.ts
+++ b/src/rules/no-alias-methods.ts
@@ -9,7 +9,7 @@ export default createRule({
       recommended: 'warn',
     },
     messages: {
-      replaceAlias: `Replace {{ replace }}() with its canonical name of {{ canonical }}()`,
+      replaceAlias: `Replace {{ alias }}() with its canonical name of {{ canonical }}()`,
     },
     fixable: 'code',
     type: 'suggestion',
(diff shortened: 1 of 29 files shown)


src/rules/no-alias-methods.ts
```diff
@@ -9,30 +9,28 @@ export default createRule({
...
-      ['toHaveBeenLastCalledWith', 'lastCalledWith'],
-      ['toHaveBeenNthCalledWith', 'nthCalledWith'],
-      ['toHaveReturned', 'toReturn'],
-      ['toHaveReturnedTimes', 'toReturnTimes'],
-      ['toHaveReturnedWith', 'toReturnWith'],
-      ['toHaveLastReturnedWith', 'lastReturnedWith'],
-      ['toHaveNthReturnedWith', 'nthReturnedWith'],
-      ['toThrow', 'toThrowError'],
-    ];
+    // The Jest methods which have aliases. The canonical name is the first index of each item.
```
SimenB: this comment lies now

G-Rath: 😱 so it does!

PR #496, opened 2019-12-18
feat(valid-expect): warn on `await expect()` with no assertions

motiz88:
Makes the following an error under the `valid-expect` rule:
```js
await expect('something'); // No assertion was called on expect().
```

## Rationale

As this plugin already has a rule for `expect`-with-no-assertions, I thought I'd add a similar one for `await expect`.

Speaking from recent experience, the simple mistake of awaiting the wrong thing (`await expect('something')` instead of `await 'something'`) can be a pain to track down and fix, as its effects are likely to be felt far from the line containing the actual mistake (which itself will never throw at runtime).

### DIFF
--- a/src/rules/valid-expect.ts
+++ b/src/rules/valid-expect.ts
@@ -90,6 +90,12 @@ const isAcceptableReturnNode = (
     AST_NODE_TYPES.AwaitExpression,
   ].includes(node.type);
 
+const isNoAssertionsParentNode = (node: TSESTree.Node): boolean =>
+  node.type === AST_NODE_TYPES.ExpressionStatement ||
+  (node.type === AST_NODE_TYPES.AwaitExpression &&
+    node.parent !== undefined &&
+    node.parent.type === AST_NODE_TYPES.ExpressionStatement);
+
 const promiseArrayExceptionKey = ({ start, end }: TSESTree.SourceLocation) =>
   `${start.line}:${start.column}-${end.line}:${end.column}`;
 
@@ -280,10 +286,7 @@ export default createRule<[{ alwaysAwait?: boolean }], MessageIds>({
 
       // nothing called on "expect()"
       'CallExpression:exit'(node: TSESTree.CallExpression) {
-        if (
-          isExpectCall(node) &&
-          node.parent.type === AST_NODE_TYPES.ExpressionStatement
-        ) {
+        if (isExpectCall(node) && isNoAssertionsParentNode(node.parent)) {
           context.report({ messageId: 'noAssertions', node });
         }
       },
--- a/docs/rules/valid-expect.md
+++ b/docs/rules/valid-expect.md
@@ -79,6 +79,7 @@ expect();
 expect().toEqual('something');
 expect('something', 'else');
 expect('something');
+await expect('something');
 expect(true).toBeDefined;
 expect(Promise.resolve('hello')).resolves;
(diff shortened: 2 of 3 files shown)


src/rules/valid-expect.ts
```diff
@@ -90,6 +90,15 @@ const isAcceptableReturnNode = (
     AST_NODE_TYPES.AwaitExpression,
   ].includes(node.type);
 
+const isNoAssertionsParentNode = (node: TSESTree.Node): boolean => {
+  return (
+    node.type === AST_NODE_TYPES.ExpressionStatement ||
+    (node.type === AST_NODE_TYPES.AwaitExpression &&
+      node.parent != null &&
```
G-Rath: `node.parent` can never be null, so you can just check for `undefined` (or better yet, just do `node.parent`)

src/rules/valid-expect.ts
```diff
@@ -90,6 +90,15 @@ const isAcceptableReturnNode = (
     AST_NODE_TYPES.AwaitExpression,
   ].includes(node.type);
 
+const isNoAssertionsParentNode = (node: TSESTree.Node): boolean => {
```
G-Rath: 🐘 Would prefer this to be shorthanded 🙂

src/rules/valid-expect.ts
```diff
@@ -90,6 +90,11 @@ const isAcceptableReturnNode = (
     AST_NODE_TYPES.AwaitExpression,
   ].includes(node.type);
 
+const isNoAssertionsParentNode = (node: TSESTree.Node): boolean =>
+  node.type === AST_NODE_TYPES.ExpressionStatement ||
+  (node.type === AST_NODE_TYPES.AwaitExpression &&
+    node.parent!.type === AST_NODE_TYPES.ExpressionStatement);
```
G-Rath: It's generally best to avoid using the non-null assertion operator where possible, as it reduces the effectiveness of TypeScript, and it is technically possible for `node.parent` to be undefined.

Instead just using `node.parent &&` should be enough :)

motiz88: `node.parent &&` doesn't typecheck because it's not strictly a boolean. I'll go with `node.parent !== undefined &&` if that's OK with you

motiz88:
> It's interesting that it doesn't already pick this up

There might be a case for inverting the check on the parent node altogether. What if we made it so `expect()` _had_ to be the `object` of a `MemberExpression`, and anything else would error with `noAssertions`?

Is the following a common/allowed pattern at all? (And if so, how do we feel about forbidding it?)

```js
const ex = expect(fooBar());
ex.toBe(true);
ex.not.toBe(false);
// etc
```

This doesn't have to block this PR. Just curious :)

G-Rath:
> Is the following a common/allowed pattern at all? (And if so, how do we feel about forbidding it?)
> What if we made it so expect() had to be the object of a MemberExpression, and anything else would error with noAssertions?

That's an interesting thought, which I will think on over the holidays.

When I converted this plugin to TypeScript, I looked into exactly that sort of thing, and my conclusion was that there is a fine balance to try and strike w/ AST handling, as so to support generally as much as possible w/o being too open ended that you need to do a whole bunch of checks anytime you want to sneeze.

I don't think that change would be a big win for us, as it's not something we've got lots of bugs on (that I know of?), but I'll have a think on it b/c it is an interesting thought.

Issue #432, opened 2019-10-16
test.concurrent support


daniellacosse:
it appears that the linter ignores `test.concurrent` calls, for example - currently I have the `jest/no-empty-title` rule enabled:

![Screen Shot 2019-10-16 at 16 20 16](https://user-images.githubusercontent.com/3759828/66955798-1cdfbf00-f031-11e9-89b1-ab5ecbdcd681.png)

with just `test`:

![Screen Shot 2019-10-16 at 16 20 28](https://user-images.githubusercontent.com/3759828/66955809-25d09080-f031-11e9-970d-0419466199b4.png)

G-Rath:
Ah yes - I missed that b/c of it's undocumented nature.

From memory it shouldn't be too much trouble to add it, just a matter of finding the time :joy:

daniellacosse:
@G-Rath point me at where i need to do the thing and I'll do the thing

G-Rath:
@daniellacosse from memory, a couple of things need doing:

* `concurrent` needs to be added as a value in the right enu,
* `concurrent` should be checked for what it supports, as while I suspect it is only `test.concurrent`, we need to know if `it.concurrent`, `test.concurrent.only`, `test.concurrent.each`, etc are valid in jest.
* All the rules should be checked, as `jest/no-empty-title` isn't the only one that this affects (or maybe it is - either way, need to check).

daniellacosse:
fyi it's definitely all the rules, i've just included `jest/no-empty-title` as an example.

SimenB:
> but `each` and `concurrent` can't be used together.

that's a bug, mind opening up and issue at facebook/jest?

daniellacosse:
seems there is one - https://github.com/facebook/jest/issues/8985

leonardovillela:
Can i take this issue?

SimenB:
Go for it!

G-Rath:
closed by #498 / #502

PR #498, opened 2019-12-30
feat(rules): add .concurrent support

leonardovillela:
# Refers to #432 

## Key points
- Maybe a little over test, but i tried to main the same logic for test scenarios.
- In some places i cheat in `ts` typing, i miss a TSTree definition for `.concurrent` call expression. Mainly in two files above.
  - [src/rules/no-focused-tests.ts](https://github.com/leonardovillela/eslint-plugin-jest/blob/5527d002b3459a646c58da8a4febfffa0c09c78c/src/rules/no-focused-tests.ts#L38)
  - [src/rules/utils.ts](https://github.com/leonardovillela/eslint-plugin-jest/blob/5527d002b3459a646c58da8a4febfffa0c09c78c/src/rules/utils.ts#L635)
- `xit` don't support use of `.concurrent`, so i don't change the rules for it.
  - This one i opened a [issue](https://github.com/facebook/jest/issues/9354) on jest, just to check if it's a valid behavior.
- `xtest` don't support use of `.concurrent`, so i don't change the rules for it.
- `.todo` don't support use of `.concurrent`, so i don't change the rules for it.

### DIFF
--- a/src/rules/no-focused-tests.ts
+++ b/src/rules/no-focused-tests.ts
@@ -18,11 +18,22 @@ const testFunctions = new Set<string>([
   ...validTestCaseNames,
 ]);
 
-const isConcurrentExpression = (expression: TSESTree.Node) =>
-  getNodeName(expression) &&
-  new RegExp(
-    `(${validTestCaseNames.join('|')})\.${TestCaseProperty.concurrent}`,
-  ).test(getNodeName(expression) as string);
+interface ConcurrentExpression extends TSESTree.MemberExpression {
+  parent: TSESTree.MemberExpression;
+}
+
+const isConcurrentExpression = (
+  expression: TSESTree.MemberExpression,
+): expression is ConcurrentExpression => {
+  const nodeName = getNodeName(expression);
+
+  return (
+    !!nodeName &&
+    new RegExp(
+      `(${validTestCaseNames.join('|')})\.${TestCaseProperty.concurrent}`,
+    ).test(nodeName)
+  );
+};
 
 const matchesTestFunction = (object: TSESTree.LeftHandSideExpression) =>
   'name' in object &&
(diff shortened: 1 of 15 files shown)


leonardovillela:
@G-Rath and @SimenB

src/rules/no-focused-tests.ts
```diff
@@ -5,16 +5,25 @@ import {
...
 const testFunctions = new Set<string>([
   DescribeAlias.describe,
-  TestCaseName.test,
-  TestCaseName.it,
+  ...validTestCaseNames,
 ]);
 
+const isConcurrentExpression = (expression: TSESTree.Node) =>
+  getNodeName(expression) &&
+  new RegExp(
```
G-Rath: I'm not a big fan of this, as it's hard to understand what the logic will be at runtime - instead, it would be better to use standard conditional checks against the AST structure, like we do with `isDescribeEach` & co.

src/rules/utils.ts
```diff
@@ -641,7 +658,8 @@ export const isTestCase = (
       node.callee.object.type === AST_NODE_TYPES.Identifier &&
       TestCaseName.hasOwnProperty(node.callee.object.name) &&
       node.callee.property.type === AST_NODE_TYPES.Identifier &&
-      TestCaseProperty.hasOwnProperty(node.callee.property.name))
+      TestCaseProperty.hasOwnProperty(node.callee.property.name)) ||
+    isConcurrentTestCase(node)
```
G-Rath: I think it would be better to bring the checks required from `isConcurrentTestCase` into here, to save on duplicating conditional checks as it should just be a matter of adding a third layer.

Once this PR is merged, I'm going to swing by this method to see about breaking it up a bit and making it more readable.

leonardovillela:
First of all, thanks for the review @G-Rath. I'll do the requested changes today or tomorrow. About the refactor i can try to do this, but in other PR to not block this one, if you get me some guidance ou even open an issue describing your ideias would be nice. But if you prefer to yourself do this refactor and not open an issue describing it, feel free.

G-Rath:
@leonardovillela No problem - The refactor definitely doesn't need to be done anytime soon, nor block.

It's probably better at this point for me to do it, as I've got to re-review my code anyway to get a handle on the exact changes that need to be done, and it ties in with some other maintenance work I need to do :)

I think for now let's focus on getting those other comments addressed, and then we can decide where to go after my review post-changes.

leonardovillela:
@G-Rath changes made.

leonardovillela:
@G-Rath thanks. I wish to contribute more in the future.

leonardovillela:
@G-Rath do you have a recommend issue that to me take?

PR #500, opened 2020-01-04
Tidy up

G-Rath:
Some tidying up - finally got around to refactoring some old things, such as `lowercase-name`.

@SimenB I was planning on merging this w/o review, as I've tried to keep the changes pretty low key, but it's evening now so I'll wait until tomorrow morning so that you can give it a glance if you get the chance.

### DIFF
--- a/src/rules/consistent-test-it.ts
+++ b/src/rules/consistent-test-it.ts
@@ -131,14 +131,13 @@ function getPreferredNodeName(
   nodeName: string,
   preferredTestKeyword: TestCaseName.test | TestCaseName.it,
 ) {
-  switch (nodeName) {
-    case TestCaseName.fit:
-      return 'test.only';
-    default:
-      return nodeName.startsWith('f') || nodeName.startsWith('x')
-        ? nodeName.charAt(0) + preferredTestKeyword
-        : preferredTestKeyword;
+  if (nodeName === TestCaseName.fit) {
+    return 'test.only';
   }
+
+  return nodeName.startsWith('f') || nodeName.startsWith('x')
+    ? nodeName.charAt(0) + preferredTestKeyword
+    : preferredTestKeyword;
 }
 
 function getOppositeTestKeyword(test: TestCaseName.test | TestCaseName.it) {
--- a/src/rules/consistent-test-it.ts
+++ b/src/rules/consistent-test-it.ts
@@ -1,4 +1,8 @@
-import { AST_NODE_TYPES } from '@typescript-eslint/experimental-utils';
+import {
+  AST_NODE_TYPES,
+  TSESLint,
+  TSESTree,
+} from '@typescript-eslint/experimental-utils';
 import {
   TestCaseName,
   createRule,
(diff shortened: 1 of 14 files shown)


src/rules/lowercase-name.ts
```diff
@@ -4,20 +4,18 @@ import {
   TSESTree,
```
SimenB: lovely refactors here 👍

src/rules/no-truthy-falsy.ts
```diff
@@ -29,9 +29,9 @@ export default createRule({
         }
 
         context.report({
-          data: { methodName: matcher.name }, // todo: rename to 'matcherName'
-          messageId: 'avoidMessage', // todo: rename to 'avoidMatcher'
+          messageId: 'avoidMatcher',
           node: matcher.node.property,
+          data: { matcherName: matcher.name },
```
SimenB: keep it alphabetical?

src/rules/utils.ts
```diff
@@ -1,10 +1,10 @@
-import { parse as parsePath } from 'path';
 import {
   AST_NODE_TYPES,
   ESLintUtils,
   TSESLint,
   TSESTree,
 } from '@typescript-eslint/experimental-utils';
+import { parse as parsePath } from 'path';
```
SimenB: they were sorted as node core at the top

SimenB:
@G-Rath I just pushed up 9dadd3c48508966785a2b9e46dfa1351c26441b6 which enforces sort order of imports. In theory you should be able to drop your first commit, and just rebase, sorry about the conflict 😬

src/rules/utils.ts
G-Rath: Ah fair enough - I was using WebStorms optimize imports function, which sadly doesn't support that sort of logic.

Having it as an eslint rule should take care of that :)

src/rules/consistent-test-it.ts
```diff
@@ -56,7 +71,7 @@ export default createRule<
     let describeNestingLevel = 0;
 
     return {
-      CallExpression(node) {
+      CallExpression(node: TSESTree.CallExpression) {
```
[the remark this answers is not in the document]
G-Rath: I added this because of WebStorm - for some reason it doesn't quite infer the `node` type properly, so won't show guard changes without the explicit type.

I've been meaning to make an issue on YouTrack for it, but haven't gotten around to doing so yet.

Happy to remove, but we're not overly consistent, and the selector ones are required (i.e when doing things like `'CallExpression:exit'`)

src/rules/no-truthy-falsy.ts
G-Rath: We typically have it in this order from what I saw - I personally prefer this order as `messageId` & `node` are required and so will be present for all calls to `context.report`, then followed by `data` which is optional.

G-Rath:
@SimenB Cheers for the review - I've merged for now as I've got a few more things I want to work on today and want to reduce conflicts, but am happy to continue discussing :)

PR #503, opened 2020-01-05
ci(lint): pass `configFile` to `commitlint` action

G-Rath:
@SimenB Turns out that the default is to look for `commitlint.config.js`, and to pass if it's not found 😬

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,6 +9,8 @@ jobs:
         with:
           fetch-depth: 0
       - uses: wagoid/commitlint-github-action@v1.2.2
+        with:
+          configFile: './package.json'
         env:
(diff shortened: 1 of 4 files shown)


G-Rath:
I also bumped `actions/checkout` to `v2`, as it's meant to be faster since it only checkouts the first commit 🤷‍♂

SimenB:
Weird it doesn't error out. Does it work? Mind pushing a non-conventional commit to verify?

G-Rath:
I've been trying it on my companies [ESLint config repo](https://github.com/squareweave/eslint-config-ackama), which is how I found it out.

But yeah I'll push a commit up and we can see what happens

G-Rath:
@SimenB there you go - it passes after I remove the config property 🤷‍♂

PR #504, opened 2020-01-06
feat(rules): add support for function declaration as test case

blake-newman:
Add support for the following test file structure.

```js
test('my test', myTest)

function myTest() {
  expect(true).toBe(true)
}
```

Methods that are directly referenced will be ananalyzed for the
following rules `expect-expect` `no-if` `no-test-return-statement`,
`no-try-expect`

### DIFF
--- a/src/rules/utils.ts
+++ b/src/rules/utils.ts
@@ -633,6 +633,27 @@ export const isHook = (
   node.callee.type === AST_NODE_TYPES.Identifier &&
   HookName.hasOwnProperty(node.callee.name);
 
+export const getTestCallExpressionsFromDeclaredVariables = (
+  declaredVaiables: TSESLint.Scope.Variable[],
+): Array<JestFunctionCallExpression<TestCaseName>> => {
+  return declaredVaiables.reduce<
+    Array<JestFunctionCallExpression<TestCaseName>>
+  >(
+    (acc, { references }) =>
+      acc.concat(
+        references
+          .map(({ identifier }) => identifier.parent)
+          .filter(
+            (node): node is JestFunctionCallExpression<TestCaseName> =>
+              !!node &&
+              node.type === AST_NODE_TYPES.CallExpression &&
+              isTestCase(node),
+          ),
+      ),
+    [],
+  );
+};
+
 export const isTestCase = (
   node: TSESTree.CallExpression,
 ): node is JestFunctionCallExpression<TestCaseName> =>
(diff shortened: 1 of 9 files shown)


src/rules/utils.ts
```diff
@@ -633,6 +633,37 @@ export const isHook = (
...
+export const getFunctionDeclarationTestCallExpressions = (
+  context: TSESLint.RuleContext<string, any>,
+  node: TSESTree.FunctionDeclaration,
+): Array<JestFunctionCallExpression<TestCaseName>> => {
+  const variables = context.getDeclaredVariables(node);
+  return variables.reduce<Array<JestFunctionCallExpression<TestCaseName>>>(
+    (acc, { references }) => {
+      const callExpressions = references
+        .map(({ identifier }) => identifier.parent)
+        .filter<TSESTree.CallExpression>(isCallExpression)
```
G-Rath: It would be better to inline this guard, due to its simplicity & b/c it's generally not a good idea to have the guarded variable optional as it complicates the type narrowing.

src/rules/utils.ts
```diff
@@ -633,6 +633,37 @@ export const isHook = (
...
+  node: TSESTree.FunctionDeclaration,
+): boolean => {
+  return getFunctionDeclarationTestCallExpressions(context, node).length > 0;
+};
+
+export const getFunctionDeclarationTestCallExpressions = (
+  context: TSESLint.RuleContext<string, any>,
+  node: TSESTree.FunctionDeclaration,
+): Array<JestFunctionCallExpression<TestCaseName>> => {
+  const variables = context.getDeclaredVariables(node);
```
G-Rath: Since this is the only need for `context`, I think it would be better to pass the results of `getDeclaredVariables` in.

blake-newman:
@G-Rath Thanks for the fast review; i have now updated based on your requested changes and ensured coverage in tests meets 100%

src/rules/utils.ts
```diff
@@ -633,6 +633,25 @@ export const isHook = (
   node.callee.type === AST_NODE_TYPES.Identifier &&
   HookName.hasOwnProperty(node.callee.name);
 
+export const getTestCallExpressionsFromDeclaredVariables = <
+  R extends JestFunctionCallExpression<TestCaseName>
```
G-Rath: I don't think this generic is actually adding much  - let's keep things simple, and just stick with `JestFunctionCallExpression<TestCaseName>` if possible :)

src/rules/utils.ts
```diff
@@ -633,6 +633,25 @@ export const isHook = (
   node.callee.type === AST_NODE_TYPES.Identifier &&
   HookName.hasOwnProperty(node.callee.name);
 
+export const getTestCallExpressionsFromDeclaredVariables = (
+  declaredVaiables: TSESLint.Scope.Variable[],
+): Array<JestFunctionCallExpression<TestCaseName>> => {
+  return declaredVaiables.reduce<
+    Array<JestFunctionCallExpression<TestCaseName>>
+  >((acc, { references }) => {
+    const callExpressions = references
```
G-Rath: You can remove this intermediate variable by just calling `accc.concat(...)`.

That'll also let you short hand the arrow function, and avoid the extra overhead from copying both arrays.

Issue #471, opened 2019-11-09
[expect-expect] support chained function names


erunion:
*[see #418 for previous discussion]*

Currently expect/expect doesn't recognize supertest or sinon assertions. For example:

```
const spy = sinon.spy()
it('should work as expected', async () => {
  // test code here

  sinon.assert.calledOnce(spy);
});
```

Adding sinon (or calledOnce) into the `assertFunctionNames` configuration block doesn't get recognized, so it would be cool to:

* Be able to supply full, chained, functions to `assertFunctionNames` like `sinon.assert.calledOnce`.
* It would also be neat to be able to do the same but with a glob-style syntax. Something to the effect of `sinon.assert.*` or for supertest cases, `request.*.expect`.

erunion:
Turns out adding `sinon.assert.calledOnce` into `assertFunctionNames` works, and has tests, already! It would still be nice to be able to do glob syntax since right now the only way to kind of get supertest matching is to add your common supertest constant name (likely `request`) as a function name, and that's way too general. Not to mention having to duplicate every sinon assertion.

G-Rath:
Yeah I was pretty sure I'd already implemented that for this exact reason 😄 

While I understand globbing sounds useful, I feel like it could be an easy footgun since it could result in non-assertion method being counted by mistake.

Plus as soon as you have "globbing" people generally expect it to work in a similar way to how it does everywhere else, so we'd have to add a new dependency for something like micromatch or globby.

None of these are a hard no, but I'll have it on the backburner for now.

jrnail23:
@erunion, I just stopped by to create an feature request for this exact issue (supertest's `expect`), so you already beat me to it.

@G-Rath, I tend to agree with your concerns about globbing, but would it be possible to (optionally) detect usages of a _method_, rather than only standalone functions?

erunion:
Instead of globbing, I've been playing around with a PR that'd support regex matching instead, but haven't made much progress so far due to my lack of ESLint parsing internals.

folke:
This issue can be closed since https://github.com/jest-community/eslint-plugin-jest/commit/295ca9a6969c77fadaa1a42d76e89cae992520a6 implements this

PR #508, opened 2020-01-10
feat(expect-expect): support chained function names (#471)

folke:
Added support for chained assertion function names and method calls. (see also #471)

This makes it easy to support 3rd party assertion libraries like [SuperTest](https://www.npmjs.com/package/supertest) with the option `{ "assertFunctionNames": ["expect", "request.get.expect"] }`

Examples of **correct** code for deep assertion functions with the
`{ "assertFunctionNames": ["expect", "tester.foo.expect"] }` option:

```js
/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "tester.foo.expect"] }] */
test('nested expect method call', () => {
  class Foo {
    expect(k) {
      return k;
    }
  }
  let tester = {
    foo: function() {
      return new Foo();
    },
  };
  tester.foo().expect(123);
});
```

### DIFF
--- a/docs/rules/expect-expect.md
+++ b/docs/rules/expect-expect.md
@@ -76,22 +76,23 @@ test('returns sum', () =>
 );
 ```
 
-Examples of **correct** code for deep assertion functions with the
-`{ "assertFunctionNames": ["expect", "tester.foo.expect"] }` option:
+Examples of **correct** code for working with the HTTP assertions library
+[SuperTest](https://www.npmjs.com/package/supertest) with the
+`{ "assertFunctionNames": ["expect", "request.get.expect"] }` option:
 
 ```js
-/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "tester.foo.expect"] }] */
-test('nested expect method call', () => {
-  class Foo {
-    expect(k) {
-      return k;
-    }
-  }
-  let tester = {
-    foo: function() {
-      return new Foo();
-    },
-  };
-  tester.foo().expect(123);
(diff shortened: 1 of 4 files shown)


docs/rules/expect-expect.md
```diff
@@ -75,3 +75,23 @@ test('returns sum', () =>
     .run();
 );
 ```
+
+Examples of **correct** code for deep assertion functions with the
+`{ "assertFunctionNames": ["expect", "tester.foo.expect"] }` option:
+
+```js
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "tester.foo.expect"] }] */
+test('nested expect method call', () => {
```
G-Rath: I think it would be good to have an example of `SuperTest`, as it's a pretty easy real world example - just taking the first code block from their README will be fine :)

folke: Just added an example for SuperTest!

G-Rath:
Thanks!

folke:
Fixing the coverage turned out to be trickier than expected, since `getNodeName` will almost never return `null`

Added `(function(){})()` as a test case for `no-jasmine-globals`.

Does the trick :)

folke:
Thanks for reviewing and merging in the PR!! ✨

G-Rath:
> Fixing the coverage turned out to be trickier than expected

Interesting - I'll add a note to have a look into this as part of the maintenance run I have planned, just to make sure it's all good :)

Now that you've changed the implementation, it's possible that the overload signature should be tweaked, but regardless your fix here seems to work so that's the main thing :D

PR #509, opened 2020-01-10
feat(expect-expect): support glob patterns for assertFunctionNames (#471)

folke:
Added support for glob patterns to specify assertion function names like `request.*.expect` (see
[micromatch](https://github.com/micromatch/micromatch) for syntax)

Implements #471

Examples of **correct** code for working with the HTTP assertions library
[SuperTest](https://www.npmjs.com/package/supertest) with the `{ "assertFunctionNames": ["expect", "request.*.expect"] }` option:

```js
/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "request.*.expect"] }] */
const request = require('supertest');
const express = require('express');

const app = express();

describe('GET /user', function() {
  it('responds with json', function(done) {
    request(app)
      .get('/user')
      .expect('Content-Type', /json/)
      .expect(200, done);
  });
});
```

### DIFF
--- a/docs/rules/expect-expect.md
+++ b/docs/rules/expect-expect.md
@@ -42,7 +42,9 @@ it('should work with callbacks/async', () => {
 
 ### `assertFunctionNames`
 
-This array option whitelists the assertion function names to look for.
+This array option whitelists the assertion function names to look for. Function
+names can be a glob pattern like `request.*.expect` (see
+[micromatch](https://github.com/micromatch/micromatch) for syntax)
 
 Examples of **incorrect** code for the `{ "assertFunctionNames": ["expect"] }`
 option:
@@ -78,10 +80,10 @@ test('returns sum', () =>
 
 Examples of **correct** code for working with the HTTP assertions library
 [SuperTest](https://www.npmjs.com/package/supertest) with the
-`{ "assertFunctionNames": ["expect", "request.get.expect"] }` option:
+`{ "assertFunctionNames": ["expect", "request.*.expect"] }` option:
 
 ```js
-/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "request.get.expect"] }] */
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "request.*.expect"] }] */
 const request = require('supertest');
 const express = require('express');
 
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "typecheck": "tsc -p ."
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^2.5.0"
+    "@typescript-eslint/experimental-utils": "^2.5.0",
+    "micromatch": "^4.0.2"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
@@ -50,6 +51,7 @@
     "@semantic-release/git": "^7.0.17",
     "@types/eslint": "^6.1.3",
     "@types/jest": "^24.0.15",
+    "@types/micromatch": "^4.0.0",
     "@types/node": "^12.6.6",
     "@typescript-eslint/eslint-plugin": "^2.5.0",
     "@typescript-eslint/parser": "^2.5.0",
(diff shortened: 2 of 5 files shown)


folke:
> the extra dependency

If you think it would make more sense to create a simple utility function to match wildcards instead of using micromatch, then I could definitely  look into that as well.

Ultimately, we don't really need full blown glob pattern matching for this use case. Our import aren't path names, so any functionality related to glob patterns for file paths etc, are not needed.

Imho the only two expression that seem useful are `*` and `**`

Example of `*`
```js
request.get().expect('foo')
request.post().expect('foo')

assertionFunctionNames = ["request.*.expect"]
```

Example of `**`
```js
request.get().foo().bar().alpha().whatever().expect('foo')

assertionFunctionNames = ["request.**.expect"]
```

Both expressions could be easily implemented with a `RegExp` without depending on micromatch

G-Rath:
> If you think it would make more sense to create a simple utility function to match wildcards instead of using micromatch, then I could definitely look into that as well.

I'd be keen to what you could come up with!

I'm not against using micromatch - my dependency point was a weak one; it's just more about avoiding a new dependency is nice, and globbing tends to be the sort of thing people expect to work everywhere, and so as soon as you add it it has the potential to open a whole can of worms as people assume it's everywhere.

While my main concerns are valid concerns, that's relative to the actual size of the issue, which is very small 🙂 

-------

In saying that I'd be very interested in seeing what you could come up with if you want to have a play around!

folke:
@G-Rath 
Just finished a small utility function that basically provides the same functionality as needed.

See below:
```ts
var assert = require('assert');

function starRegex(pattern: string) {
  return RegExp(
    pattern
      .split('.')
      .map(x => {
        if (x === '**') return '[a-zA-Z\\.]*';
        if (x.includes('**'))
          throw `Double star wildcards can only occur between dots as in request.**.expect`;
        return x.replace('*', '[a-zA-Z]*');
      })
      .join('\\.'),
  );
}

export function starMatch(string: string, pattern: string): boolean;
export function starMatch(string: string, pattern: string[]): boolean;
export function starMatch(string: string, pattern: string | string[]): boolean {
  const patterns = Array.isArray(pattern) ? pattern : [pattern];
  for (const p of patterns) {
    if (starRegex(p).test(string)) return true;
  }
  return false;
}

const test: Array<[string, string, boolean]> = [
  ['request.get.post.expect', 'request.**.expect', true],
  ['request.get.post.expect', 'request.*.expect', false],
  ['request.get.expect', 'request.*.expect', true],
  ['request.get.expect', 'request.**.expect', true],
  ['request.get.expectSomething', 'request.**.expect*', true],
];

for (const [string, pattern, expectedResult] of test) {
  const result = starMatch(string, pattern);
  console.log({
    string,
    pattern,
    regex: starRegex(pattern),
    result,
  });
  assert(result === expectedResult);
}
```

Should I create a PR with this change? I'm happy with this approach or using micromatch :)

G-Rath:
@folke go wild!

Even if it doesn't get merged, I'm keen to check it out, as it looks impressive :D

folke:
@G-Rath should I add this to rules/utils.ts? Or in a seperate file like rules/starmatch.ts?

folke:
utils.ts seems to be the best location.

Function is pretty small:
```ts
/**
 * Checks if string matches any of the given star patterns
 * String is what is returned by getNodeName, like request.get.expect
 * Pattern examples:
 *   request.*.expect
 *   request.**.expect
 *   request.**.expect*
 */
export function starMatch(string: string, pattern: string): boolean;
export function starMatch(string: string, pattern: string[]): boolean;
export function starMatch(string: string, pattern: string | string[]): boolean {
  const patterns = Array.isArray(pattern) ? pattern : [pattern];
  for (const p of patterns) {
    if (
      new RegExp(
        p
          .split('.')
          .map(x => {
            if (x === '**') return '[a-zA-Z\\.]*';
            if (x.includes('**'))
              throw `Double star wildcards can only occur between dots as in request.**.expect`;
            return x.replace('*', '[a-zA-Z]*');
          })
          .join('\\.'),
      ).test(string)
    )
      return true;
  }
  return false;
}
```

Issue #482, opened 2019-11-23
rename `no-try-expect` to `no-conditional-expect` and add support for `if`


SimenB:
Sort of overlaps with `no-if`, but doesn't go as far as banning _all_ conditionals in tests - just assertions in them.

The case we're wanting to guard against with `no-try-expect` is `expect` not being run - I think this is the case for conditionals as well. There are other cases (like iterating through an empty array), but it's probably too hard to properly detect that.

@G-Rath thoughts on this?

G-Rath:
I think that makes sense, and serves as a good general rule for handling conditional expectations (surprise surprise) in whatever form they come in.

PR #515, opened 2020-01-12
feat(no-if): support `switch` statements

### LINKED ISSUE #505: [no-if] Request no switch cases ([no-switch] ?)
Switch statements should also be blocked since they are conditional logic. Docs of `no-if` state it blocks conditional statements, but doesn't cover this.

G-Rath:
@SimenB I think it might be worth refactoring this into `no-conditional-expect` at some point 🤔 

Looking over `no-try-expect`, I'm pretty sure they're doing almost the exact same thing - I'm going to look to reimplement one of these two rules so that they at least use the same style (one pushes booleans into an array, the other nodes, or something like that 🤷‍♂ )

Fixes #505

### DIFF
--- a/src/rules/__tests__/no-if.test.ts
+++ b/src/rules/__tests__/no-if.test.ts
@@ -9,7 +9,413 @@ const ruleTester = new TSESLint.RuleTester({
   },
 });
 
-ruleTester.run('no-if', rule, {
+ruleTester.run('conditional expressions', rule, {
+  valid: [
+    {
+      code: 'const x = y ? 1 : 0',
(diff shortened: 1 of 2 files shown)


SimenB:
> @SimenB I think it might be worth refactoring this into `no-conditional-expect` at some point 🤔

Yeah, I think I brought that up at some point

G-Rath:
> Yeah, I think I brought that up at some point

Yeah I think you did actually - I'll add it to my todo list for next weekend :)

SimenB:
#482

PR #517, opened 2020-01-12
chore(deps): removed dependency on micromatch for expect-expect

folke:
I've created a small utility function `nodeNameMatch` that can be used to match node names against a list of patterns.

Patterns can have single and double wildcards.

For example `request.get().bar().expectSomething(456)`
- `request.**.expectSomething` will pass
- `request.**.expect*` will pass
- `request.get.bar.expect*` will pass
- `request.*.expectSomething` will fail (you need to use `**` for multiple segments)

This removes the dependency on micromatch

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "@semantic-release/git": "^7.0.17",
     "@types/eslint": "^6.1.3",
     "@types/jest": "^24.0.15",
-    "@types/micromatch": "^4.0.0",
     "@types/node": "^12.6.6",
     "@typescript-eslint/eslint-plugin": "^2.5.0",
     "@typescript-eslint/parser": "^2.5.0",
--- a/package.json
+++ b/package.json
@@ -37,8 +37,7 @@
     "typecheck": "tsc -p ."
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^2.5.0",
-    "micromatch": "^4.0.2"
+    "@typescript-eslint/experimental-utils": "^2.5.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
--- a/src/rules/__tests__/expect-expect.test.ts
+++ b/src/rules/__tests__/expect-expect.test.ts
@@ -28,66 +28,32 @@ ruleTester.run('expect-expect', rule, {
       options: [{ assertFunctionNames: ['expectSaga'] }],
     },
     {
-      code: `test('verifies expect method call', () => {
-        class Foo {
-          expect(k) {
-            return k;
-          }
-        }
-        new Foo().expect(123);
-      });`,
+      code: `test('verifies expect method call', () => new Foo().expect(123));`,
       options: [{ assertFunctionNames: ['Foo.expect'] }],
     },
     {
-      code: `test('verifies deep expect method call', () => {
-              class Foo {
-                expect(k) {
-                  return k;
-                }
-              }
-              let tester = {
-                  foo: function() {
-                      return new Foo()
-                  }
-              }
-              tester.foo().expect(123);
-            });`,
+      code: `test('verifies deep expect method call', () => tester.foo().expect(123));`,
       options: [{ assertFunctionNames: ['tester.foo.expect'] }],
     },
     {
-      code: `test('wildcard chained function', () => {
-                    class Foo {
-                      expect(k) {
-                        return k;
-                      }
-                    }
-                    let tester = {
-                        foo: function() {
-                            return new Foo()
-                        }
-                    }
-                    tester.foo().expect(123);
-                  });`,
+      code: `test('wildcard chained function', () => tester.foo().expect(123));`,
       options: [{ assertFunctionNames: ['tester.*.expect'] }],
     },
     {
-      code: `test('verifies recursive expect method call', () => {
-                    class Foo {
-                      expect(k) {
-                        return this;
-                      }
-                      bar() {
-                          return this;
-                      }
-                    }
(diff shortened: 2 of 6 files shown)


src/rules/utils.ts
```diff
@@ -758,3 +758,32 @@ export const scopeHasLocalReference = (
...
+ *   request.*.expect
+ *   request.**.expect
+ *   request.**.expect*
+ */
+
+export function nodeNameMatch(
+  nodeName: string,
+  ...patterns: readonly string[]
+): boolean {
+  for (const p of patterns) {
```
G-Rath: I think this'd look cleaner if it was using functional array methods - specifically `#some`.

That would save you having to use the if condition, and (combined with an arrow function) explicit returns.

src/rules/__tests__/expect-expect.test.ts
```diff
@@ -28,67 +28,33 @@ ruleTester.run('expect-expect', rule, {
       options: [{ assertFunctionNames: ['expectSaga'] }],
```
G-Rath: Could you add a test for what happens if you have `*` & `**` at the end? i.e `request.*` & `request.**`.

Also having ones for them standalone might be useful too, just to display the behaviour (`*` & `**`).

(It might be worth creating a new run specifically for the wildcard matching behaviour; something like `ruleTester.run('wildcards', rule, { ... });`)

package.json
```diff
@@ -37,8 +37,7 @@
     "typecheck": "tsc -p ."
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^2.5.0",
-    "micromatch": "^4.0.2"
```
G-Rath: Don't forget to remove `@types/micromatch`, and commit the updated `yarn.lock` ;)

src/rules/expect-expect.ts
```diff
@@ -7,12 +7,12 @@ import {
   AST_NODE_TYPES,
   TSESTree,
 } from '@typescript-eslint/experimental-utils';
-import micromatch from 'micromatch';
 import {
   TestCaseName,
   createRule,
   getNodeName,
   getTestCallExpressionsFromDeclaredVariables,
+  nodeNameMatch,
```
G-Rath: I think it would better to have this in the rule rather than in `utils`, since it's only used here - we can easily move it out later if we find further use for it, but right now it's API is lightly coupled to `assertFunctionNames`.

Also, bit of a nit but I think something like `matchesAssertFunctionName` would be a better name, to give the impression it's a check.

src/rules/utils.ts
```diff
@@ -758,3 +758,32 @@ export const scopeHasLocalReference = (
...
 };
+
+/**
+ * Checks if node names returned by getNodeName matches any of the given star patterns
+ * Pattern examples:
+ *   request.*.expect
+ *   request.**.expect
+ *   request.**.expect*
+ */
+
```
G-Rath: 🐘 There doesn't need to be a newline here :)

src/rules/utils.ts
```diff
@@ -758,3 +758,32 @@ export const scopeHasLocalReference = (
...
+export function nodeNameMatch(
+  nodeName: string,
+  ...patterns: readonly string[]
```
G-Rath: 🐘 There's no real gain here to using rest, since I don't expect we'll be calling this in non-array conditions, and rest is requiring you to spread an array (effectively you're iterating over an array to extract elements out to be collected back into an array).

src/rules/utils.ts
```diff
@@ -758,3 +758,32 @@ export const scopeHasLocalReference = (
...
+): boolean {
+  for (const p of patterns) {
+    if (
+      new RegExp(
+        `^${p
+          .split('.')
+          .map(x => {
+            if (x === '**') return '[a-zA-Z\\.]*';
```
G-Rath: 🐘 You could use the `i` flag to let you use just `a-z`.

src/rules/utils.ts
```diff
@@ -758,3 +758,32 @@ export const scopeHasLocalReference = (
...
+ *   request.**.expect*
+ */
+
+export function nodeNameMatch(
+  nodeName: string,
+  ...patterns: readonly string[]
+): boolean {
+  for (const p of patterns) {
+    if (
+      new RegExp(
```
G-Rath: [This should use the `u` flag](https://eslint.org/docs/rules/require-unicode-regexp).

(I'll add this to our eslint config soon)

folke:
Thank you for reviewing!

For some reason, I always forget that `Array.some` exists :)

All requested changes have been implemented.

src/rules/__tests__/expect-expect.test.ts
folke: > Also having ones for them standalone might be useful too, just to display the behaviour (* & **).
Forgot this one. Making some additional changes right now

folke: Added the additional rules and also made sure that wildcard always match up until the next `.` or the end of the string

src/rules/expect-expect.ts
```diff
@@ -7,14 +7,38 @@ import {
...
 } from './utils';
 
+/**
+ * Checks if node names returned by getNodeName matches any of the given star patterns
+ * Pattern examples:
+ *   request.*.expect
+ *   request.**.expect
+ *   request.**.expect*
+ */
+function matchesAssertFunctionName(
```
G-Rath: 🐘 You could make this an arrow function, and short hand it.

❤️ the name!

PR #522, opened 2020-01-20
feat(valid-title): support `disallowedWords` option

### LINKED ISSUE #313: [new rule] no-vague-titles
At Shopify, we have a [rule that prevents specific words](https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/jest/no-vague-titles.md) from test titles. This is pretty nuanced/ specific, but I'd love to add it to this repo.

G-Rath:
Fixes #313

### DIFF
--- a/docs/rules/valid-title.md
+++ b/docs/rules/valid-title.md
@@ -87,7 +87,7 @@ describe(6, function() {});
 
 **duplicatePrefix**
 
-A describe/ test block should not start with duplicatePrefix
+A describe / test block should not start with duplicatePrefix
 
 Examples of **incorrect** code for this rule
 
@@ -117,7 +117,7 @@ describe('foo', () => {
 
 **accidentalSpace**
 
-A describe/ test block should not contain accidentalSpace
+A describe / test block should not contain accidentalSpace
 
 Examples of **incorrect** code for this rule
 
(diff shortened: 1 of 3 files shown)


docs/rules/valid-title.md
```diff
@@ -87,7 +87,7 @@ describe(6, function() {});
 
 **duplicatePrefix**
 
-A describe/ test block should not start with duplicatePrefix
+A describe / test block should not start with duplicatePrefix
```
SimenB:
```suggestion
A `describe` or `test` block should not start with `duplicatePrefix`.
```

same for others like it

docs/rules/valid-title.md
```diff
@@ -148,3 +148,45 @@ describe('foo', () => {
...
+
+When enabled, the type of the first argument to `describe` blocks won't be
+checked.
+
+#### `disallowedWords`
+
+Default: `[]`
+
+A string array of words that are not allowed to be used in test titles. Matching
+is not case-sensitive, and looks for complete words:
```
SimenB:
```suggestion
is not case sensitive, and looks for complete words:
```
?

docs/rules/valid-title.md
```diff
@@ -148,3 +148,45 @@ describe('foo', () => {
...
+
+Examples of **incorrect** code using `disallowedWords`:
+
+```js
+// with disallowedWords: ['correct', 'all', 'every', 'properly']
+describe('the correct way to do things', function() {});
```
SimenB: use arrow functions here as well to keep them similar except for what the rule enforces?

src/rules/__tests__/valid-title.test.ts
```diff
@@ -31,7 +118,7 @@ ruleTester.run('title-must-be-string', rule, {
     },
     {
       code: 'xdescribe(skipFunction, () => {});',
-      options: [{ ignoreTypeOfDescribeName: true }],
+      options: [{ ignoreTypeOfDescribeName: true, disallowedWords: [] }],
```
SimenB: we do we have to define it here if it's the default? type error?

src/rules/valid-title.ts
```diff
@@ -113,6 +124,20 @@ export default createRule({
           return;
         }
 
+        if (disallowedWords.length) {
```
SimenB:
```suggestion
        if (disallowedWords.length > 0) {
```

docs/rules/valid-title.md
G-Rath: Generally this [should be hyphenated](https://community.languagetool.org/rule/show/EN_COMPOUNDS?lang=en).

src/rules/__tests__/valid-title.test.ts
G-Rath: Yeah it seems that if we don't have it in at least one of the test cases that have `options`, TS gets snippy.

There's a couple of different ways around it, but they've all got pros & cons: primarily we could define an actual interface for the options, but that'd also require us to double type the message ids since we don't have [partial type arguments](https://github.com/microsoft/TypeScript/issues/26242), and then that'd require us to provide a default value in the actual function (as in, in the parameter in the `create` function).

This made TS happy, so I just decided to go with it :)

PR #524, opened 2020-01-23
feat(no-large-snapshots): add setting to define maxSize by snapshot type

### LINKED ISSUE #244: Add different setting for no-large-snapshots for inline vs external snaps
It would be cool to be able to specify the max length of inline snapshots separately from snapshots in files.

Main use case is setting inlineSnapshots to be something like 15 lines max, but a larger number like 50 for  regular snapshots.

leonardovillela:
Closes #244

### DIFF
--- a/src/rules/no-large-snapshots.ts
+++ b/src/rules/no-large-snapshots.ts
@@ -14,7 +14,6 @@ import {
 interface RuleOptions {
   maxSize?: number;
   inlineMaxSize?: number;
-  externalMaxSize?: number;
   whitelistedSnapshots?: Record<string, Array<string | RegExp>>;
 }
 
@@ -25,10 +24,7 @@ type RuleContext = TSESLint.RuleContext<MessageId, [RuleOptions]>;
 const reportOnViolation = (
   context: RuleContext,
   node: TSESTree.CallExpression | TSESTree.ExpressionStatement,
-  {
-    maxSize: lineLimit = 50,
-    whitelistedSnapshots = {},
-  }: Omit<RuleOptions, 'inlineMaxSize' | 'externalMaxSize'>,
+  { maxSize: lineLimit = 50, whitelistedSnapshots = {} }: RuleOptions,
 ) => {
   const startLine = node.loc.start.line;
   const endLine = node.loc.end.line;
@@ -76,14 +72,6 @@ const reportOnViolation = (
   }
 };
 
-const buildOptionsByMaxSizeAttributeName = (
-  options: RuleOptions,
-  attributeName: 'inlineMaxSize' | 'externalMaxSize',
-) => ({
-  ...options,
-  maxSize: options[attributeName] || options.maxSize,
-});
-
 export default createRule<[RuleOptions], MessageId>({
   name: __filename,
   meta: {
@@ -100,45 +88,22 @@ export default createRule<[RuleOptions], MessageId>({
     type: 'suggestion',
     schema: [
       {
-        anyOf: [
-          {
-            oneOf: [
-              {
-                type: 'object',
-                properties: {
-                  maxSize: {
-                    type: 'number',
-                  },
-                },
-                additionalProperties: false,
-              },
-              {
-                type: 'object',
-                properties: {
-                  externalMaxSize: {
-                    type: 'number',
-                  },
-                  inlineMaxSize: {
-                    type: 'number',
-                  },
-                },
-                additionalProperties: false,
-              },
-            ],
+        type: 'object',
+        properties: {
+          maxSize: {
+            type: 'number',
           },
-          {
+          inlineMaxSize: {
+            type: 'number',
+          },
+          whitelistedSnapshots: {
             type: 'object',
-            properties: {
-              whitelistedSnapshots: {
-                type: 'object',
-                patternProperties: {
-                  '.*': { type: 'array' },
-                },
-              },
+            patternProperties: {
+              '.*': { type: 'array' },
             },
-            additionalProperties: false,
           },
-        ],
+        },
+        additionalProperties: false,
       },
     ],
   },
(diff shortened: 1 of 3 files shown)


src/rules/no-large-snapshots.ts
```diff
@@ -23,7 +25,10 @@ type RuleContext = TSESLint.RuleContext<MessageId, [RuleOptions]>;
 const reportOnViolation = (
   context: RuleContext,
   node: TSESTree.CallExpression | TSESTree.ExpressionStatement,
-  { maxSize: lineLimit = 50, whitelistedSnapshots = {} }: RuleOptions,
+  {
```
G-Rath: I think it'll be best to break this into two arguments: that'll work better with my above comment about using `maxSize` & `inlineMaxSize`, since you'll be able to just do `options.maxSize` & `options.inlineMaxSize ?? options.maxSize` depending on the branch :)

src/rules/no-large-snapshots.ts
```diff
@@ -13,6 +13,8 @@ import {
 
 interface RuleOptions {
   maxSize?: number;
+  inlineMaxSize?: number;
+  externalMaxSize?: number;
```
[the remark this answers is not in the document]
leonardovillela: Hi @G-Rath, sorry for the delay to answer these comments. 

In this PR I also updated the schema definition for this rule, so it's impossible to have a definition like above because the JSON schema validation will throw an error. I understand your point, but even the docs of Jest refer to this type of snapshot as ["external"](https://jestjs.io/docs/en/snapshot-testing#inline-snapshots):

> "Inline snapshots behave identically to external snapshots (.snap files), except the snapshot values are written automatically back into the source code."

In this way, I think it is more explicit than it is being configured. However, if you still think the refactor is necessary I can do it.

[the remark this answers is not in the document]
G-Rath: I did see the schema update, but that's also why I think it's worth going down this route: it's overall less complex, and still makes logical sense.

Here's my logic:
  * Dev has rule configured with `maxSize` set to `x`
  * Dev decides that they want inline snapshots to have max size of `y`

At this point, currently the flow is like this:
  * Dev adds `inlineMaxSize: y`
  * ESLint errors complaining about invalid configuration
  * Dev renames `maxSize` to `externalMaxSize`

By just reusing `maxSize` it would become:
  * Dev adds `inlineMaxSize: y`

-------

Thinking ahead a few moves, I think you're right down the line, as this feature actually raises another question: what if I only want to limit the size of *one* of the snapshot types?

Currently, if a size is not provided, we fallback to using `maxSize`, which has a default value of 50, meaning enabling this result requires you setting a size on both snapshot types, where as someone might want to only limit the size of inline snapshots for example.

*However* I think this is a distinctive enough use-case that we can ignore it for this PR, and implement it later if requested - generally I've found it's better to start with tighter rules and loosen over time by adding options as they're requested, rather than trying to prematurely guess what people want :)

src/rules/no-large-snapshots.ts
G-Rath: @leonardovillela this is definitely a refactor we should do, as it cuts out the need for `buildOptionsByMaxSizeAttributeName` entirely :)

src/rules/no-large-snapshots.ts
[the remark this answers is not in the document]
leonardovillela: Okay, I'll refactor and remove `externalMaxSize`.

leonardovillela:
@G-Rath changes made, sorry for the delay.

src/rules/no-large-snapshots.ts
```diff
@@ -23,7 +24,10 @@ type RuleContext = TSESLint.RuleContext<MessageId, [RuleOptions]>;
 const reportOnViolation = (
   context: RuleContext,
   node: TSESTree.CallExpression | TSESTree.ExpressionStatement,
-  { maxSize: lineLimit = 50, whitelistedSnapshots = {} }: RuleOptions,
+  {
+    maxSize: lineLimit = 50,
+    whitelistedSnapshots = {},
+  }: Omit<RuleOptions, 'inlineMaxSize'>,
```
G-Rath:
```suggestion
  }: RuleOptions,
```

There's no need to omit this property, since the options we pass in still match the type.
We can always add `Omit` later if we need to :)

src/rules/no-large-snapshots.ts
```diff
@@ -125,7 +132,10 @@ export default createRule<[RuleOptions], MessageId>({
                 'toThrowErrorMatchingInlineSnapshot',
               ))
           ) {
-            reportOnViolation(context, node, options);
+            reportOnViolation(context, node, {
+              ...options,
+              maxSize: options.inlineMaxSize,
```
G-Rath: These something not quite right here: this will mean that if I don't pass `inlineMaxSize`, it'll use the default of `50` rather than `maxSize`.

So this should be:

```suggestion
              maxSize: options.inlineMaxSize ?? options.maxSize,
```

*However*, it seems that the tests don't cover this. Could you look into that for me, and see what's up?
I suspect that committing this will make coverage reveal the lack of ... well, coverage 😅 

(I suspect we're missing a test where we don't have `inlineMaxSize`, `maxSize` set greater than 50, and an inline snapshot that's greater than 50)

src/rules/no-large-snapshots.ts
leonardovillela: Yes, I added omit because `inlineMaxSize` is not used in this function.

G-Rath: True, but for an internal function like this I think it's fine - helps keep things a bit simpler and require less thinking.

I'd totally agree with you if it was an exported function, like in `utils` :)

src/rules/no-large-snapshots.ts
leonardovillela: Done, I have added a test to this scenario.

src/rules/no-large-snapshots.ts
leonardovillela: Ok, I removed the omit.

leonardovillela:
You're welcome. Thanks for the patient in the review. I'm excited to contribute more in jest related projects.

G-Rath:
> Thanks for the patient in the review

Thanks for your patiences in being reviewed :)

> I'm excited to contribute more in jest related projects.

For sure!

> if you have any other recommended issues, let me know.

Will do; this package is pretty stable & feature-complete right now (most of the open issues are discussion, or waiting replies), but if anything comes up I'll keep you in mind.

If you want to go looking through `jest` itself & other related projects, feel free to @ me if you ever want a second pair of eyes or opinion on something :)

PR #542, opened 2020-03-06
fix(prefer-to-contain): check that expect argument is defined before use

G-Rath:
🍦


G-Rath:
@SimenB I had to regenerate the lockfile because the tests were failing due to https://github.com/nodejs/node/issues/32107, and then manually revert `eslint-plugin-import` back in the lock due to https://github.com/benmosher/eslint-plugin-import/issues/1655 which meant the Windows tests would fail 😬 

Just a heads up in case you run into this elsewhere (it's already happened in `eslint-plugin-jest-extended`).

PR #545, opened 2020-03-20
feat(rules): add no-expect-to-match-object rule

viestat:
### Motivation

Jest allows you to to check that a JavaScript object matches a **subset** of the properties of an object using `expect().toMatchObject(object)`.

Checking just a subset of an object might lead to not catching changes in an object in the tests, this rule bans `expect().toMatchObject` in favor of `expect().toStrictEqual` for a **deep** equality check or `expect().objectContaining` to consciously check the properties that are interesting.


src/rules/utils.ts
```diff
@@ -339,6 +339,7 @@ export enum ModifierName {
 export enum EqualityMatcher {
   toBe = 'toBe',
   toEqual = 'toEqual',
+  toMatchObject = 'toMatchObject',
```
G-Rath, line 342: This isn't an equality matcher.

i.e `prefer-to-contain` will error on `expect(a.includes(b)).toMatchObject(true)`.

Instead just doing `matcher.name === 'toMatchObject'` should be enough.
(That'll actually improve things anyway, as the rule currently won't catch `toMatchObject` unless it's called & has only one argument).

src/rules/__tests__/no-expect-to-match-object.test.ts
```diff
@@ -0,0 +1,23 @@
...
+import rule from '../no-expect-to-match-object';
+
+const ruleTester = new TSESLint.RuleTester();
+
+ruleTester.run('no-expect-to-match-object', rule, {
+  valid: [
+    'expect({ a: 1, b: 2 }).toStrictEqual({ a: 1, b: 2 });',
+    'expect({ a: 1, b: 2 }).objectContaining({ a: expect.any(Number), b: expect.any(Number) });',
+  ],
+  invalid: [
```
G-Rath, line 11: Could you chuck in some invalid cases for when:

* `toMatchObject` is expected but not actually called
* `toMatchObject` is expected with no arguments
* `toMatchObject` is expected with multiple arguments
* expect is called without a value
* expect is called with a non-object value
* a few of these tests with modifiers

src/rules/__tests__/no-expect-to-match-object.test.ts
```diff
@@ -0,0 +1,23 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+import rule from '../no-expect-to-match-object';
+
+const ruleTester = new TSESLint.RuleTester();
+
+ruleTester.run('no-expect-to-match-object', rule, {
+  valid: [
```
G-Rath, line 7: Could you chuck in some valid cases for when:

* expect is called without any matcher (i.e `expect(a);`)
* expect is called with a modifier
* expect is called with a modifier + matcher
* `toMatchObject` is called by itself, as if it's a standalone function (i.e `'toMatchObject(...)'`)

docs/rules/no-expect-to-match-object.md
```diff
@@ -0,0 +1,40 @@
...
+
+The following patterns are not considered warning:
+
+```js
+test('some test', () => {
+  expect({ a: 1, b: 2 }).toStrictEqual({ a: 1, b: 2 });
+});
+
+test('some test', () => {
+  expect({ a: 1, b: 2 }).objectContaining({
```
G-Rath, line 35: `objectContaining` is not a matcher, so cannot be used by this.

Could you update the documentation & code to reflect the correct way of using `objectContaining` (including the linting error message).

src/rules/utils.ts
viestat: Removed in https://github.com/jest-community/eslint-plugin-jest/pull/545/commits/0f8ada9303b31b3e7936595529614921d90c64f4

src/rules/__tests__/no-expect-to-match-object.test.ts
viestat: Added tests in https://github.com/jest-community/eslint-plugin-jest/pull/545/commits/0f8ada9303b31b3e7936595529614921d90c64f4

src/rules/__tests__/no-expect-to-match-object.test.ts
viestat: added tests in https://github.com/jest-community/eslint-plugin-jest/pull/545/commits/0f8ada9303b31b3e7936595529614921d90c64f4

docs/rules/no-expect-to-match-object.md
viestat: fixed with https://github.com/jest-community/eslint-plugin-jest/pull/545/commits/0f8ada9303b31b3e7936595529614921d90c64f4

viestat:
> Thanks for this - sorry for the delay in reviewing!
> 
> I've got a few changes to request; I've also asked if you could chuck in a bunch more tests, just because they're really cheap - some of them might seem odd, but they're just-in-case to catch weird edge cases we've hit in the past with other rules.
>


Thank you for your review 😄 , I added your suggestions.
> Also looking over the rule, if the behaviour is positive, I'm wondering if it might be better to focus on `objectContaining`; so call the rule `prefer-object-containing`, focus on saying "prefer expect.objectContaining" (i.e because you can use it with `toStrictEqual`, but also with other matchers), and maybe even implement an autofix that replaces `toMatchObject` -> `toStrictEqual(expect.objectContaining)`?
> 
> That last one might be going a bit too far for now, but let me know what you think :)


I am happy to adapt this rule in any way. The motivation for me was more on the safety of a nested equality check. I believe an auto fix would be making too many assumptions since the options are to either explicitly state what you are expecting the object to be like:
```js
expect(obj).toStrictEqual({ a: 1, b: 2 });
```
or to check specific properties but still be aware of other properties: 
```js
expect(obj).toStrictEqual(
  expect.objectContaining({
    a: 1,
    b: expect.any(Number),
  }),
);
```

src/rules/no-expect-to-match-object.ts
```diff
@@ -0,0 +1,36 @@
...
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Disallow expect.toMatchObject',
+      recommended: false,
+    },
+    messages: {
+      useToStrictEqual: 'Use toStrictEqual() for a deep equality check',
```
SimenB, line 12: should this include something about `expect.objectContaining` as well?

G-Rath: Yeah I'm thinking maybe this rule should be `prefer-object-containing` and push that, rather than pushing for `toStrictEqual` since that's covered by `prefer-strict-equal` where as you can use `expect.objectContaining` with other matches.

(I can't remember off the top of my head if `toBe` with `expect.objectContaining` behaviours any similar to `toStrictEqual`)

SimenB: `toBe` wouldn't make sense as that's for referential equality

viestat: > should this include something about `expect.objectContaining` as well?
> @SimenB 

This was like this in the first commit of this PR: https://github.com/jest-community/eslint-plugin-jest/blob/5ecba757567d02f0a618fb6c12fbba7074d1b79e/src/rules/no-expect-to-match-object.ts#L17-L20
Should it be something like it then? , how about ?:

`Use toStrictEqual() for a deep equality check, use expect.objectContaining() to check specific properties`

viestat: > Yeah I'm thinking maybe this rule should be `prefer-object-containing` and push that, rather than pushing for `toStrictEqual` since that's covered by `prefer-strict-equal` where as you can use `expect.objectContaining` with other matches.
> @G-Rath 

`prefer-strict-equal` does not prevent the use of `expect().toMatchObject(object)`

G-Rath: Correct - my point was about that you can use `expect.objectContaining` with matchers other than `toStrictEqual` (`toEqual` for example, which would have the same result in most cases).

Looking back over this, I think maybe this is just another one for a general "no-banned-matches" rule since the matcher does have uses, and doesn't have any particular conditions around it like `prefer-to-contain`, `prefer-to-have-length`, etc... 🤔 

If I get the time this weekend I might try implementing such a rule.

PR #553, opened 2020-04-02
feat: create `no-interpolation-in-snapshots` rule

viestat:
### Motivation
This rule disallows the use of string interpolations within `toMatchInlineSnapshot` since interpolation prevents snapshots from being updated. 

Instead, properties should be overloaded with a matcher by passing the optional `propertyMatchers` argument to `toMatchInlineSnapshot`.

```js

expect(something).toMatchInlineSnapshot(
  `Object {
    property: ${interpolated}
  }`
);

expect(something).toMatchInlineSnapshot({ other: expect.any(Number) }, 
  `Object {
    other: Any<Number>,
    property: ${interpolated}
  }`
);
```

```js

expect(something).toMatchInlineSnapshot();

expect(something).toMatchInlineSnapshot(
  `Object {
    property: 1
  }`
);

expect(something).toMatchInlineSnapshot({ property: expect.any(Date) }, 
  `Object {
    property: Any<Date>
  }`
);

```


src/rules/__tests__/no-interpolation-inline-snapshot.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: resolveFrom(require.resolve('eslint'), 'espree'),
+  parserOptions: {
+    ecmaVersion: 2017,
+  },
+});
+
+ruleTester.run('no-interpolation-inline-snapshot', rule, {
+  valid: [
```
G-Rath, line 13: Could you add some more tests for cases like:

* when expect is called without a matcher or modifier
* when expect is called with a dangling modifier (i.e no matcher)
* when a property is accessed on `expect` (i.e `expect.toHaveAssertions()`
* when a method called `toMatchInlineSnapshot` is called with interpolation, both by itself and on an object that looks suspiciously similar to `expect`

src/rules/no-interpolation-inline-snapshot.ts
```diff
@@ -0,0 +1,48 @@
...
+    schema: [],
+    type: 'suggestion',
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      CallExpression(node) {
+        const { callee, arguments: nodeArguments } = node;
+
+        if (
```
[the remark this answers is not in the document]
viestat: fixed in https://github.com/jest-community/eslint-plugin-jest/pull/553/commits/b6c50ec9a1ae5cdea27c0e6af9002e0c421d15b7

src/rules/__tests__/no-interpolation-inline-snapshot.test.ts
viestat: added in https://github.com/jest-community/eslint-plugin-jest/pull/553/commits/b6c50ec9a1ae5cdea27c0e6af9002e0c421d15b7

viestat:
Thanks for the feedback!

docs/rules/no-interpolation-inline-snapshot.md
```diff
@@ -0,0 +1,50 @@
+# No string interpolation inside inline snapshots (no-interpolation-inline-snapshot)
```
G-Rath, line 1:
```suggestion
# Disallow string interpolation inside inline snapshots (`no-interpolation-inline-snapshot`)
```

src/rules/no-interpolation-inline-snapshot.ts
```diff
@@ -0,0 +1,51 @@
...
+      CallExpression(node) {
+        if (!isExpectCall(node)) {
+          return;
+        }
+
+        const { matcher } = parseExpectCall(node);
+
+        if (
+          matcher?.name !== 'toMatchInlineSnapshot' ||
+          matcher?.arguments === null
```
G-Rath, line 31: I think you might as well just use optional chaining on the `forEach` below, since `matcher?.name !==` will also check that `matcher` is defined

viestat: I am not sure I understand what you mean here.

G-Rath: Optional chaining is cool, but if you use it multiple times in a row on the same object, that's usually a smell that you should just check for existence.

When using optional chaining in a condition, it has the same guarding as doing "obj && obj.prop <condition>", meaning it does a "is defined" check.

For example this:

```
if(node.parent && node.parent.type === AST_NODE_TYPES.AssignmentExpression) {
  const { right } = node.parent;

  ...
}
```

Can be simplified to just:

```
if(node.parent?.type === AST_NODE_TYPES.AssignmentExpression) {
  const { right } = node.parent;

  ...
}
```

----------

So originally what I was saying when your code was just `matcher?.name === 'value'` was that you could drop the `matcher.arguments === null` in favor of just later doing `matcher.arguments?.forEach`.

However, now that the code has changed to include multiple compares to name, for which I'd recommend using `include`, and so in turn just do a `matcher &&` as `matcher?.name` returns type `string | undefined` which is not accepted by `string[].includes` (as that expects just `string`).

----------

tl;dr

* I would replace the multiple `matcher?.name ===`s with just `matcher && ['toMatchInlineSnapshot', 'toThrowErrorMatchingInlineSnapshot'].includes(matcher.name)`
* I would replace `matcher.arguments === null` with `matcher.arguments?.forEach`

src/rules/__tests__/no-interpolation-inline-snapshot.test.ts
G-Rath: Awesome, thanks!

If you could check one more in that has two or more interpolations, just for completeness, that would be cool but not blocking.

CHANGELOG.md
```diff
@@ -1,96 +1,111 @@
 ## [23.17.1](https://github.com/jest-community/eslint-plugin-jest/compare/v23.17.0...v23.17.1) (2020-06-23)
```
G-Rath, line 1: This file is automatically generated, so should not be formatted with prettier as it'll get overridden the next time a release happens :)

docs/rules/no-interpolation-inline-snapshot.md
```diff
@@ -0,0 +1,50 @@
+# Disallow string interpolation inside inline snapshots (`no-interpolation-inline-snapshot`)
+
+Prevents the use of string interpolations within `toMatchInlineSnapshot`.
+
+## Rule Details
+
+Interpolation prevents snapshots from being updated. Instead, properties should
+be overloaded with a matcher by passing the optional `propertyMatchers` argument
+to `toMatchInlineSnapshot`.
```
G-Rath, line 9:
```suggestion
Interpolation prevents snapshots from being updated. Instead, properties should
be overloaded with a matcher by using [property matchers](https://jestjs.io/docs/en/snapshot-testing#property-matchers).
```

docs/rules/no-interpolation-inline-snapshot.md
```diff
@@ -0,0 +1,50 @@
...
+
+Examples of **incorrect** code for this rule:
```
G-Rath, line 11: Could you chuck a in/correct example for `toThrowErrorMatchingInlineSnapshot` as well?

viestat: added in https://github.com/jest-community/eslint-plugin-jest/pull/553/commits/a405636f7128de898187ac9dd3b54e90d65529b2

src/rules/no-interpolation-inline-snapshot.ts
viestat: changed in https://github.com/jest-community/eslint-plugin-jest/pull/553/commits/96eaf9b56e404c1c6f835792232067fdc7af9151

src/rules/__tests__/no-interpolation-inline-snapshot.test.ts
viestat: added in 96eaf9b

CHANGELOG.md
viestat: This was edited by husky commit tasks when I merged the latest changes with master 😓 
Removed them with https://github.com/jest-community/eslint-plugin-jest/pull/553/commits/ab2c365896fe7e61daa8dbd2fd4ccdae29ac5046

src/rules/no-interpolation-in-snapshots.ts
```diff
@@ -0,0 +1,52 @@
+import { AST_NODE_TYPES } from '@typescript-eslint/experimental-utils';
+import { createRule, isExpectCall, parseExpectCall } from './utils';
+
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Disallow string interpolation inside inline snapshots.',
```
SimenB, line 9:
```suggestion
      description: 'Disallow string interpolation inside inline snapshots',
```

to match the others

src/rules/no-interpolation-in-snapshots.ts
```diff
@@ -0,0 +1,52 @@
...
+      recommended: false,
+    },
+    messages: {
+      noInterpolation:
+        'Do not use string interpolation inside of inline snapshots',
+    },
+    schema: [],
+    type: 'suggestion',
```
SimenB, line 17:
```suggestion
    type: 'problem',
```

src/rules/no-interpolation-in-snapshots.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        const { matcher } = parseExpectCall(node);
+
+        if (
+          matcher &&
+          [
+            'toMatchInlineSnapshot',
+            'toThrowErrorMatchingInlineSnapshot',
+          ].includes(matcher.name)
+        ) {
```
SimenB, line 35:
```suggestion
        const { matcher } = parseExpectCall(node);

        if (!matcher) {
          return;
        }

        if (
          [
            'toMatchInlineSnapshot',
            'toThrowErrorMatchingInlineSnapshot',
          ].includes(matcher.name)
        ) {
```

docs/rules/no-interpolation-in-snapshots.md
```diff
@@ -0,0 +1,60 @@
+# Disallow string interpolation inside inline snapshots (`no-interpolation-in-snapshots`)
```
G-Rath, line 1:
```suggestion
# Disallow string interpolation inside snapshots (`no-interpolation-in-snapshots`)
```

Might make sense to drop the `inline` given the name change, just to keep things short and snappy 🤔

Issue #222, opened 2019-02-07
New rule: no istanbul ignore without reason


eranshabi:
When using the `coverageThreshold` option of jest you sometimes you have to add `/* istanbul ignore next */` comments in order for your coverage to much the threshold.
Often people will add this comment with no reason of why the code is not covered (examples: covered in e2e, can't be covered, will be covered later...).
I think the reason why it's uncovered should be documented so other developers will know how to handle this code in the future, and that's why this rule would be useful.

What do you think? Is it related enough to jest to be here?

SimenB:
Good idea!

>What do you think? Is it related enough to jest to be here?

I'd think so - quick googling didn't yield any istanbul plugin.

/cc @bcoe thoughts?

antonk52:
Are there any more thoughts on this topic? This looks like a healthy feature to go along with a linter rule to enforce it

SimenB:
Agreed, would be nice to get this in. Wanna provide a PR? 😀 Ideally this would be in some sort of `eslint-plugin-istanbul`, but as far as I can tell that doesn't exist. Istanbul backs all of JS code coverage (via `nyc` or `c8` (as well as Jest, of course)), and such a rule seems useful for all

antonk52:
I am happy to put a PR together, any advice on the rule setting, I have in mind two rules to implement:
- `prefer-istanbul-ignore-description`
  essentially errors/warns on any "Istanbul ignore" comment without a specified comment.
- `no-istanbul-ignore-file`
  errors for any mention of `Istanbul ignore file` comment. 

What do you think? Should there be any custom settings for the proposed rules?

SimenB:
Those sounds good to me! No options come to mind 🙂

SimenB:
We should link to `eslint-plugin-istanbul` from our README

PR #558, opened 2020-04-29
istanbul rules

antonk52:
This PR is an extension of the discussion from #222 

It introduces two new rules
- no-istanbul-ignore-file
- prefer-instanbul-ignore-reason


src/rules/prefer-istanbul-ignore-reason.ts
```diff
@@ -0,0 +1,53 @@
...
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description:
+        'This rule raises a warning about "istanbul ignore" comments missing a reason.',
+      recommended: false,
+    },
+    messages: {
+      noReason: 'Add a reason why code should be ignored',
```
SimenB, line 22:
```suggestion
      noReason: 'Add a reason why code coverage should be ignored',
```

src/rules/no-istanbul-ignore-file.ts
```diff
@@ -0,0 +1,39 @@
...
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description:
+        'This rule raises a warning about "istanbul ignore file" comments.',
+      recommended: false,
+    },
+    messages: {
+      istanbulIgnoreFile: 'Ignoring whole files is not allowed',
```
SimenB, line 14:
```suggestion
      istanbulIgnoreFile: 'Ignoring coverage for whole files is not allowed',
```

src/rules/prefer-istanbul-ignore-reason.ts
```diff
@@ -0,0 +1,53 @@
+import { TSESTree } from '@typescript-eslint/experimental-utils';
+import { createRule } from './utils';
+
+function hasIgnore(comment: string): boolean {
+  return /istanbul ignore \w+/u.test(comment);
```
G-Rath, line 5:
```suggestion
  return /^istanbul ignore \w+/u.test(comment);
```

src/rules/prefer-istanbul-ignore-reason.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function hasReason(comment: string): boolean {
+  return /istanbul ignore \w+: .*/u.test(comment);
```
G-Rath, line 9:
```suggestion
  return /^istanbul ignore \w+: .*/u.test(comment);
```

src/rules/prefer-istanbul-ignore-reason.ts
```diff
@@ -0,0 +1,53 @@
...
+    const sourceCode = context.getSourceCode();
+
+    function checkNode(node: TSESTree.Comment) {
+      const trimmedComment = node.value.trim();
+
+      if (!hasIgnore(trimmedComment)) {
+        return;
+      }
+
+      if (hasReason(trimmedComment)) {
```
G-Rath, line 38: nit: this could be merged with the previous `if`

docs/rules/prefer-istanbul-ignore-reason.md
```diff
@@ -0,0 +1,27 @@
+# Disallow disabling coverage for whole files (no-istanbul-ignore-file)
+
+This rule raises a warning about "istanbul ignore" comments missing a reason why
+code is ignored.
+
+## Rule Details
+
+The rule finds all comments including patter `istanbul ignore <word>` and raises
```
G-Rath, line 8: This doesn't read right to me - I assume it's meant to be:

```suggestion
The rule finds all comments matching the pattern `istanbul ignore <word>` and raises
```

(`including` wasn't part of the typo, but I feel `matching` sounds more organic in this sentence)

src/rules/prefer-istanbul-ignore-reason.ts
G-Rath: Istanbul ignores are not required to have a colon; you can use any non-word char including whitespace: `/* istanbul ignore next as it's only required to please typescript */` & `// istanbul ignore if - we've not yet got a nice way to test this branch`.

G-Rath:
I updated my suggested regexs with the [actual regexs](https://github.com/istanbuljs/istanbuljs/blob/a99a13ee6931fc124a2a723c3f511cdbcb0aa81d/packages/istanbul-lib-instrument/src/visitor.js#L7-L12) that istanbul uses, which is why they've seemed to have gotten overly complex.

G-Rath:
I updated my suggested regexs with the [actual regexs](https://github.com/istanbuljs/istanbuljs/blob/a99a13ee6931fc124a2a723c3f511cdbcb0aa81d/packages/istanbul-lib-instrument/src/visitor.js#L7-L12) that istanbul uses, which is why they've seemed to have gotten overly complex.

Turns out you can have as much whitespace as you'd like in the comments, so should probably chuck in a test or two for that as well.

src/rules/prefer-istanbul-ignore-reason.ts
antonk52: I found an example with a colon in Istanbul docs and considered it to be a good practice from their side. You're right. This rule should not enforce this.

antonk52:
Thanks @G-Rath, I've applied the proposed changes and it looks much better now. As for the auto formatting from `ignore file` to `ignore next` I have consider it prior writing the rule and I really wish for it to not exist. The reason being making it too easy to still ignore everything in a file. If a person wants to ignore something, that the team agreed to cover, this person should not be allowed to take shortcuts in this regard. I hope this clears the intent for this rule. 

Thanks for the review =)

G-Rath:
> As for the auto formatting from ignore file to ignore next

I think you misunderstood - I was referring to a common mistake I make when using istanbul ignores; I use them so rarely that I always have to look up the documentation for how to do them, as they don't follow the pattern used by tools such as typescript (`ts-ignore`), eslint (`eslint-ignore-next`), stylelint (`stylelint-ignore-next`), etc.

So usually the follow is something like this:

1. write an if condition to please typescript
2. write `/* istanbul-ignore-next */`
3. sigh as the line isn't ignored, and google "istanbul ignore"
4. write `/* istanbul ignore next */`

So my half-joke was about having a rule that checks if you're doing `-` instead of ` `, which could easily be autofixed. It's an easy, but I suspect it's not a big enough problem to be worth it :joy:

antonk52:
Sorry, it's late over here and my poor brain didn't get that right😅
Thanks for expanding on it 👍

SimenB:
> I updated my suggested regexs with the [actual regexs](https://github.com/istanbuljs/istanbuljs/blob/a99a13ee6931fc124a2a723c3f511cdbcb0aa81d/packages/istanbul-lib-instrument/src/visitor.js#L7-L12) that istanbul uses, which is why they've seemed to have gotten overly complex.

@correyfarrel would love your thoughts on this, if any 🙂

G-Rath:
@SimenB you missed an 'l'; @coreyfarrell 🙂

SimenB:
Hah, I guess that's my cue to call it a night 😅

docs/rules/prefer-istanbul-ignore-reason.md
```diff
@@ -0,0 +1,27 @@
+# Disallow disabling coverage for whole files (no-istanbul-ignore-file)
```
coreyfarrell, line 1: Wrong heading here.

docs/rules/prefer-istanbul-ignore-reason.md
```diff
@@ -0,0 +1,27 @@
...
+
+The rule finds all comments matching the pattern `istanbul ignore <word>` and
+raises a warning if it is missing a reason afterwards.
+
+The following patterns are considered warnings:
+
+```js
+/* istanbul ignore file */
+
+// istanbul ignore file
```
coreyfarrell, line 16: Should list ignore next/if/else here as well.

I would expect that this rule not block `istanbul ignore file: vendored from https://github.com/blah/blah` but `no-istanbul-ignore-file` would block that?

docs/rules/no-istanbul-ignore-file.md
```diff
@@ -0,0 +1,30 @@
+# Disallow disabling coverage for whole files (no-istanbul-ignore-file)
```
coreyfarrell, line 1: Should the documentation suggest manipulation of the `include` or `exclude` configuration options?  Maybe in a follow-up?

SimenB:
@coreyfarrell wonderful, let's do it! We'll add a link to it from this repo's readme

coreyfarrell:
I just quickly created https://github.com/istanbuljs/eslint-plugin-istanbul and pushed a stub package.json to it to have something for a PR to be based on.

SimenB:
Cool. Do you want some help to scaffold it out? I'm assuming you don't want TS there, but copying e.g. https://github.com/sindresorhus/eslint-plugin-unicorn and ripping out all the rules and setting up the linting/testing you want should be a good start 🙂

coreyfarrell:
I'd love any help I can get!  I do prefer standard JS without any 'build' step (original sources published to npm).  eslint-plugin-unicorn would probably be a reasonable base for this.  One thing for now it would probably be nice to still support node.js 8?  If we use `xo` for self-linting I think we could use `"pretest": "if-ver -ge 10 || exit; xo"` that way we skip the lint step for node.js 8 but still test that the plugin runs on node.js 8.  I suspect support for node.js 8 will be short lived but it seems like having a release that supports the same engines as jest and nyc would be best.

SimenB:
I'd just copy linting etc from the monorepo, or nyc, or something

G-Rath:
While I personally prefer TypeScript for the IDE support & type safety, I'm still happy to help with this new plugin :)

coreyfarrell:
@SimenB I've been using `xo` in new istanbuljs repos (schema and load-nyc-config for example).  Currently the new istanbuljs repos are using xo `^0.25.3` since it's the last to support node.js 8 but in `append-transform` for example I updated xo and used the `if-ver` conditional execution.

SimenB:
ah ok, cool 👍

antonk52:
guess it makes sense to avoid any further work in this PR and aim these rules at https://github.com/istanbuljs/eslint-plugin-istanbul. I am happy to help with the repo set up as well 👍

SimenB:
yes please @antonk52! I've gotten distracted by memory leaks unfortunately, so I won't get to this until tomorrow

SimenB:
@antonk52 `eslint-plugin-istanbul` has been released now, so I think we can close this. Thanks for all your help here!

PR #564, opened 2020-05-05
feat(no-deprecated-functions): support jest `version` setting

### LINKED ISSUE #561: [no-deprecated-functions] support jest version setting
We should support taking a `jestVersion` setting that rules like `no-deprecated-functions` can use to only warn about deprecations that apply to the version of jest being used.

G-Rath:
This is pretty cool, if I do say so myself.

For now while it's implemented all as util functions, I've lumped them in `no-deprecated-functions` as `utils` is getting pretty big & I'd rather have this reviewed first since it's easy to move them out.

I've tested this locally against a few projects, including `@typescript-eslint`, and nothing went wrong.

fixes #561

### DIFF
--- a/docs/rules/no-deprecated-functions.md
+++ b/docs/rules/no-deprecated-functions.md
@@ -9,9 +9,19 @@ of majors, eventually they are removed completely.
 ## Rule details
 
 This rule warns about calls to deprecated functions, and provides details on
-what to replace them with.
+what to replace them with, based on the version of Jest that is installed.
 
 This rule can also autofix a number of these deprecations for you.
+
+### `jest.resetModuleRegistry`
+
+This function was renamed to `resetModules` in Jest 15, and is scheduled for
+removal in Jest 27.
+
+### `jest.addMatchers`
+
+This function was replaced with `expect.extend` in Jest 17, and is scheduled for
+removal in Jest 27.
 
 ### `require.requireActual` & `require.requireMock`
 
@@ -24,16 +34,6 @@ These functions were later moved onto the `jest` object in order to be easier
 for type checkers to handle, and their use via `require` deprecated. Finally,
 the release of Jest 26 saw them removed from the `require` function all
 together.
-
-### `jest.addMatchers`
-
-This function was replaced with `expect.extend` in Jest 17, and is scheduled for
-removal in Jest 27.
-
-### `jest.resetModuleRegistry`
-
-This function was renamed to `resetModules` in Jest 15, and is scheduled for
-removal in Jest 27.
 
 ### `jest.runTimersToTime`
 
--- a/src/rules/__tests__/no-deprecated-functions.test.ts
+++ b/src/rules/__tests__/no-deprecated-functions.test.ts
@@ -3,7 +3,10 @@ import * as os from 'os';
 import * as path from 'path';
 import { JSONSchemaForNPMPackageJsonFiles } from '@schemastore/package';
 import { TSESLint } from '@typescript-eslint/experimental-utils';
-import rule, { JestVersion } from '../no-deprecated-functions';
+import rule, {
+  JestVersion,
+  _clearCachedJestVersion,
+} from '../no-deprecated-functions';
 
 const ruleTester = new TSESLint.RuleTester();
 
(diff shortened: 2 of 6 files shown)


G-Rath:
@SimenB I think I might break `utils` up into a set of grouped ts files, barrelled in `utils/`; something like:

```
utils/index.ts
utils/misc.ts
utils/getNodeName.ts
utils/parseExpectCall.ts
utils/followTypeAssertionChain.ts
...etc...
```

src/rules/__tests__/no-deprecated-functions.test.ts
```diff
@@ -1,49 +1,230 @@
+import * as fs from 'fs';
```
G-Rath, line 1: All the `fs` function calls are `await`d on as they're all setup to be drop-in replaced with `{ promises as fs }` when we drop node 7 support.

SimenB: that'll print an experimental warning I think? Maybe not, and if so we can do `util.promisify` manually. Either way, it doesn't matter much in a test I guess

(as a side note, in order to support ESM node has added `import {readFile} from 'fs/promises'` in node 14)

src/rules/no-deprecated-functions.ts
```diff
@@ -1,9 +1,56 @@
...
+      paths: [process.cwd()],
+    });
+
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const jestPackageJson = require(jestPath) as JSONSchemaForNPMPackageJsonFiles;
+
+    if (jestPackageJson.version) {
+      const [majorVersion] = jestPackageJson.version.split('.');
+
+      return parseInt(majorVersion);
```
SimenB, line 45:
```suggestion
      return parseInt(majorVersion, 10);
```

src/rules/no-deprecated-functions.ts
```diff
@@ -1,9 +1,56 @@
...
+  | 25
+  | 26
+  | 27
+  | number;
+
+interface EslintPluginJestSettings {
+  version: JestVersion;
+}
+
+const detectJestVersion = (): JestVersion => {
```
SimenB, line 33: could we make this so that once we've found the version it's cached? so we don't do extra fs per test file.

For the test you might need `jest.resetModules()` so it'll be recalculated

src/rules/__tests__/no-deprecated-functions.test.ts
G-Rath: I had completely forgotten that warning, good point.

(yeah I noticed they slipped that one into 14.0.0s release - it's an interesting one)

src/rules/no-deprecated-functions.ts
G-Rath: For now I've just gone with a one line `_clearCachedJestVersion` function, as `jest.resetModules` didn't work, and so we'd have to have that even if it was in another file.

I'll break it out into another file later :)

src/rules/no-deprecated-functions.ts
```diff
@@ -1,9 +1,65 @@
...
+    const jestPackageJson = require(jestPath) as JSONSchemaForNPMPackageJsonFiles;
+
+    if (jestPackageJson.version) {
+      const [majorVersion] = jestPackageJson.version.split('.');
+
+      return (cachedJestVersion = parseInt(majorVersion, 10));
+    }
+  } catch {}
+
+  throw new Error(
```
G-Rath, line 58: @SimenB I just remembered: I originally used `toThrowErrorMatchingInlineSnapshot`, but it kept complaining the snapshot was different everytime due to missing indentation at the start & end.

I'll try and put it into a reproduction 🐛

src/rules/no-deprecated-functions.ts
SimenB: you'd need to re-import the file after clearing so you don't hold on to a reference to the old module.

no biggie, tho

src/rules/__tests__/no-deprecated-functions.test.ts
```diff
@@ -1,49 +1,265 @@
...
+ * @param {JestVersion} jestVersion
+ *
+ * @return {Promise<string>}
+ */
+const setupFakeProjectDirectory = async (
+  jestVersion: JestVersion,
+): Promise<string> => {
+  const jestPackageJson: JSONSchemaForNPMPackageJsonFiles = {
+    name: 'jest',
+    version: `${jestVersion}.0.0`,
```
SimenB, line 34: we might add new "better" versions in minors, but this is fine for now 👍

docs/rules/no-deprecated-functions.md
```diff
@@ -9,10 +9,20 @@ of majors, eventually they are removed completely.
 ## Rule details
```
SimenB, line 9: Should this mention that it respects/varies based on jest version?

src/rules/no-deprecated-functions.ts
```diff
@@ -22,13 +78,27 @@ export default createRule({
   },
   defaultOptions: [],
   create(context) {
+    const jestVersion =
+      (context.settings as ContextSettings)?.jest?.version ||
```
SimenB, line 82: Can we do something so we don't need the type cast?

docs/rules/no-deprecated-functions.md
G-Rath: You mean something like [this](https://github.com/jest-community/eslint-plugin-jest/pull/564/files#diff-8471e3f5cb1f63cdf65debe2f509b2ddR12)?

(But could be improved by mentioning the actual setting, or at least linking back to that section in the readme)

src/rules/no-deprecated-functions.ts
G-Rath: No, as `context.settings` is just type `object` - it ideally should take a generic, but it's already two that are heavily inferred which means a generic would be unwieldy without support for a sigil to indicate [partial type argumentation inference](https://github.com/Microsoft/TypeScript/pull/26349)

docs/rules/no-deprecated-functions.md
SimenB: Linking back or something yeah, so people can be explicit

PR #571, opened 2020-05-09
Update recommended ruleset

G-Rath:
Let the games begin! 🎉 

The following rules are not in the recommended:
consistent-test-it
lowercase-name
no-alias-methods*
no-duplicate-hooks
no-expect-resolves
no-hooks
no-if
no-large-snapshots
no-test-return-statement
no-truthy-falsy
prefer-called-with
prefer-expect-assertions
prefer-hooks-on-top
prefer-inline-snapshots
prefer-spy-on
prefer-strict-equal
prefer-to-be-null*
prefer-to-be-undefined*
prefer-to-contain*
prefer-to-have-length*
prefer-todo
require-top-level-describe
require-to-throw-message
valid-title

*: in `style` ruleset

I think we should have `valid-title` as a recommended.
I wouldn't mind having `prefer-strict-equal` as recommended too, but don't think it's as important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ installations requiring long-term consistency.
 | [no-hooks](docs/rules/no-hooks.md)                                           | Disallow setup and teardown hooks                               |                  |              |
 | [no-identical-title](docs/rules/no-identical-title.md)                       | Disallow identical titles                                       | ![recommended][] |              |
 | [no-if](docs/rules/no-if.md)                                                 | Disallow conditional logic                                      |                  |              |
-| [no-interpolation-in-snapshots](docs/rules/no-interpolation-in-snapshots.md) | Disallow string interpolation inside snapshots                  |                  |              |
+| [no-interpolation-in-snapshots](docs/rules/no-interpolation-in-snapshots.md) | Disallow string interpolation inside snapshots                  | ![recommended][] |              |
 | [no-jasmine-globals](docs/rules/no-jasmine-globals.md)                       | Disallow Jasmine globals                                        | ![recommended][] | ![fixable][] |
 | [no-jest-import](docs/rules/no-jest-import.md)                               | Disallow importing Jest                                         | ![recommended][] |              |
 | [no-large-snapshots](docs/rules/no-large-snapshots.md)                       | disallow large snapshots                                        |                  |              |
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ installations requiring long-term consistency.
 | [valid-describe](docs/rules/valid-describe.md)                               | Enforce valid `describe()` callback                             | ![recommended][] |              |
 | [valid-expect](docs/rules/valid-expect.md)                                   | Enforce valid `expect()` usage                                  | ![recommended][] |              |
 | [valid-expect-in-promise](docs/rules/valid-expect-in-promise.md)             | Enforce having return statement when testing with promises      | ![recommended][] |              |
-| [valid-title](docs/rules/valid-title.md)                                     | Enforce valid titles                                            |                  | ![fixable][] |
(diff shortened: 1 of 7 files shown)


SimenB:
`'jest/no-jasmine-globals': 'warn',` error instead of warn, Circus will be default in 27, might nudge people to upgrade.

Other than that I think we're good? Not sure about `toStrictEqual` - I think it's better but it doesn't make much of a difference and the diff would be huge. Maybe though 🙂

jeysal:
Haven't got to looking at the full list yet, but quick comment on toStrictEqual: I also want this in recommended (because cases where only toEqual works are generally confusing and you should rewrite your code 😅), but we can't just autofix it because it may change the assertion result and we can't predict it. This is why I would wait until ESLint has the 'suggested change' editor integration that is currently under development afaik. React's hook dependencies rule will also use this.

G-Rath:
>  it may change the assertion result and we can't predict it

Can't you "just" run your test suite?

But for some reason I thought it already had an auto fix - if it did I wouldn't want to downgrade it, but since it doesn't I'm happy implementing it as a suggestion instead of a fix.

jeysal:
It does have one, but adding it to recommended will still break lots of tests when people upgrade the preset. And "downgrading" is exactly what React will also do with their rule, because it's the safer behavior.

Re running Jest - yeah that's a way to find out, but a bit hard for an ESLint plugin to do, sure we can spawn a Jest proc but how do we even know e.g. which CLI flags the user passes to Jest normally or what needs to be running before starting Jest 😅 not an option

G-Rath:
> It does have one

I literally opened the tests and went "oh yeah no autofix" 🤦 

> Re running Jest

I was meaning people on their codebases running their test suites, but I guess that is also why the suggestions api is good - it's opt-in model

SimenB:
>This is why I would wait until ESLint has the 'suggested change' editor integration that is currently under development afaik

It has been available in ESLint since November - I don't think we should wait for IDE integration. FWIW WebStorm has it, and while I don't know (or care 😀) if VS Code has it, we shouldn't wait for IDE support to use it.

thymikee:
my vote goes to:
- no-large-snapshots

jeysal:
I think a lot more matchers would make sense in recommended:

* agree with `no-large-snapshots`
* `no-restricted-matchers` with things like `toBeFalsy`, `toBeUndefined`, ... see also https://github.com/facebook/jest/issues/8172
* `prefer-called-with` with `HaveBeen` (past form makes clear it doesn't look into a crystal ball and see future calls), and consistency by settling on one is good, similar to `toBeUndefined()` vs `toBe(undefined)`
* `prefer-strict-equal` - hey, it's fixable
* `prefer-todo`
* `require-to-throw-message`
* `valid-title`

Also remove `prefer-to-be-{null,undefined}` from 'style' because why do they even exist given that we have `toBe`, I'd like `no-restricted-matchers` to enforce the **opposite**. May be controversial though? idk

G-Rath:
I think [`no-interpolation-inline-snapshot`](#553) would make sense to recommend.

SimenB:
@G-Rath there's some conflicts after #615, could you resolve them?

G-Rath:
Yup will do

G-Rath:
I've added `no-conditional-expect` to the recommended, as it's the replacement for `no-try-expect` which was recommened.

SimenB:
@g-rath every commit message needs `BREAKING CHANGE` in them. Add that then rebase merging to `next` should produce the result we want

SimenB:
@G-Rath I believe the `BREAKING CHANGE` part also needs a message (which is listed separately from the `Features` section in the release), see https://github.com/semantic-release/semantic-release/blob/master/README.md#commit-message-format

PR #572, opened 2020-05-09
ci: run tests against `eslint@7`

G-Rath:
Closes #568 
Closes #569

PR #573, opened 2020-05-09
feat: Support ESLint 7.x

MichaelDeBoey:
Follow-up of #572 

Rebased my previous branch, since you didn't change `eslint` dependency

Closes #567

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -18,11 +18,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [8.x, 10.x, 12.x, 13.x]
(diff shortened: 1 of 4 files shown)


G-Rath:
Thanks, but I'd prefer to not add more unactionable PRs to our pile, since this is covered by another PR (#567), and can't actually be done yet due to needing to drop support for node@8, and our downstream dependencies need to be updated, meaning this PR won't be actionable for least a week.

Additionally, node 14 isn't LTS yet, and we don't need to drop the `.x` since those are minor versions.

I appreciate your enthusiasm, but going to ask you hold fire until we've had a chance to catch up 😜

MichaelDeBoey:
No problem, just wanted to help out 🙂

PR #577, opened 2020-05-12
chore: update outdated dev dependencies

SimenB:
Closes #562
Closes #529
Closes #567
Closes #526
Closes #547
Closes #554
Closes #528


package.json
```diff
@@ -28,7 +28,7 @@
   "scripts": {
     "prepare": "yarn build && yarn postbuild",
     "lint": "eslint . --ignore-pattern '!.eslintrc.js' --ext js,ts",
-    "prettylint": "prettylint docs/**/*.md README.md package.json",
+    "prettylint": "prettier --check docs/**/*.md README.md package.json",
```
G-Rath, line 31: nit: should we maybe change the name?

package.json
```diff
@@ -47,47 +47,42 @@
...
+    "husky": "^4.2.5",
+    "jest": "^26.0.1",
+    "jest-runner-eslint": "^0.8.0",
+    "lint-staged": "^10.2.2",
+    "prettier": "^2.0.5",
     "resolve-from": "^5.0.0",
     "rimraf": "^3.0.0",
-    "semantic-release": "^15.13.28",
+    "semantic-release": "^17.0.7",
     "typescript": "^3.5.3"
```
G-Rath, line 74: nit: Would be cool to bump typescript - while symbolic, I feel it's nice to have it at a higher number.

package.json
SimenB: sure

package.json
SimenB: it's bumped to latest in the lockfile

SimenB:
This is going into `next`, so I think it'll conflict regardless

G-Rath:
> This is going into next

Woops completely missed that - makes a lot more sense now :joy:

PR #580, opened 2020-05-12
Update README.md

### LINKED ISSUE #222: New rule: no istanbul ignore without reason

antonk52:
Hi, @SimenB as per your [comment](https://github.com/jest-community/eslint-plugin-jest/issues/222#issuecomment-626325618) here is the addition of `eslint-plugin-istanbul` to the readme.

Closes #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,12 @@ ensure consistency and readability in jest test suites.
 
 https://github.com/dangreenisrael/eslint-plugin-jest-formatting
 
+### eslint-plugin-istanbul
+
+A set of rules to enforce good practices for istanbuljs.
+
+https://github.com/istanbuljs/eslint-plugin-istanbul
+
 [recommended]: https://img.shields.io/badge/-recommended-lightgrey.svg
 [fixable]: https://img.shields.io/badge/-fixable-green.svg
 [style]: https://img.shields.io/badge/-style-blue.svg


README.md
```diff
@@ -187,6 +187,12 @@ ensure consistency and readability in jest test suites.
 
 https://github.com/dangreenisrael/eslint-plugin-jest-formatting
 
+### eslint-plugin-istanbul
+
+A set of rules to enforce good practices for istanbuljs.
```
SimenB, line 192:
```suggestion
A set of rules to enforce good practices for Istanbul.
```

We should say something about how it's for code coverage, perhaps

antonk52: What do you think about this?
```
A set of rules to enforce good practices for Istanbul, a tool used to measure test coverage within Jest.
```

SimenB: I like it! Maybe "the tool" instead of "a tool"?

SimenB: that's not true for v8 code coverage, though... Maybe "the tool used ... expect when using V8 code coverage" or something like that?

@coreyfarrell ideas?

coreyfarrell: "... Istanbul, the default coverage collection tool"?  Sorry not sure I have good ideas.

SimenB: we'd keep using istanbul for reporting and stuff even if we stop using it for instrumenting. Maybe if we say "... Istanbul, the coverage reporting tool" or something? Not sure if it really matters

coreyfarrell: These hints are not processed by the istanbul reporting modules, they are only processed during instrumentation.  This is contrary to c8 which *I think* processes `c8 ignore next` during the v8-to-istanbul step.

G-Rath: How about:

```
A set of rules to enforce good practices for Istanbul, one of the code coverage tools used by Jest.
```

Also, is this is the only thing blocking? Because I think we might be bikeshedding a bit :joy:

G-Rath:
I'm going to close this to keep things tidy - happy to re-open if someone wants to pick it back up :)

PR #583, opened 2020-05-16
Deprecate rules

G-Rath:
Per discussion in #575


G-Rath:
Can I get a sanity check on the layout for squashing and merging?

Ideally we'd like a changelog entry for each deprecation, but I'm not sure if squashing this will give that - Semantic Pull Request is recommending a merge or rebase instead of a squash.

SimenB:
Yeah, rebase merge is the way to go 👍

SimenB:
might have been considered breaking, dunno... I have no idea how eslint handles deprecated rules. If it doesn't warn at all I guess it's fine, but then what's the point?

Meh 🤷

G-Rath:
No it's not breaking - I can't remember the exact way it handles them, but I know it's at least not breaking enough for eslint & typescript-eslint to deprecate rules in minors :)

PR #586, opened 2020-05-16
Use suggestion instead of fix

### LINKED ISSUE #483: [prefer-expect-assertions]: Suggest expect.hasAssertions fix
Requires [suggestion support in typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/pull/1250).

`prefer-expect-assertions` should suggest fixing by adding `expect.hasAssertions`.

We could also do one for `expect.assertions`, but the suggestions api doesn't support any form of user input, and it might upset people having to find out the hard way that `expect.assertions` requires them to actually provide an argument?

@SimenB thoughts?

G-Rath:
I've done these as `fix`s because they're safer, so feels like it's ok to do a patch in?

Also I realised how to break `no-test-callback` by not checking that it's type is an `Identifier` :joy:

Are there any other rules we think should use an suggestion instead of autofixing? (or that don't have any fixes already that could suggest).

fixes #483


G-Rath:
Also I just realised that we can simplify the fixer for `no-test-callback` 🥳

G-Rath:
Ah bugger - because `eslint@5` doesn't support suggestions, we can't hit our coverage target.

@SimenB would you be ok with ignoring the coverage target for eslint v5?

src/rules/prefer-strict-equal.ts
```diff
@@ -36,14 +37,19 @@ export default createRule({
...
           context.report({
-            fix: fixer => [
-              fixer.replaceText(
-                matcher.node.property,
-                EqualityMatcher.toStrictEqual,
-              ),
-            ],
             messageId: 'useToStrictEqual',
             node: matcher.node.property,
+            suggest: [
```
SimenB, line 42: add `suggestion: true` to the `meta.docs` field as well

src/rules/__tests__/no-test-callback.test.ts
```diff
@@ -16,41 +17,132 @@ ruleTester.run('no-test-callback', rule, {
     'test("something", function() {})',
     'test("something", async function () {})',
     'test("something", someArg)',
+    'test("something", (...args) => {args[0]();})',
```
SimenB, line 20: we should error here as well, although we can't apply the fix.

Jest just provides the one argument, so it's just the fixer that doesn't work, with the rule active we should still warn about this case

src/rules/prefer-strict-equal.ts
G-Rath: [We need `@typescript-eslint` to define the property first](https://github.com/typescript-eslint/typescript-eslint/pull/2033)

SimenB: `@ts-expect-error` in the meantime seems prudent 🙂

G-Rath: For that we'd have to update typescript ;)

(I'll do it quickly on another branch)

G-Rath: actually the PR got merged, so might as well wait until the release tomorrow & then update 🤷

G-Rath: Also fyi using `@ts-expect-error` is a bad move here because it completely breaks inference, so TS won't error on things like invalid `messageId` values 😬

src/rules/prefer-expect-assertions.ts
```diff
@@ -61,8 +74,17 @@ export default createRule({
...
+        '`expect.hasAssertions` expects no arguments',
+      assertionsRequiresOneArgument:
+        '`expect.assertions` excepts a single argument of type number',
+      assertionsRequiresNumberArgument: 'This argument should be a number',
       haveExpectAssertions:
         'Every test should have either `expect.assertions(<number of assertions>)` or `expect.hasAssertions()` as its first expression',
+      suggestAddingHasAssertions: 'Add `expect.hasAssertions()`',
+      suggestAddingAssertions:
+        'Add `expect.assertions(<number of assertions>)`',
+      suggestRemovingExtraArguments: 'Remove extra arguments',
```
G-Rath, line 87: @SimenB interested in hearing your thoughts on doing this as a suggestion vs autofix.

I did initially have it just fixing, but that causes a knock on effect by requiring setting `fixable` which gets pulled into the README, which is ok but at the same time I'm not entirely happy about because only 1 specific part of this rule is actually "fixable".

I'm reasonably happy with using suggestions, but interested in your thoughts on if this is valuable as a suggestion/fix/at-all, as can apply this to other rules too, such as `valid-expect`, `valid-describe`, etc

SimenB: can we have some "partially fixable" in the readme?

PR #587, opened 2020-05-17
Minor fixes

G-Rath:
Noticed this while working on `no-conditional-expect`, and figured it would be better as a bug release.
Also renamed the messageId to be different from the one used by `no-conditional-expect`.

### DIFF
--- a/src/rules/__tests__/no-if.test.ts
+++ b/src/rules/__tests__/no-if.test.ts
@@ -31,7 +31,7 @@ ruleTester.run('conditional expressions', rule, {
       errors: [
         {
           data: { condition: 'conditional' },
-          messageId: 'noConditionalExpect',
+          messageId: 'conditionalInTest',
         },
       ],
     },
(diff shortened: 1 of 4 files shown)


src/rules/__tests__/no-if.test.ts
```diff
@@ -552,30 +552,14 @@ ruleTester.run('if statements', rule, {
     },
   ],
   invalid: [
-    {
```
SimenB, line 555: why is this deleted?

G-Rath: Because it's a switch test in the `if` test block; switches are tested in [this group](https://github.com/jest-community/eslint-plugin-jest/pull/587/files#diff-af685d385381b07d87cda847c6b0ae4dR71).

I must have done a bad copy and paste; this test fails because it's data property has it expecting `if` instead of `switch` for `condition`.

G-Rath: [Here is it's duplicate](https://github.com/jest-community/eslint-plugin-jest/pull/587/files#diff-af685d385381b07d87cda847c6b0ae4dR179-R193) 🙂

SimenB: 👍

PR #594, opened 2020-05-26
chore: remove unneeded conditions

G-Rath:
closes #590

### DIFF
--- a/src/rules/no-test-callback.ts
+++ b/src/rules/no-test-callback.ts
@@ -63,13 +63,6 @@ export default createRule({
               fix(fixer) {
                 const { body } = callback;
 
-                /* istanbul ignore if https://github.com/typescript-eslint/typescript-eslint/issues/734 */
-                if (!body) {
-                  throw new Error(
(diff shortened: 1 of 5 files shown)


SimenB:
@G-Rath the bump itself is already in `next`, but the code changes are not. Conflicts with #607 - could you resolve?

G-Rath:
@SimenB as much as I'd like to get the major out, `@typescript-eslint` is going to release v4 very soon (I think they're going to release when TS4 does, which should be in a month tops).

Do we want to wait for that, or just cut another major in a few months? I'm fine with either.

SimenB:
The only reason v2 is breaking for us is because it drops old versions of node, right? Are there any breaking changes planned for v4 of the utils?

G-Rath:
Yeah, [not that I know of](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v4.0.0-rc).

The release is just for the new AST shape for optional chaining & co, which we don't use so we should be fine to bump in a minor.

SimenB:
Let's aim for that 🙂 Would be awesome to get a new major out in teh next couple of days

Issue #539, opened 2020-02-25
[expect-expect] reports the entire test as erroneous


DanielSWolf:
Within each unit test, the actual `expect` statement usually follows some amount of setup. As long as I'm writing this setup code, the test is technically invalid according to the `expect-expect` rule because it doesn't include any `expect`s. However, this means that about 80% of the time I spend writing unit tests, the entire function I'm writing is disfigured with red squigglies:

![image](https://user-images.githubusercontent.com/5963162/75225896-b66f6800-57ab-11ea-80c1-f0cb4c0bdc3d.png)

This doesn't just look bad. It also hides any actual errors I may have made. Only when I type the first `expect` do I see whether there are any other problems with my code.

Would it be possible for this rule to report *only part* of the test function as erroneous? For instance, just the `it`? Here's a mockup of how this might look:

![image](https://user-images.githubusercontent.com/5963162/75227433-78277800-57ae-11ea-8455-8fa774ad6efb.png)

G-Rath:
@DanielSWolf the `it` node is the node we report as erroneous.

You'll want to create a new issue for your editor, as we don't control how the errors are rendered.

For reference, IntelliJ/WebStorm does render the error as you'd like:

![image](https://user-images.githubusercontent.com/3151613/75231931-34b72300-581b-11ea-8582-fa1ee619538b.png)

To be honest, I'd like to report the callback rather than the `it` node, as it's a pretty overloaded node (in the screenshot above, there are four errors on it).

PR #599, opened 2020-05-29
fix(expect-expect): stop reporting whole test body

mmkal:
The expect-expect rule is quite annoying while you're writing a test, because until you put an assertion in it (often as the last thing you do), the entire test is marked as an error in IDEs which visually hides other lint/compiler errors. This makes the rule report on the `test`/`it` (callee) part rather than the whole `test('foo', () => {...})` expression.

Before:
image

After:
image

Opening this as a draft because I imagine some tests will fail - I made this change via the GitHub UI, I'll see what goes red when opening as a PR.

### DIFF
--- a/src/rules/expect-expect.ts
+++ b/src/rules/expect-expect.ts
@@ -109,7 +109,7 @@ export default createRule<
       },
       'Program:exit'() {
         unchecked.forEach(node =>
-          context.report({ messageId: 'noAssertions', node }),
+          context.report({ messageId: 'noAssertions', node: node.callee }),
         );
       },
     };


G-Rath:
The node we're reporting is the most accurate node for the rule; the issue you're describing is a problem with your IDE, as they've chosen to highlight the entire node.

While this change fixes the issue for your specific IDE, it changes the way the error is reported in other IDEs and the cli output.

This is a duplicate of #539

mmkal:
I don't think it's a problem with the IDE (which is VSCode, using the eslint plugin, so not exactly an unusual setup). The IDE is highlighting the full node because eslint is reporting with the full node, so it's working as expected.

It's a UX issue that so much noise is created for pretty much every test. This isn't an edge-case; it's _very normal_ to write tests this way - it's just [arrange-act-assert](https://wiki.c2.com/?ArrangeActAssert). The rule as it currently is makes the entire test look like an error for the "arrange" and "act" parts.

Issue #591, opened 2020-05-22
Resolve CI problems with jest-runner-eslint


G-Rath:
For the last week or so `jest-runner-eslint` seems to be failing on all eslint versions being run against node 8 & 10, with this error:

```
FAIL lint src/rules/utils.ts
  ● Test suite failed to run

    Failed to load plugin 'eslint-plugin' declared in '../../home/runner/work/eslint-plugin-jest/eslint-plugin-jest/.eslintrc.js': Cannot find module 'eslint-plugin-eslint-plugin'
    Require stack:
    - /tmp/eslint-plugin-jest-xuJrTI/__placeholder__.js
    Referenced from: /home/runner/work/eslint-plugin-jest/eslint-plugin-jest/.eslintrc.js

      at Object.resolve (node_modules/eslint/lib/shared/relative-module-resolver.js:44:50)
          at _normalizeObjectConfigDataBody.next (<anonymous>)
          at _normalizeObjectConfigData.next (<anonymous>)
```

I've played around with CI a bit, and confirmed:

* Running `yarn lint` passes fine
* `ls`ing in places in `node_modules` shows that the `eslint-plugin-eslint-plugin` is there and contains the expected files & folders.
* Running locally works fine for me against Node 10.20.1
* The error happens using the previous 2 or 3 versions of `jest-runner-eslint`, so I'm pretty sure it's not because of a change in `jest-runner-eslint`

This seems to be happening on all PRs, so it doesn't seem to be because of a specific change. I also can't find any evidence of a change in actions, Node version, etc (i.e it doesn't seem like a bad version of Node 8 & 10 was released).

I've also noticed it seems like `Semantic Pull Request` has stopped working on PRs; CodeCov is also not showing, but I suspect that's just because jest is failing so it doesn't think all coverage has been collected.

I've hit a bit of a dead end in terms of things to investigate, so need a fresh set of eyes & creds.

PRs that are failing with this error include #585 & #588

PR #603, opened 2020-06-19
chore: minor tidy up of code

G-Rath:
I made a new eslint plugin for linting config files to check they're invalid, and for use of deprecated rules.

There's actually a couple of commits, but wanting to test out github actions annotations.

### DIFF
--- a/src/rules/__tests__/no-deprecated-functions.test.ts
+++ b/src/rules/__tests__/no-deprecated-functions.test.ts
@@ -10,6 +10,11 @@ import rule, {
 
 const ruleTester = new TSESLint.RuleTester();
 
+// pin the original cwd so that we can restore it after each test
+const projectDir = process.cwd();
+
+afterEach(() => process.chdir(projectDir));
+
 /**
  * Makes a new temp directory, prefixed with `eslint-plugin-jest-`
  *
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 coverage/
 lib/
+!.eslintrc.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,7 @@ const globals = require('./src/globals.json');
 module.exports = {
   parser: require.resolve('@typescript-eslint/parser'),
   extends: [
+    'plugin:eslint-config/rc',
     'plugin:eslint-plugin/recommended',
     'plugin:eslint-comments/recommended',
     'plugin:node/recommended',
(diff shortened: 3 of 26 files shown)


src/rules/__tests__/no-deprecated-functions.test.ts
```diff
@@ -3,20 +3,36 @@ import * as os from 'os';
...
 import { JSONSchemaForNPMPackageJsonFiles } from '@schemastore/package';
 import { TSESLint } from '@typescript-eslint/experimental-utils';
+import rimraf from 'rimraf';
 import rule, {
   JestVersion,
   _clearCachedJestVersion,
 } from '../no-deprecated-functions';
 
+const root = process.cwd();
+afterEach(() => process.chdir(root));
```
SimenB, line 13: `process.chdir` is not actually supported by Jest. It _works_ since it's the real `chdir`, but it breaks a bunch of Jest's assumptions. Since it actually changes dir, it leaks between test files and the jest process itself.

It's not new though, so no need to change now. But something to keep in mind 🙂

G-Rath: Yeah, which is what's happening here - turns out the tests I wrote for the rule are what's been causing the weird CI failure.

I'm not sure why it only happens on Node 8 & 10, but the issue is that by changing the cwd to be outside of the projects folder, when looking for plugins `eslint` looks into the `node_modules` in `/tmp/eslint-plugin-jest-<tmpdir>` while using the config from `eslint-plugin-jest`, and so fails to find all its plugins.

I hacked this together to confirm, so now I'll break it out into another PR, and then after that I should be able to merge & ship a bunch of our open PRs.

G-Rath: On an aside, is there anything we could do in `jest` core to improve `process.chdir` support?

One such idea that comes to mind would be to do that I'm doing here: "pin" the cwd as part of `jest`'s setup phase, and then somehow make that available in places.

I'm not sure how worth it would be, but if there are any low hanging fruit that could be done to even just reduce a few of the assumptions, I'd be happy to explore :)

G-Rath:
This is based off #591 so that tests pass consistently, so that PR should be merged first.

I enabled `padding-line-between-statements` configured to require a newline before `return`s, and after variable blocks, as imo it produces tidier code that's easier to read by introducing some natural grouping + it's auto-fixable (which I've applied); can drop that commit if there's disagreement.

I also used `@typescript-eslint/no-unnecessary-condition` to find some conditional checks that are redundant 🎉

PR #604, opened 2020-06-19
test: ensure `cwd` is restored after each test

### LINKED ISSUE #591: Resolve CI problems with jest-runner-eslint

G-Rath:
fixes #591

I'm really curious to why this only failed on Node 8 & 10 - my guess is that `Module.createRequire` behaviour works slightly differently than the fallback methods, but haven't tested that in full.

### DIFF
--- a/src/rules/__tests__/no-deprecated-functions.test.ts
+++ b/src/rules/__tests__/no-deprecated-functions.test.ts
@@ -10,6 +10,11 @@ import rule, {
 
 const ruleTester = new TSESLint.RuleTester();
 
+// pin the original cwd so that we can restore it after each test
+const projectDir = process.cwd();
+
+afterEach(() => process.chdir(projectDir));
+
 /**
  * Makes a new temp directory, prefixed with `eslint-plugin-jest-`
  *


SimenB:
`Module.createRequire` is implemented in Jest, so shouldn't differ between node versions...

G-Rath:
Interesting - it's probably not that, but out of interest I'm guessing Jest provides that regardless of Node version? because it wasn't implemented until Node 12....

It's failing very inconsistently, so I think it's a race condition overall.

SimenB:
It's conditional bases on what's available in the running version of node, see https://github.com/facebook/jest/pull/9469

G-Rath:
> It's conditional

Good to know.

PR #606, opened 2020-06-21
chore: replace scripts with single `prepack` script

G-Rath:
I suspect most of these were legacy from when the project was Javascript?

It's small fry, but still each build + remove takes around ~3 seconds, which is done after every install because of `prepare` (which itself calls `build` & then `postbuild`, which is already called by `build`). `prepare` is also called as part of the publishing scripts, as well as `prepublishOnly`; and then finally we call it before running our tests via `pretest`, which just isn't needed because the tests are run against `src` not `lib`.

`prepack` is called before packing the project which is done as part of publishing when `npm pack` is called; this means you can also run `pack` locally to check what'll be included in the published pack (vs using `prepublishOnly`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,10 +23,8 @@
     "build": "babel --extensions .js,.ts src --out-dir lib --copy-files",
     "postbuild": "rimraf lib/__tests__ lib/**/__tests__",
     "lint": "eslint . --ignore-pattern '!.eslintrc.js' --ext js,ts",
-    "prepare": "yarn build && yarn postbuild",
-    "prepublishOnly": "yarn build",
+    "prepack": "yarn build",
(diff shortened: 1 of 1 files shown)


SimenB:
We need to build before test as we lint ourselves, no?

G-Rath:
Actually no we don't 😅

Most of our tests are using `RuleTester`, which is covered by `eslint-plugin-eslint-plugin`.

SimenB:
Ah, we just need the globals which comes from a JSON file now https://github.com/jest-community/eslint-plugin-jest/blob/c4f8c978411b2fdf32848e42c0dd5dffb1e6205b/.eslintrc.js#L3

PR #608, opened 2020-06-21
feat(valid-title): support `mustMatch` & `mustNotMatch` options

### LINKED ISSUE #233: New rule request: title-matches-pattern

G-Rath:
Closes #233
Closes #605

I implemented both `mustMatch` & `mustNotMatch` for the sake of completeness, since RegExp is not good at handling "not" matches.

### DIFF
--- a/docs/rules/valid-title.md
+++ b/docs/rules/valid-title.md
@@ -152,9 +152,11 @@ describe('foo', () => {
 ## Options
 
 ```ts
-interface {
+interface Options {
   ignoreTypeOfDescribeName?: boolean;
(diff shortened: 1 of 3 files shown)


mkosir:
Thx for the PR, exactly what I need, would be great if we could also add an array of keywords for `mustMatch` and `mustNotMatch`, the same as we do for `disallowedWords` (we can always go with regex of course).

PR #610, opened 2020-06-21
Adjust types packages

G-Rath:
`@typescript-eslint` provides much more accurate types for eslint (some of which are exported from `@types/eslint` directly), so overall we should always just use those types instead.

`jest-snapshot` brings in `@types/prettier`, which is why the import in `tools/generate-rules-table` doesn't error with tsc, but we should depend on it explicitly for safety.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "@types/dedent": "^0.7.0",
     "@types/jest": "^25.1.0",
     "@types/node": "^12.6.6",
+    "@types/prettier": "^1.19.0",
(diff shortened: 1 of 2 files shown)


SimenB:
We could use Yarn v2, it would have errored on the missing `@types/prettier`. Will need node 10 minimum, though

G-Rath:
eh I don't think there's much gain - I only added `@types/prettier` because I was removing `@types/eslint` and figured might as well do them at the same time 🤷

PR #611, opened 2020-06-22
feat(lowercase-name): support `ignoreTopLevelDescribe` option

### LINKED ISSUE #247: Option for `lowercase-name` rule to allow uppercase for top level describes
My preferred style is for top-level `describe`/`it`s to have names starting with a capital letter, and all nested blocks lower case. So then the concatenated test name reads like a sentence starting with a capital letter.

e.g.:

```js
describe('Monkeys', () => {
  describe('eat', () => {
    it('bananas', () => { /* ... */ } );
  });

  it('sleep in the daytime', () => { /* ... */ } );
});

it('Elephants are very big', () => { /* ... */ } );

// Monkeys eat bananas
// Monkeys sleep in the daytime
// Elephants are very big
```

At present this style is not possible to enforce with the `lowercase-name` rule.

How about:

1. Add `lowercase-name` `ignore` options `'top-describe'`, `'top-it'`, `'top-test'` (and `'top'` as a shortcut for all three)
2. Add a new `uppercase-name` rule with the same options.

Alternatively, a new rule called e.g. `name-case` could cover all possible permutations, with options something like:

```js
"lower"
"upper"

// Override for different statements
"lower", { "describe": "upper", "it": "upper", "test": "upper" }
"upper", { "describe": "ignore" }

// Override for all top-level statements
"lower", { "top": "upper" }

// Override for specific top-level statements
"lower", { "describe": "upper", "top": { "describe": "lower" } }
```

Do you feel this would be useful?

G-Rath:
closes #247
closes #428 

I think this should be the default in the next major, as it's a pretty common-case to have when testing classes.

### DIFF
--- a/docs/rules/lowercase-name.md
+++ b/docs/rules/lowercase-name.md
@@ -86,3 +86,21 @@ Example of **correct** code for the `{ "allowedPrefixes": ["GET"] }` option:
 
 describe('GET /live');
 ```
+
+### `ignoreTopLevelDescribe`
+
+This option can be set to allow only the top-level `describe` blocks to have a
+title starting with an upper-case letter.
+
+Example of **correct** code for the `{ "ignoreTopLevelDescribe": true }` option:
+
+```js
+/* eslint jest/lowercase-name: ["error", { "ignoreTopLevelDescribe": true }] */
+describe('MyClass', () => {
+  describe('#myMethod', () => {
+    it('does things', () => {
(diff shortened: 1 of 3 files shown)


G-Rath:
Really? I agree except for with things like classes & components, which I feel is pretty common; but then again it's definitely not the most obvious or pressing option 🤷

PR #615, opened 2020-06-25
chore: build recommended ruleset based on rules meta doc property

G-Rath:
While we don't change the recommended ruleset often, this ensures we'll keep the right properties up to date.

Looking over our tests I think some might be a bit overkill - I don't mind per say, but am conscientious of how many places changes have to happen i.e snapshotting + updating the rule count constant.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,12 @@
 import { readdirSync } from 'fs';
 import { join, parse } from 'path';
+import { TSESLint } from '@typescript-eslint/experimental-utils';
 import globals from './globals.json';
 import * as snapshotProcessor from './processors/snapshot-processor';
+
+type RuleModule = TSESLint.RuleModule<string, unknown[]> & {
+  meta: Required<Pick<TSESLint.RuleMetaData<string>, 'docs'>>;
+};
 
 // can be removed once we've on v3: https://github.com/typescript-eslint/typescript-eslint/issues/2060
 declare module '@typescript-eslint/experimental-utils/dist/ts-eslint/Rule' {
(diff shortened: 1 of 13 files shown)


src/index.ts
```diff
@@ -25,50 +28,37 @@ const excludedFiles = ['__tests__', 'utils'];
 const rules = readdirSync(rulesDir)
   .map(rule => parse(rule).name)
   .filter(rule => !excludedFiles.includes(rule))
-  .reduce(
+  .reduce<Record<string, RuleModule>>(
     (acc, curr) =>
-      Object.assign(acc, { [curr]: importDefault(join(rulesDir, curr)) }),
+      Object.assign(acc, {
```
SimenB, line 33: can use object spread instead of `Object.assign`

src/index.ts
```diff
@@ -25,50 +28,37 @@ const excludedFiles = ['__tests__', 'utils'];
...
+        [curr]: importDefault(join(rulesDir, curr)) as RuleModule,
+      }),
     {},
   );
 
+const recommendedRules = Object.entries(rules).reduce(
+  (acc, [name, rule]) =>
+    rule.meta.docs?.recommended
```
SimenB, line 41: `docs` shouldn't be optional, can you narrow the type?

G-Rath: Yeah its a bit annoying - the types provided by `@typescript-eslint` are almost perfect, except for these edge cases which are somewhat tedious to improve on at the source.

I can actually improve this typing without too much work, which is nice.

src/index.ts
```diff
@@ -1,8 +1,13 @@
 import { readdirSync } from 'fs';
 import { join, parse } from 'path';
+import { TSESLint } from '@typescript-eslint/experimental-utils';
 import globals from './globals.json';
 import * as snapshotProcessor from './processors/snapshot-processor';
 
+type RuleModule = TSESLint.RuleModule<string, unknown[]> & {
```
SimenB, line 7: should we use this type for all rules where we declare them as well?

G-Rath: How do you mean?

Ideally we should try and use the inferred type that's returned from `createRule` as much as possible for each rule, as that has the proper types for their message ids & options.

We could switch to using a barrel export like `@typescript-eslint` does, where we have `index.ts` that explicitly exports each rule:

```
import noIfs from './no-ifs';

export default  {
  'no-ifs': noIfs,
}
```

That would let us import all the rules here as a single default import.

<details>
<summary>
we can even script it
</summary>

```
#!/usr/bin/env ts-node-transpile-only

import * as fs from 'fs';
import * as path from 'path';
import prettier from 'prettier';
import { prettier as prettierConfig } from '../package.json';

const toCamelCase = (str: string) => {
  const words = str.split('-');

  return [
    words.shift(),
    ...words.map(w => w.substring(0, 1).toUpperCase() + w.substring(1)),
  ].join('');
};

const excludedFiles = ['__tests__', 'utils', 'index'];

const ruleFiles = fs
  .readdirSync(path.resolve(__dirname, '../src/rules'))
  .map(file => path.parse(file).name)
  .filter(file => !excludedFiles.includes(file));

const contents = [
  ...ruleFiles.map(rule => `import ${toCamelCase(rule)} from './${rule}';`),
  '',
  'export default {',
  ...ruleFiles.map(rule => `  '${rule}': ${toCamelCase(rule)},`),
  '};', //
].join('\n');

fs.writeFileSync(
  path.resolve(__dirname, '../src/rules/index.ts'),
  prettier.format(contents, { parser: 'typescript', ...prettierConfig }),
);
```
</details>

G-Rath: (I'm going to merge this, but happy to continue the discussion 🙂 )

SimenB: I meant doing something here so we use the same type: https://github.com/jest-community/eslint-plugin-jest/blob/75f8897944059baffd80d665e3d9cbd5641d46b1/src/rules/utils.ts#L12-L16
Not sure how feasible it is, tho?

PR #632, opened 2020-07-26
feat(no-large-snapshots): deprecate `whitelistedSnapshots` for new name

G-Rath:
I'd be happy to deprecate passing in RegExp instances at the same time if we want to simplify things. I don't use the rule myself so not sure how heavily the RegExp support is used.


G-Rath:
Yes (thats the plan) and yes - I think we're pretty good to go for the new major - the main thing left is to decide on what rules to make recommended.

PR #638, opened 2020-08-02
Remove unneeded checks

G-Rath:
Spotted this while doing #637 - it's impossible to craft an AST tree where a Node is both an `ExpectMember` & does not have a parent.

Also found a few conditions that can be simplified using optional chaining 🎉

### DIFF
--- a/src/rules/valid-expect.ts
+++ b/src/rules/valid-expect.ts
@@ -249,7 +249,11 @@ export default createRule<
 
         const parentNode = matcher.node.parent;
 
-        if (!modifier || modifier.name === ModifierName.not) {
+        if (
+          !parentNode.parent ||
+          !modifier ||
+          modifier.name === ModifierName.not
+        ) {
           return;
         }
         /**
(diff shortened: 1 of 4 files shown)


src/rules/valid-expect.ts
```diff
@@ -249,20 +249,15 @@ export default createRule<
 
         const parentNode = matcher.node.parent;
 
-        if (
-          !modifier ||
-          !parentNode ||
-          !parentNode.parent ||
-          modifier.name === ModifierName.not
-        ) {
+        if (!modifier || modifier.name === ModifierName.not) {
```
SimenB, line 252: We don't wanna bail early?

G-Rath: Do you mean for `parentNode.parent`?

It's actually an impossibility: the only ast node that can not have a parent is `Program`.

See https://github.com/typescript-eslint/typescript-eslint/issues/1417

SimenB: We have an optional chain further down, tho. I'd prefer an early return or a type refinement which captures the impossibility

G-Rath: Fair enough 🙂

PR #643, opened 2020-08-13
fix(config): lint only test files

arvigeus:
It matches the behavior of jest when it's looking for files to test
https://github.com/facebook/jest/blob/96097ade1794bfe21c6207e080aa4a7192d93041/packages/jest-config/src/Defaults.ts#L60


G-Rath:
That property is a configurable, so users can provide their own options to jest for that property that would defeat this.

If you're having problems with `eslint-plugin-jest` reporting on files that are not for tests, you should use [`overrides`](https://eslint.org/docs/user-guide/configuring#disabling-rules-only-for-a-group-of-files).

arvigeus:
My reason for it is that sometimes eslint will incorrectly report files that looks like test files, but are not. It's an extreme case, but I thought it might worth the enhancement.

PR #651, opened 2020-08-23
No test callback/support hooks

### LINKED ISSUE #649: `no-test-callback` should support hooks
Since these can take `done` as well.

G-Rath:
Fixes #649

@SimenB are we considering this a breaking change?


SimenB:
@G-Rath target `next`, land it, then land `next`?

Issue #641, opened 2020-08-10
[expect-expect]: For usability, narrow warning range to the test definition, not the body.


jeffvandyke:
Often when working (in either Vim using ALE or in VSCode using ESLint), I'll be writing a test, this rule violation will cover the whole test body until I've written in the `expect()` line, which hides other warnings that I'm more interested in while I'm writing the test, because I'm still setting up things to `expect`, which usually takes some time to write. Sometimes, I put an `expect(false)` just so that I can see lints about arrays, promises, and other such things while I'm typing, and needing to put in a dummy `expect` while I'm writing the tests seems unnecessary and flow-interrupting.

I think this would be best solved by keeping this rule, but only highlighting as a warning the name of the test instead of the entire body, to make room for other lints to show up. Thoughts or other ideas?

G-Rath:
Duplicate of #539.

The tl;dr is that this is an IDE problem - the node we're reporting on is the right one as it's the closest to the actual error which is that the function is missing an `expect` assertion. We can't account for how IDEs render these errors, and a change to improve the experience in one might worsen it in another.

jeffvandyke:
@G-Rath, I disagree with this analysis: I think there's an important semantic in the `range` returned from lint rules, and this rule is hurting more than helping in what it reports as the problem range.

I think more than one IDE (VSCode and Vim) is interpreting a violation's range in a reasonable way - by marking that code from beginning to end as a problem visually, but violations of this rule are normal when writing tests, so it's normal to have this rule say that there's a problem with the test from beginning to end, hiding all other more important errors (I place this error in the same usefulness category as no-unused-vars - good check, but likely going to be fixed soon). I think in the tradeoff between what's convenient for this rule's logic and what's convenient for the user, the balance shifts heavily toward reporting a smaller range due to how very inconvenient it is for the user and how often we run into this - every time we write a test.

For an analogy, imagine a teacher marking up a student's first draft, highlighting grammar and spelling errors. Imagine the student missed one of the requirements that this paper needed to use a source from Encyclopedia Britannica. Should the teacher a): write a note on the top indicating the missed requirement, or b): highlight the entire paper from beginning to end, saying that he/she couldn't find the required citation anywhere in the paper? I think that the same answer applies here as well.

[...]

PR #654, opened 2020-08-30
refactor(expect-expect): report test callee instead of call expression

remcohaszing:
When one starts writing a test, they often start with something like the following:

image

This triggers the `jest/expect-expect` rule for a good reason, but it’s annoying, confusing even, when the entire test block has a red underline until an expect statement is added.

This changes the rule to report the callee, only. This limits the scope of the reported code, so the red lines that do appear while typing are more meaningful, i.e. `result` is unused:

image

### DIFF
--- a/src/rules/__tests__/expect-expect.test.ts
+++ b/src/rules/__tests__/expect-expect.test.ts
@@ -63,7 +63,7 @@ ruleTester.run('expect-expect', rule, {
       errors: [
         {
           messageId: 'noAssertions',
-          type: AST_NODE_TYPES.CallExpression,
+          type: AST_NODE_TYPES.Identifier,
         },
       ],
     },
@@ -72,7 +72,7 @@ ruleTester.run('expect-expect', rule, {
       errors: [
         {
           messageId: 'noAssertions',
-          type: AST_NODE_TYPES.CallExpression,
+          type: AST_NODE_TYPES.Identifier,
         },
       ],
     },
@@ -81,7 +81,7 @@ ruleTester.run('expect-expect', rule, {
       errors: [
         {
           messageId: 'noAssertions',
-          type: AST_NODE_TYPES.CallExpression,
+          type: AST_NODE_TYPES.Identifier,
         },
       ],
     },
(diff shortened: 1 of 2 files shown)


G-Rath:
Please see #641 & related issues - the node we're reporting on is the right one for this rule, as the problem is with the callback argument, *not* the function body.

It's up to the editor on how it renders eslint results.

remcohaszing:
I agree technically the function callback, not the identifier. However, I disagree it’s up to the editor how the ESLint results are highlighted.

ESLint rules report a text range, or a node which has a text range. I think it’s fair to say the editor should underline the text range with a red squirly line.

I believe it’s a bad UX if such a large piece of the code is underlined. Reporting identifier may not be the best solution. What about reporting the last character of the callback.

```js
   it('should do something', () => {
     const result = something();
   });
// ^-- Report this character

   it('should do something', () =>
     something()
     //        ^-- Report this character, because the test has no closing body bracket
   );
```

This could be interpreted as: I got to the end of your test case, but still didn’t find an expectation.

Alternatively all code between the last statement and the closing bracket of the test could be reported

```js
   it('should do something', () => {
     const result = something();
   });
// ^-- Report this character and all whitespace before it.

   it('should do something', () =>
     something()
   );
// ^-- Report all whitespace before this character.
```

G-Rath:
You've just said you disagree that it's up to the editor, and then proceeded to provide possible solutions that are entirely based around assuming how editors render the results :)

Our UX scope does not extend to how editors render ranges from ESLint, since we can't control that at all.

While we can technically report "all whitespace before this character", we have no assurance on how that'll be render, and in some cases ESLint itself will overrule us in favor of more accurate ranges (i.e it'll lack on to the closest node).

IMO the UX you're describing of highlighting the last set of characters in a callback is a worse one, because both of the above and because it makes it harder to notice.

Also remember that ESLint output is not just for IDEs - take for example annotations in GIthub Actions:

```
it(
  'should do something',
  () => {
    //
  }
);
```

The above code is perfectly valid, and I've seen it in the wild - right now by reporting on the correct node, GHA will show its annotation on the *callback* which is where the error is.

I don't have a major problem with the change that is suggested in this case - my problem is that this is a UX problem with *VSCode*, not us. We already have a few other rules that are technically reporting in the wrong node that I've held off correcting because of their UX problem.

The role of an eslint-plugin is to report where errors are as accurately as possible, both in presence (i.e little to no false positives or negatives) and location. The role of an IDE integration is to present what we report in a manner that ideally has a good user experience.

PR #660, opened 2020-09-05
fix(no-deprecated-functions): do not throw error when jest is missing

foray1010:
Throwing error causes trouble on my general eslint config because it means to be used in project with jest or not. Jest is not listed as `peerDependencies` of this package too, so I think it makes sense not to enforce it.

### DIFF
--- a/src/rules/__tests__/no-deprecated-functions.test.ts
+++ b/src/rules/__tests__/no-deprecated-functions.test.ts
@@ -190,7 +190,7 @@ describe('the rule', () => {
       return true;
     });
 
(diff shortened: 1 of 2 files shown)


G-Rath:
This plugin only makes sense when you're linting files that are used with Jest, so you should always have a Jest version even if you don't have `jest` installed.

The reason why `jest` isn't a peer dependency is that we don't *require* it to run - by having the rule on, you're wanting it to run, but in order to run the rule needs to know what version of jest you're targeting to work; hence the error.

PR #672, opened 2020-09-19
fix(no-if): check both types of function expression

### LINKED ISSUE #670: [no-if]: Works with arrow function, fails in standard function
When `jest/no-if` is enabled, An if statement within `() => {` is caught no problem. But when it's inside `function () {`, it is not caught.

Example:

```js
const condition = true

test('arrow function with condition', () => {
  let sum = 0
  if (condition) { // this fails jest/no-if
    sum = 1 + 2
  }
  expect(sum).toBe(3);
});

test('standard function with condition', function () {
  let sum = 0
  if (condition) { // this is not caught by jest/no-if
    sum = 1 + 2
  }
  expect(sum).toBe(3);
});
```

[Reproducible Code](https://github.com/dtom90/jest-no-if)

G-Rath:
Fixes #670

@SimenB I've looked at the [original PR](https://github.com/jest-community/eslint-plugin-jest/pull/293) and there's no discussion around this, so I'm guessing it was working off the fact that you're meant to use arrow functions with jest?

### DIFF
--- a/src/rules/__tests__/no-if.test.ts
+++ b/src/rules/__tests__/no-if.test.ts
@@ -13,15 +13,20 @@ const ruleTester = new TSESLint.RuleTester({
 ruleTester.run('conditional expressions', rule, {
   valid: [
     'const x = y ? 1 : 0',
-    {
-      code: dedent`
-        it('foo', () => {
-          const foo = function(bar) {
-            return foo ? bar : null;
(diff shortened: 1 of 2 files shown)


SimenB (approved):
Just forgot since I haven\t used them for 5+ years 😛

PR #677, opened 2020-09-25
feat(prefer-expect-assertions): support `onlyFunctionsWithAsyncKeyword ` option

mcampa:
We would like to add this new option on the `prefer-expect-assertions` rule, so the rule only applies to `async` tests.
I hope this you like this idea. I'm open for suggestions.

### DIFF
--- a/docs/rules/prefer-expect-assertions.md
+++ b/docs/rules/prefer-expect-assertions.md
@@ -55,3 +55,45 @@ test('my test', () => {
   expect(someThing()).toEqual('foo');
 });
 ```
+
+## Options
+
+#### `onlyFunctionsWithAsyncKeyword`
+
+When `true`, this rule will only warn for tests that use the `async` keyword.
+
+```json
+{
+  "rules": {
+    "jest/prefer-expect-assertions": [
+      "warn",
+      { "onlyFunctionsWithAsyncKeyword": true }
+    ]
+  }
+}
+```
+
+When `onlyFunctionsWithAsyncKeyword` option is set to `true`, the following
+pattern would be a warning:
+
+```js
+test('my test', async () => {
+  const result = await someAsyncFunc();
+  expect(result).toBe('foo');
+});
+```
+
+While the following patterns would not be considered warnings:
+
+```js
+test('my test', () => {
+  const result = someFunction();
+  expect(result).toBe('foo');
+});
+
+test('my test', async () => {
+  expect.assertions(1);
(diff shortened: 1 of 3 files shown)


src/rules/__tests__/prefer-expect-assertions.test.ts
```diff
@@ -205,6 +205,21 @@ ruleTester.run('prefer-expect-assertions', rule, {
         },
       ],
     },
+    {
+      code: dedent`
+      it("it1", async function() {
```
G-Rath, line 210: Could you indent your `dedent`'d tests by one level?

G-Rath (changes requested):
This seems like a reasonable option to have - aside from the code change I'd requested, all that's missing is updating the rules documentation :)

mcampa:
Hi @G-Rath, I think I addressed all your comments.
Let me know if there is anything else needed.

src/rules/__tests__/prefer-expect-assertions.test.ts
```diff
@@ -205,6 +205,21 @@ ruleTester.run('prefer-expect-assertions', rule, {
         },
       ],
     },
+    {
+      code: dedent`
+        it("it1", async function() {
+          expect(someValue).toBe(true)
+        })
+    `,
```
G-Rath, line 213: Could you ensure these are indenting to be at the same level as the object properties?

(Prettier won't change them because they're technically content)

docs/rules/prefer-expect-assertions.md
```diff
@@ -55,3 +55,17 @@ test('my test', () => {
   expect(someThing()).toEqual('foo');
 });
 ```
+
+## Options
+
+#### `asyncOnly`
+
+Run this rule only on async test functions
```
G-Rath, line 63:
```suggestion
When `true`, this rule will only warn for tests that use the `async` keyword.
```

Could you also include some examples in the format further up in the doc?

G-Rath:
@mcampa cheers, it's looking good - I am now wondering if maybe it should be something like `onlyFunctionsWithAsyncKeyword`, to be really clear that by "async" we're meaning the keyword, *not* functions that return a promise.

What do you think?

mcampa:
> What do you think?

makes perfect sense, I'll make the change.

mcampa:
I believe all comments were addresses. Back to you @G-Rath 🤞

G-Rath (approved):
Awesome, let's ship it!

PR #684, opened 2020-10-05
chore: enable `eslint-plugin/test-case-property-ordering` rule

G-Rath:
I was going to enable other rules too I think, but then forgot what ones. Also I tried enabling some of the other `eslint-plugin` rules but turns out they only work with `module.exports`.


SimenB (approved):
nice!

PR #689, opened 2020-10-14
chore: run release workflow when other workflows complete

SimenB:
I think this makes sense

https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,10 @@
 name: Release
 on:
-  push:
-    branches:
-      - master
+  workflow_run:
+    workflows: [Lint, 'Unit tests']
+    branches: [master]
+    types:
+      - completed
 
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,8 +1,10 @@
 name: Release
 on:
-  push:
-    branches:
-      - master
+  workflow_run:
+    workflows: [Lint, 'Unit tests']
+    branches: [master]
+    types:
+      - completed
```
SimenB, line 7: I _assume_ this is "completed successfully". Might be wrong though

SimenB: When you assume...

https://github.community/t/workflow-run-success-type-reference/133194/2

G-Rath: Oh yeah no this doesn't run only on success - I played around with it last month for something else and found this out 😬

Issue #692, opened 2020-10-17
`valid-title` false positive on method that returns a string


golergka:
I have code for automatic test construction that looks somewhat like this:
```Typescript
	export class Thing {
		public prettyPrint(): string { return "description" }
	}

	function describeThing(
		thingParams: ThingParams,
		description: (state: Thing) => void
	) {
		const thing = new Thing(thingParams)
		describe(thing.prettyPrint(), () => description(thing))
	}
```
(Names changed and code removed, of course). Unfortunately, linting this gives me a false positive on describe line: `error  Title must be a string  jest/valid-title`.

G-Rath:
The point of valid-title is to valid string titles, so there's no point in using it if you're not going to have string titles.
You can use the standard eslint-disable comments to control applying the rule on a case-by-case :)

golergka:
But it is a string. [Documentation says](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-title.md):

> Titles for test blocks should always be a string literal or expression.

And `thing.PrettyPrint()` is a string expression.

G-Rath:
> And thing.PrettyPrint() is a string expression.

No, it's a call expression:

![image](https://user-images.githubusercontent.com/3151613/96333524-eebf1000-10c6-11eb-894c-9db97b938706.png)

golergka:
Hm, it seems that I am technically wrong; I wasn't aware of terms "call expression" and "string expression" in Javascript, and meant "string expression" as "expression of type string" — thanks for educating me.

However, I don't think that I'm the only developer that would make such mistake. May be the docs could be updated to eliminate this misunderstanding?

G-Rath:
> May be the docs could be updated to eliminate this misunderstanding?

PRs are always welcome 🙂

PR #694, opened 2020-10-17
docs(valid-title): reduce ambiguity

golergka:
Discussed in #692

### DIFF
--- a/docs/rules/valid-title.md
+++ b/docs/rules/valid-title.md
@@ -43,7 +43,7 @@ xtest('foo', () => {});
 
 **titleMustBeString**
 
-Titles for test blocks should always be a string literal or expression.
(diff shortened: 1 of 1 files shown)


docs/rules/valid-title.md
```diff
@@ -43,7 +43,7 @@ xtest('foo', () => {});
 
 **titleMustBeString**
 
-Titles for test blocks should always be a string literal or expression.
+Titles for test blocks should always be a string literal or string template.
```
G-Rath, line 46:
```suggestion
Titles for test blocks should always be a string.
```

I think probably best to just `string` and leave it at that.

G-Rath (changes requested):

G-Rath (approved):

PR #701, opened 2020-10-28
fix: improve support for it.each involving tagged template literals

k-yle:
`it.each` has [an unusual syntax](https://jestjs.io/docs/en/api#2--testeachtablename-fn-timeout-) which means `no-disabled-tests`, `no-test-prefixes`, and `consistent-test-it` don't work for it. 

This PR makes `it.each` work with those three rules

### DIFF
--- a/src/rules/__tests__/no-standalone-expect.test.ts
+++ b/src/rules/__tests__/no-standalone-expect.test.ts
@@ -58,6 +58,10 @@ ruleTester.run('no-standalone-expect', rule, {
   ],
   invalid: [
     {
+      code: `(() => {})('testing', () => expect(true))`,
+      errors: [{ endColumn: 41, column: 29, messageId: 'unexpectedExpect' }],
+    },
+    {
       code: `
         describe('scenario', () => {
           const t = Math.random() ? it.only : it;
--- a/src/rules/__tests__/no-standalone-expect.test.ts
+++ b/src/rules/__tests__/no-standalone-expect.test.ts
@@ -58,7 +58,7 @@ ruleTester.run('no-standalone-expect', rule, {
   ],
   invalid: [
     {
-      code: `(() => {})('testing', () => expect(true))`,
+      code: "(() => {})('testing', () => expect(true))",
       errors: [{ endColumn: 41, column: 29, messageId: 'unexpectedExpect' }],
     },
     {
--- a/src/rules/no-test-prefixes.ts
+++ b/src/rules/no-test-prefixes.ts
@@ -1,3 +1,4 @@
+import { AST_NODE_TYPES } from '@typescript-eslint/experimental-utils';
 import { createRule, getNodeName, isDescribe, isTestCase } from './utils';
 
 export default createRule({
@@ -27,12 +28,17 @@ export default createRule({
 
         if (!preferredNodeName) return;
 
+        const funcNode =
+          node.callee.type === AST_NODE_TYPES.TaggedTemplateExpression
+            ? node.callee.tag
+            : node.callee;
+
         context.report({
           messageId: 'usePreferredName',
           node: node.callee,
           data: { preferredNodeName },
           fix(fixer) {
-            return [fixer.replaceText(node.callee, preferredNodeName)];
+            return [fixer.replaceText(funcNode, preferredNodeName)];
           },
         });
       },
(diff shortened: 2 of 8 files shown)


src/rules/__tests__/no-standalone-expect.test.ts
```diff
@@ -57,6 +57,10 @@ ruleTester.run('no-standalone-expect', rule, {
     },
   ],
   invalid: [
+    {
+      code: `(() => {})('testing', () => expect(true))`,
```
G-Rath, line 61:
```suggestion
      code: "(() => {})('testing', () => expect(true))",
```

src/rules/no-test-prefixes.ts
```diff
@@ -43,12 +49,14 @@ export default createRule({
 function getPreferredNodeName(nodeName: string) {
```
G-Rath, line 49: nit: a good refactor for this could be to do `split('.')` - then you could just splice the `only`/`skip` in as element 1, and return `.join('.')`.

I'm fine with this how it is, so that doesn't have to be done - just thinking of the future :)

src/rules/utils.ts
```diff
@@ -580,10 +580,16 @@ export interface JestFunctionCallExpressionWithIdentifierCallee<
   callee: JestFunctionIdentifier<FunctionName>;
 }
 
+export interface JestFunctionCallTaggedTemplateExpression
```
G-Rath, line 583:
```suggestion
interface JestFunctionCallExpressionWithTaggedTemplateCallee
```

src/rules/utils.ts
```diff
@@ -580,10 +580,16 @@ export interface JestFunctionCallExpressionWithIdentifierCallee<
...
+  extends TSESTree.CallExpression {
+  callee: TSESTree.TaggedTemplateExpression;
+}
+
 export type JestFunctionCallExpression<
   FunctionName extends JestFunctionName = JestFunctionName
 > =
   | JestFunctionCallExpressionWithMemberExpressionCallee<FunctionName>
+  | JestFunctionCallTaggedTemplateExpression
```
G-Rath, line 592:
```suggestion
  | JestFunctionCallExpressionWithTaggedTemplateCallee
```

src/rules/utils.ts
```diff
@@ -592,6 +598,7 @@ const joinNames = (a: string | null, b: string | null): string | null =>
 export function getNodeName(
   node:
     | JestFunctionMemberExpression<JestFunctionName>
+    | TSESTree.TaggedTemplateExpression
     | JestFunctionIdentifier<JestFunctionName>,
```
G-Rath, line 602:
```suggestion
    | JestFunctionIdentifier<JestFunctionName>
    | TSESTree.TaggedTemplateExpression,
```

nit: me appealing to my perfectionist brain that likes the alignment, feel free to ignore 😅

G-Rath (changes requested):
Wow, thanks for this - I'm not sure how we missed it for so long!

I've left a few nits, and asked to have the interface renamed and its export removed (since its not needed); otherwise I think this is good to go!

src/rules/utils.ts
k-yle: i've changed it (: was just trying to reduce noise in the git diff

src/rules/no-test-prefixes.ts
k-yle: I think I'll leave this one as-is

src/rules/__tests__/no-standalone-expect.test.ts
SimenB: a lint rule should pick this up, fwiw

k-yle:
hey @G-Rath, I've made the changes you requested - is there anything else I need to do?

G-Rath (approved):
> is there anything else I need to do?

You need to prod me because I apparently forgot to approve and merge this 😅

shobhitsharma:
@k-yle getting some `error  Return a Promise instead of relying on callback parameter  jest/no-done-callback` after this upgrade. Maybe more info for this change would help.

k-yle:
@shobhitsharma yep looks like this PR broke `jest/no-done-callback` for `it.each` 🙊 working on it right now

k-yle:
fixed in #708 

cc @G-Rath

SimenB:
@k-yle @G-Rath (is it a new zealand thing to go with that structure for usernames? 😀) I've revert this change in #713, but kept all the passing tests (plus an extra regression test for #710). A new go would be great 👍

k-yle:
@SimenB I won't have time until next Tuesday so if someone else wants to do it before then, go for it 👍   

it may also be above my skill level to fix [this edge case](https://github.com/jest-community/eslint-plugin-jest/issues/710#issuecomment-726081134) - I don't know how `no-done-callback` could work with that syntax

also, do we know how to reproduce #711?

PR #708, opened 2020-11-12
fix(no-done-callback): fix regression with it.each

### LINKED ISSUE #709: test.each no-done-callback false positive in version 24.1.1
Plugin version: 24.1.1
Last working version: 24.1.0
Rule **jest/no-done-callback** has false positives with test.each
```
test.each`
  size   | chunkSize    | sum
  ${50}  | ${undefined} | ${(50 * 49) / 2}
`(
  "sums array indices of size $size with chunkSize=$chunkSize => $sum",
  async ({size, chunkSize, sum}) => {
    await expect(
      reduce(new Array(size).fill(0).map((v, index) => index), (ret, it, index) => ret + index, 0, chunkSize)
    ).resolves.toBe(sum);
  }
);
```

k-yle:
Closes #709

my recent PR #701 caused a false positive for `no-done-callback` when used with jest-each (sorry!)

`it.each` uses the first argument for options, and the second argument for the done callback.

This PR fixes the bug and adds test cases for it.each

### DIFF
--- a/src/rules/no-done-callback.ts
+++ b/src/rules/no-done-callback.ts
@@ -2,11 +2,20 @@ import {
   AST_NODE_TYPES,
   TSESTree,
 } from '@typescript-eslint/experimental-utils';
-import { createRule, isFunction, isHook, isTestCase } from './utils';
+import {
+  createRule,
+  getNodeName,
+  isFunction,
+  isHook,
+  isTestCase,
+} from './utils';
 
 const findCallbackArg = (
   node: TSESTree.CallExpression,
+  isJestEach: boolean,
 ): TSESTree.CallExpression['arguments'][0] | null => {
+  if (isJestEach) return node.arguments[1];
+
   if (isHook(node) && node.arguments.length >= 1) {
     return node.arguments[0];
   }
@@ -41,17 +50,21 @@ export default createRule({
   create(context) {
     return {
       CallExpression(node) {
-        const callback = findCallbackArg(node);
+        // done is the second argument for it.each, not the first
+        const isJestEach = !!getNodeName(node.callee)?.endsWith('.each');
+
+        const callback = findCallbackArg(node, isJestEach);
+        const callbackArgIndex = +isJestEach;
 
         if (
           !callback ||
           !isFunction(callback) ||
-          callback.params.length !== 1
+          callback.params.length !== 1 + callbackArgIndex
         ) {
           return;
         }
 
-        const [argument] = callback.params;
+        const argument = callback.params[callbackArgIndex];
 
         if (argument.type !== AST_NODE_TYPES.Identifier) {
           context.report({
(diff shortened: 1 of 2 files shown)


src/rules/no-done-callback.ts
```diff
@@ -41,17 +50,21 @@ export default createRule({
   create(context) {
     return {
       CallExpression(node) {
-        const callback = findCallbackArg(node);
+        // done is the second argument for it.each, not the first
+        const isJestEach = !!getNodeName(node.callee)?.endsWith('.each');
+
+        const callback = findCallbackArg(node, isJestEach);
+        const callbackArgIndex = +isJestEach;
```
G-Rath, line 57:
```suggestion
        const callbackArgIndex = Number(isJestEach);
```

src/rules/no-done-callback.ts
```diff
@@ -2,11 +2,20 @@ import {
...
+  isFunction,
+  isHook,
+  isTestCase,
+} from './utils';
 
 const findCallbackArg = (
   node: TSESTree.CallExpression,
+  isJestEach: boolean,
 ): TSESTree.CallExpression['arguments'][0] | null => {
+  if (isJestEach) return node.arguments[1];
```
G-Rath, line 17:
```suggestion
  if (isJestEach) {
    return node.arguments[1];
  }
```

G-Rath (approved):

src/rules/no-done-callback.ts
```diff
@@ -41,17 +52,21 @@ export default createRule({
   create(context) {
     return {
       CallExpression(node) {
-        const callback = findCallbackArg(node);
+        // done is the second argument for it.each, not the first
+        const isJestEach = !!getNodeName(node.callee)?.endsWith('.each');
```
SimenB, line 56:
```suggestion
        const isJestEach = getNodeName(node.callee)?.endsWith('.each') != null;
```

G-Rath: Don't we have `eqeqeq` on? 🤔

k-yle: `.endsWith('.each')` returns `boolean`, I'm not sure if this has the same effect

G-Rath:
I mucked up my commit, so it needs to be reworded 🤦‍♂️

k-yle:
@G-Rath shall I amend that commit? I don't see what's wrong with the commit message though

src/rules/no-done-callback.ts
SimenB:
```suggestion
        const isJestEach = getNodeName(node.callee)?.endsWith('.each') ?? false;
```

G-Rath:
@k-yle yes please - it's the capital letter, changing it to `apply` should fix it :)

src/rules/no-done-callback.ts
SimenB: > Don't we have `eqeqeq` on? 🤔

I assume we allow "smart" equality which has exceptions for `== null`. regardless, that was the wrong thing 😀

voda:
@k-yle I found another false positive, reported in #710

Issue #710, opened 2020-11-12
test.each no-done-callback false positive in version 24.1.2


voda:
Very similar to #709, but there is still a false positive when the test function has 2 arguments.

Plugin version: 24.1.2
Last working version: 24.1.0
Rule jest/no-done-callback has false positives with test.each

```ts
it.each([
	[1, 2],
	[2, 4],
	[3, 6],
])('.double(%i)', async (a: number, expected: number) => {
	expect(2 * a).toBe(expected);
});
```

jsaguet:
I just had the same issue.

After having a quick look at the rule implementation, it seems the logic is to check the number of arguments of the test function.

But for `test.each` / `it.each`, just checking the number or arguments of the function will never work.
I think the only way to enforce this rule on `test.each` is to check if the last argument is actually called in the function block statement.

voda:
> I think the only way to enforce this rule in this case is to check if the last argument is actually called in the function block statement.

The list of test cases can also contain a callback, e.g.:
```js
it.each([
	[1, () => 2],
])(...);
```

I think jest is also checking number of arguments: https://github.com/facebook/jest/blob/f9132f9106e3f1b3ae6f81d1f36222055c3fe111/packages/jest-each/src/bind.ts#L81-L83

PR #722, opened 2020-11-27
fix: proper support for it.each 

k-yle:
retrying #701, this time without the regression for `no-done-callback` 🙊

`no-done-callback` now works for jest-each's tagged-template syntax, but intentionally ignores it.each's table/array syntax because I have no idea how to handle [this edge case](https://github.com/jest-community/eslint-plugin-jest/issues/710#issuecomment-726081134).

### DIFF
--- a/src/rules/no-disabled-tests.ts
+++ b/src/rules/no-disabled-tests.ts
@@ -39,6 +39,9 @@ export default createRule({
       CallExpression(node) {
         const functionName = getNodeName(node.callee);
 
+        // prevent duplicate warnings for it.each()()
+        if (node.callee.type === 'CallExpression') return;
+
         switch (functionName) {
           case 'describe.skip':
             context.report({ messageId: 'skippedTestSuite', node });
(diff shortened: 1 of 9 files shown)


src/rules/no-disabled-tests.ts
```diff
@@ -39,6 +39,9 @@ export default createRule({
       CallExpression(node) {
         const functionName = getNodeName(node.callee);
 
+        // prevent duplicate warnings for it.each()()
+        if (node.callee.type === 'CallExpression') return;
```
G-Rath, line 43:
```suggestion
        if (node.callee.type === 'CallExpression') {
          return;
        }
```

G-Rath (approved):

SimenB:
@k-yle is this still draft?

k-yle:
sorry, it's ready @SimenB 

I thought someone else might try to make it work for this edge case:  https://github.com/jest-community/eslint-plugin-jest/issues/710#issuecomment-726047708 but it seems to be very hard

SimenB:
We can do that later 👍

G-Rath (approved):
sorry for the late review - looks good to me, lets ship it!

PR #726, opened 2020-12-07
docs(readme): remove outdated renovate badge

dargmuesli:
Currently there is no official renovate badge and the one used here was pointing to an example endpoint that got removed.


dargmuesli:
Please explain to me why a commit body's lines must not be longer than 100 characters.

G-Rath:
@dargmuesli because it keeps things nicely formatted for viewing in terminals, which are traditionally ~80-100 characters.

dargmuesli:
Well well, it'd be cool if the terminal showed just the first line and this rule would just apply to that, allowing additional information to not having to be truncated. I don't know enough to say if this could work, I just want to point out the obvious :D
here you go^^

SimenB:
Thanks!

PR #728, opened 2020-12-15
Update prefer-expect-assertions.md

hakubo:
expect.assertions accepts a number as argument


G-Rath:
That's the point - this rule checks that you're passing a number to `expect.assertions`; what you've changed are examples of *invalid* arguments being passed to `expect.assertions` that it will catch :)

hakubo:
🤦

Issue #112, opened 2018-05-03
[new rule] no jest.clearAllMocks


dylang:
When using the jest config option [clearMocks: true](https://facebook.github.io/jest/docs/en/configuration.html#clearmocks-boolean), mocks are cleared automatically before each test. This means using `jest.clearAllMocks();` is not necessary, so it can be removed.

SimenB:
Sure, we could probably have a rule for this. The same applies for a few other things, which should be covered at the same time: https://github.com/facebook/jest/blob/6fe69d1467c329f872ab85b7325e7908796b35d0/packages/jest-jasmine2/src/index.js#L79-L99

macklinu:
> The same applies for a few other things, which should be covered at the same time

This seems like it could fall into one rule that would report unnecessary function calls based on Jest configuration (something like `no-unnecessary-configuration`, but not sure what a good name would be).

Would users opt-in by including the functions they want to report in their ESLint config?

```json
{
  "rules": {
    "jest/no-unnecessary-config": ["error", { "reportOn": ["clearMocks"] }]
  }
}
```

Is this something that we could read the Jest config in the local repo and automatically figure out which functions to report if this rule is enabled?

Just throwing out ideas here for implementation and naming to get the ball rolling - seems like a useful idea for a rule. 👍

G-Rath:
I'm going to close this issue because the specific request of disabling calls to
`jest.clearAllMocks` can be enforced on projects using the built-in
[`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax)
rule:

```json
{
  "rules": {
    "no-restricted-syntax": [
      "error",
      {
        "selector": "CallExpression[callee.object.name='jest'][callee.property.name='clearAllMocks']",
        "message": "`clearMocks` has been enabled for tests, so is already being called automatically by Jest`"
      }
    ]
  }
}
```

We do hope to one day have a "no-unnecessary-config" that leverages the jest
config to be able to know which files have what config options enabled, but that
cannot happen without the ability for rules to be async which ESLint currently
does not support. It's expected that this _will_ eventually be possible since
it's a requirement of ESM, but there isn't a clear timeline so could be a few
years out.

For now, I'm labeling these issues with "config support" to allow us to easily
look them up in the future when we're unblocked.

PR #754, opened 2021-02-04
feat(no-unnecessary-functions): create rule

AndrewLeedham:
Creates a new rule called "no-unnecessary-functions", which prevents calls to `jest.resetModules`, `jest.clearAllMocks`, `jest.resetAllMocks` and `jest.restoreAllMocks`

fixes: #112


G-Rath:
Thanks for the contribution!

I don't think this is worth shipping in `jest` as its own rule as-is, as you can do it easily with [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax), i.e.:

```
{
  "selector": "CallExpression[callee.name='resetModules']",
  "message": "Do not use resetModules"
}
```

AndrewLeedham:
> Thanks for the contribution!
> 
> 
> 
> I don't think this is worth shipping in `jest` as its own rule as-is, as you can do it easily with [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax), i.e.:
> 
> 
> 
> ```
> 
> {
> 
>   "selector": "CallExpression[callee.name='resetModules']",
> 
>   "message": "Do not use resetModules"
> 
> }
> 
> ```

Fair point, perhaps it would be more use if it automatically detected which are unnecessary from the users jest config? Are there any other rules that use the jest config?

G-Rath:
Sorry for the late reply. You've hit the nail on the head regarding the jest config -  we need to figure out a way to get the config that's stable, performant, and secure.

There's a few ways we could do it, but overall it's a hard challenge as the config can come from a number of places: from `package.json`, depending on paths, if it's javascript then it could import things, if its typescript it needs to be transpiled, etc

AndrewLeedham:
Perhaps we could utilise https://github.com/facebook/jest/blob/master/packages/jest-config/src/index.ts#L33. Although I am not 100% sure how to determine which project relates to which file, perhaps there is another of Jest's packages that could help us there. Re-implementing that behaviour would likely be un-stable, so I would have thought improving what Jest exposes would be a more robust route if needs be.

G-Rath:
Yeah, there's a lot of edge-cases to cover, which is why we've not done it yet - iirc there was talk about possibly calling the jest cli, but even that has edge cases 🤷

SimenB:
Should definitely use `jestConfig.readConfig`. But there's (at least) 3 issues with that currently

1. It's async, so we cannot do it at all (`eslint@7` introduced some async APIs (when using it programatically see https://eslint.org/blog/2020/05/eslint-v7.0.0-released#new-eslint-class), but I don't think rules can use it)
1. As mentioned, figuring out which config (project) applies to which file is buried in Jest's source code and would need to be surfaced somehow from it (it's a combination of include and ignore patterns in non-obvious ways)
1. We'd need to get it once and cache it for the rest of the rules that might want it. Might work with a util that caches it locally though

AndrewLeedham:
1. Would it be plausible to provide a `jestConfig.readConfig.sync` function for this purpose?
2. Is there a ticket open for this on Jest or does one need creating?
3. 👍

SimenB:
>Would it be plausible to provide a `jestConfig.readConfig.sync` function for this purpose?

No, as we need to use `import()` to load ESM config, and that is asynchronous (by specification, as mentioned)

>Is there a ticket open for this on Jest or does one need creating?

Feel free to create one! Or send a PR 🙂 Note that due to 1 we are blocked, but having the APIs ready from Jest's side would be nice so we only need to wait for ESLint

AndrewLeedham:
> No, as we need to use import() to load ESM config, and that is asynchronous (by specification, as mentioned)

Ah because a Jest config can be ESM. I see what you mean now 👍 

> Feel free to create one! Or send a PR 🙂 Note that due to 1 we are blocked, but having the APIs ready from Jest's side would be nice so we only need to wait for ESLint

I am not sure where this logic currently belongs, or really how it works, but I have raised an issue for now: https://github.com/facebook/jest/issues/11171

G-Rath:
@AndrewLeedham are you ok if I close this to keep things tidy?

PR #765, opened 2021-02-20
feat(unbound-method): create rule

G-Rath:
This is an initial attempt at creating an "jest aware" version of [`@typescript-eslint/unbound-method`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md) that won't report on passing "unbound" methods to `expect`.

For now I've copied the rule and its tests (including fixtures) over verbatim and then kept modifications to a minimum so that it's easier to keep up to date with the base rule. (ideally would be nice if we could watch the base rule for changing some how 🤔).

It turns out that the logic is actually pretty straightforward, at least apparently to catch the common cases, but for the same reason I suspect I've missed an edge-case or two.

Aside from "what edge cases have I missed", my primary question right now is what should we do if we can't get the type service?

Personally, I'd really like to be able to enable this rule without having to worry about if the TypeScript service is available, as that'll make it a lot easier for sharable configs. However, the default behaviour is to throw saying that you need type information and its possible that people could be confused that the rule isn't do anything (however I think this'll be less likely for us, as the rule implies some exposure to the base rule and therefore TypeScript).

If we think the rule should error when type information is not available, I think it would be good to add an option to allow choosing between throwing an error or silently doing nothing, which could be used in shareable configs to safely enable the rule without having to worry.

I've created a test that hopefully captures the behaviour of the rule when running without the TypeScript service, but I could be completely wrong on that.

(I have just realised the rule actually imports `ts` for some enum values, so will have a think about this - we could drop that dependency easily by just inlining the enums and making it a `type` import 🤔)

@bradzacher would be good to have your input :)

Closes https://github.com/typescript-eslint/typescript-eslint/issues/2951


G-Rath:
> I.e. it could be taught that expect(clazz.method).toBeTruthy() is safe, but expect(clazz.method).toThrow() is unsafe.

(will implement this now)

bradzacher (commented):
if the only change you're making it to make this not report for expect calls, you can simplify this as an extension rule.

```ts
import TSESLintPlugin from '@typescript-eslint/eslint-plugin';
const baseRule = TSESLintPlugin.rules['unbound-method'];

// ...

export default util.createRule<Options, MessageIds>({
  ...baseRule,
  create(context) {
    const baseSelectors = baseRule.create(context);
    return {
      ...baseSelectors,
      CallExpression(node: TSESTree.CallExpression): void {
        inExpectCall = isJestExpectCall(node);
      },
      'CallExpression:exit'(node: TSESTree.CallExpression): void {
        if (inExpectCall && isJestExpectCall(node)) {
          inExpectCall = false;
        }
      },
      MemberExpression(node: TSESTree.MemberExpression): void {
        if (inExpectCall) {
          return;
        }
        baseRule.MemberExpression(node);
      },
    };
  },
})
```

Note that: this does create a dep on ts-eslint plugin.

You could make this safer and not add a direct dependency on the plugin with something like:

```ts
const baseRule = (() => {
  try {
    const TSESLintPlugin = require('@typescript-eslint/eslint-plugin');
    return TSESLintPlugin.rules['unbound-method'];
  } catch (e) {
    return null;
  }
})();

export default util.createRule<Options, MessageIds>({
  ...baseRule ?? { /* some defaults */ },
  create(context) {
    if (!baseRule) { return {} }
    return { ... };
  },
})
```

G-Rath:
> you can simplify this as an extension rule

That's a really good point! It's a lot cleaner to do it that way (thanks for the snippets btw).

I'd be happy with doing it that way, using the `try/catch` to avoid the hard dependency and make it an optional peer dependency to keep package managers happy.

G-Rath:
ugh while I like the idea, re-writing this as an extension rule is proving hard to test because I can't re-require the actual rule to cover it's behaviour when the package "is not there" (aka throws an error):

```
describe('when @typescript-eslint/eslint-plugin is not available', () => {
  let rule = require('../unbound-method').default;

  beforeEach(() => {
    jest.resetModules();
    jest.mock('@typescript-eslint/eslint-plugin', () => {
      console.log('require');
      return () => {
        throw new Error();
      };
    });
    rule = require('../unbound-method').default;
  });

  new ESLintUtils.RuleTester({
    parser: '@typescript-eslint/parser',
    parserOptions: {
      sourceType: 'module',
      tsconfigRootDir: rootPath,
      project: './tsconfig.json',
    },
  }).run('unbound-method jest edition without type service', rule, {
    // valid: validTestCases.concat(invalidTestCases.map(({ code }) => code)),
    valid: invalidTestCases.map(({ code }) => code),
    invalid: [],
  });
});
```

This is what I've been hacking together, which fails because you can't have `RuleTester` in an `it` (as it creates `it`s from the given test cases), but beforeEach` runs for each `it` statement - so by the time the tester runs, the require has already happened :/

@SimenB what are you thoughts on what we should cover for this testing wise?

G-Rath:
(in classic fashion, 5 minutes after my last comment, I glanced at some of the test code for [`eslint-plugin-eslint-config`](https://www.npmjs.com/package/eslint-plugin-eslint-config), which has similar tests, and realised how I should be able to test this 🤦)

G-Rath:
I've left the original fixtures + tests in there for now, but I've gotten full cover with minimal hacking and without blowing out the run time of the test suite.

The only issue is that apparently there's improper teardown going on:

> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.

I think maybe the TestRunner is staying alive longer than it should because the rule is erroring? Not sure why though 🙁

G-Rath:
@SimenB would be keen to hear your thoughts on this - aside from the doc tweaks, this rule should be good to go.

I've also got a few ideas on some more type-based rules, one of which I've already got a semi-working implementation for (checking that `.toThrow` & co matchers are passed a function to be called).

src/rules/unbound-method.ts
```diff
@@ -0,0 +1,101 @@
...
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const TSESLintPlugin = require('@typescript-eslint/eslint-plugin');
+
+    return TSESLintPlugin.rules['unbound-method'] as TSESLint.RuleModule<
+      MessageIds,
+      Options
+    >;
+  } catch (e) {
+    return null;
```
SimenB, line 31: you should check `e.code` and re-throw if it's not `MODULE_NOT_FOUND`

src/rules/unbound-method.ts
```diff
@@ -0,0 +1,101 @@
...
+];
+
+const isJestExpectCall = (node: TSESTree.CallExpression) => {
+  if (!isExpectCall(node)) {
+    return false;
+  }
+
+  const { matcher } = parseExpectCall(node);
+
+  return !toThrowMatchers.includes(matcher?.name ?? '');
```
SimenB, line 18: `'foobar'.includes('')` is `true`. Can we get the name earlier then bail rather than calling `includes`?

src/rules/unbound-method.ts
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const baseRule = (() => {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const TSESLintPlugin = require('@typescript-eslint/eslint-plugin');
```
SimenB, line 24: we need to add an optional peer dependency for PnP to work (and possibly npm 7 without legacy flags)

SimenB (approved):
I'm a fan of type based rules 👍 I don't think they should go in recommended, but we can have a different rule set with type based ones

G-Rath:
Awesome! I'll look to continue going 🎉

I'm updating the docs now, and then will address your changes - hopefully it should be out of draft sometime today.

src/rules/unbound-method.ts
G-Rath: I agree - do you have a preference on `!matcher ||` vs `if`?

G-Rath:
```
    /home/runner/work/eslint-plugin-jest/eslint-plugin-jest/src/rules/__tests__/unbound-method.test.ts:10
          code = 'MODULE_NOT_FOUND';
               ^

    SyntaxError: Unexpected token =

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
```

Interesting - seems our transpiling setup doesn't support class properties?

It's erroring on this:

```
    throw new (class extends Error {
      public code = 'MODULE_NOT_FOUND';
    })();
```

G-Rath:
ofph

```
g-rath@DESKTOP-MUGH739:/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest$ nrun build
> nrun.js build
> babel --extensions .js,.ts src --out-dir lib --copy-files && rimraf lib/__tests__ 'lib/**/__tests__'
SyntaxError: /c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/src/rules/__tests__/fixtures/indent/indent-invalid-fixture-1.js: Identifier 'c' has already been declared (180:9)

  178 | };
  179 |
> 180 | function c(a, b) {
      |          ^
  181 |   if (a || (a &&
  182 |             b)) { // NO ERROR: DON'T VALIDATE MULTILINE STATEMENTS
  183 |     return d;
    at Parser._raise (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:776:17)
    at Parser.raiseWithData (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:769:17)
    at Parser.raise (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:737:17)
    at ScopeHandler.checkRedeclarationInScope (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:1392:12)
    at ScopeHandler.declareName (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:1358:12)
    at Parser.registerFunctionStatementId (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:12587:16)
    at Parser.parseFunction (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:12567:12)
    at Parser.parseFunctionStatement (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:12194:17)
    at Parser.parseStatementContent (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:11884:21)
    at Parser.parseStatement (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/@babel/parser/lib/index.js:11846:17) {
  loc: Position { line: 180, column: 9 },
  pos: 2320,
  code: 'BABEL_PARSE_ERROR'
}
```

babel.config.js
```diff
@@ -7,4 +7,5 @@ module.exports = {
     '@babel/preset-typescript',
     ['@babel/preset-env', { targets: { node: 10 } }],
   ],
+  ignore: ['src/**/__tests__/**'],
```
G-Rath, line 10: This change means we're releasing ~1 second faster, since we're now only compiling ~48 files instead of ~90 😂

G-Rath:
I think we're good to go - @SimenB I'll leave this around for a bit in case you want to review the docs and changes; feel free to merge if you're happy and want to get it out, otherwise I'll look to do so sometime today or tomorrow :)

src/rules/unbound-method.ts
SimenB:
```
if (!matcher?.name) {
  // or true, whatever
  return false;
}

return !toThrowMatchers.includes(matcher.name);

File: README.md [hunk 5]
```diff
@@ -173,7 +173,29 @@ installations requiring long-term consistency.
 | [valid-expect-in-promise](docs/rules/valid-expect-in-promise.md)             | Enforce having return statement when testing with promises      | ![recommended][] |              |
 | [valid-title](docs/rules/valid-title.md)                                     | Enforce valid titles                                            | ![recommended][] | ![fixable][] |
 
-<!-- end rules list -->
+<!-- end base rules list -->
+
+## TypeScript Rules
```
SimenB, line 178: something about needing to install the upstream plugin as well? Possibly in the table if we'll have othe rrules in the future that do _not_ use type info

docs/rules/unbound-method.md
```diff
@@ -0,0 +1,53 @@
+# Enforces unbound methods are called with their expected scope (`unbound-method`)
```
SimenB, line 1: this needs to mention that the user needs to depend on `@typescript-eslint/eslint-plugin`

src/rules/__tests__/fixtures/tsconfig-withmeta.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "emitDecoratorMetadata": true,
+    }
+}
```
SimenB, line 6:
```suggestion
}

```

or something 😅 add a trailing newline

src/rules/unbound-method.ts
```diff
@@ -0,0 +1,111 @@
...
+  const { matcher } = parseExpectCall(node);
+
+  if (!matcher) {
+    return false;
+  }
+
+  return !toThrowMatchers.includes(matcher.name);
+};
+
+const baseRule = (() => {
```
SimenB, line 25:
```suggestion
const baseRule: TSESLint.RuleModule<
      MessageIds,
      Options
    > | null = (() => {
```

rather than casting

src/rules/unbound-method.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    throw error;
+  }
+})();
+
+const tryCreateBaseRule = (
+  context: Readonly<TSESLint.RuleContext<MessageIds, Options>>,
+) => {
+  try {
+    return baseRule?.create(context);
```
SimenB, line 49: what errors can this throw?

SimenB (approved):

src/rules/unbound-method.ts
G-Rath: I don't know, but figured it'd be best to play it safe.

The only error I expect it will throw is that if you don't provide `project`, it'll throw an error about not having the type checking service.

src/rules/unbound-method.ts
G-Rath: I'm happy to add the return type, but would prefer to keep the cast as it's narrowing an `any`, which otherwise is [an unsafe return](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-return.md)

README.md
G-Rath: >  Possibly in the table if we'll have othe rrules in the future that do not use type info

You mean if we have other rules that don't extend base rules.

-------

You're completely right - I don't expect us to make any more rules that extend base rules, only more rules that use type information; so I might leave the table as is for now and just add a sentence saying that `unbound-method` is an extension rule (as well as updating its own docs).

G-Rath:
Aright, it's been a week - lets ship it!

Issue #775, opened 2021-03-06
False-positive jest/no-disabled-tests


fabb:
In a non-test file (trackingXiti.ts), I have a function like this:

```
const mapPageViewEventName = (...) => {
    ....
    const map = xitiViewMap(...)
    ...
}
```

The rule warns about the line with the `xitiViewMap` function, probably because it starts with `xit`.

1. Shouldn't the jest rules only run on `*.test.{js,tsx?}` files?
2. The rule should match `xit(` and not `xitSomethingElse(`

G-Rath:
> Shouldn't the jest rules only run on *.test.{js,tsx?} files?

No, because `jest` doesn't require you to name your test files like that - you should be configuring `eslint` to only apply `eslint-plugin-jest` to your test files using [overrides](https://eslint.org/docs/rules/keyword-spacing#overrides).

> The rule should match xit( and not xitSomethingElse(

Correct - it's because our matcher is incorrect, I'll have a PR up shortly.

fabb:
Would it be possible that the plugin reads jest.config.js to know which files it should consider?

G-Rath:
We do hope to read the config one day to enhance the rules - [see here](https://github.com/jest-community/eslint-plugin-jest/pull/754#issuecomment-780174615) for why that's currently not possible.

However, I don't think we'd do that for determining what files to run on, as that is exactly what you're meant to be using `overrides` for and some of our rules don't require being rule on files that jest runs as tests (i.e test helpers).

fabb:
I‘m not completely sure how I should use overrides here, currently I‘m just using an `/* eslint-disable jest/no-disabled-tests */` rule at the top of the (non-test) file. Would be nice If I could somehow generally declare which rules to enable for which files without having to specify it in the files themselves.

G-Rath:
> Would be nice If I could somehow generally declare which rules to enable for which files without having to specify it in the files themselves.

That's what `overrides` is for :)

[Here's one of my projects that uses `overrides` to apply the `eslint-plugin-jest` rules only to test files](https://github.com/G-Rath/audit-app/blob/main/.eslintrc.js) (via applying my company's `jest` shared config).

fabb:
Ah nice, thanks! The first link above to overrides confused me.

G-Rath:
[...]

PR #777, opened 2021-03-06
fix(no-disabled-tests): adjust selector to match only test functions

G-Rath:
The missing brackets means we're currently checking for `(^xit) OR (xtest$)`


SimenB (approved):

G-Rath:
This fixed #775

fabb:
Ouch, regex operator precendence is weird sometimes.

Issue #776, opened 2021-03-06
Autofix for xdescribe.each is wrong


fabb:
Autofix for `xdescribe.each` replaces by `describe.each.skip` which does not exist. It should be `describe.skip.each`.

https://jestjs.io/docs/uk/api#describeskipeachtablename-fn

fabb:
And `describe.skip.each` does not trigger the rule `jest/no-disabled-tests`.

G-Rath:
The issue with `no-test-prefixes` has been fixed by #722

PR #778, opened 2021-03-06
fix(no-disabled-tests): support `describe.skip.each` & `xdescribe.each`

G-Rath:
#722 makes this just a matter of adding the extra `case`s 🎉 

Relates to #776


SimenB (approved):
nice 👍

PR #780, opened 2021-03-06
test(no-test-prefixes): add case for `xdescribe.each`

### LINKED ISSUE #776: Autofix for xdescribe.each is wrong

G-Rath:
#722 fixed the first half of #776 - this is just adding an explicit test case for this situation.

Fixes #776

### DIFF
--- a/src/rules/__tests__/no-test-prefixes.test.ts
+++ b/src/rules/__tests__/no-test-prefixes.test.ts
@@ -47,6 +47,18 @@ ruleTester.run('no-test-prefixes', rule, {
       ],
     },
     {
+      code: 'xdescribe.each([])("foo", function () {})',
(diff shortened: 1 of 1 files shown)


SimenB (approved):
does `test` trigger a release?

G-Rath:
No it shouldn't - semantic-release should only trigger on `fix`, `feat`, and changes marked as breaking:

> [The commit analyzer that ships with semantic-release only catches fix and feat commits (#12).](https://www.npmjs.com/package/semantic-release-conventional-commits)

SimenB:
good stuff 👍

PR #782, opened 2021-03-07
fix(consistent-test-it): false positives inside describe.each

idan-at:
I think [this](https://github.com/jest-community/eslint-plugin-jest/pull/722) PR added false positive ESLing errors where `test` inside `describe.each` was actually thought to be `it` inside `describe.each`. (and the other way around) - see failing tests.

The bug exists on the last release (`24.1.7`) but was not present before.

The fix was in the `isDescribe` and `isTestCase` helpers, which ignored the `.each` version with templates.

PR #783, opened 2021-03-07
ci: don't run git hooks when in ci

G-Rath:
Husky now doesn't handle checking if you're in CI, so we have to ourselves.

This prevented #782 from going out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "test": "jest",
     "tools:regenerate-docs": "ts-node -T tools/regenerate-docs",
     "typecheck": "tsc -p .",
-    "postinstall": "husky install",
+    "postinstall": "is-ci || husky install",
     "prepublishOnly": "pinst --disable",
     "postpublish": "pinst --enable"
   },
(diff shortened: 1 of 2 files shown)


SimenB:
I think skipping the install is better

SimenB:
Semantic release should skip hooks when they commit, but that's a different issue 😅

package.json
```diff
@@ -78,7 +78,8 @@
     ]
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^4.0.1"
+    "@typescript-eslint/experimental-utils": "^4.0.1",
+    "is-ci": "^3.0.0"
```
SimenB, line 82: dev dep

SimenB: (`pinst` deals with it so it's not actually published as a postinstall)

G-Rath: (it'd be nice if `yarn` v2 supported `prepare` 😞)

SimenB (approved):

PR #785, opened 2021-03-08
fix(valid-describe): false positive with template describe.each

idan-at:
Due to [my previous fix](https://github.com/jest-community/eslint-plugin-jest/pull/782), now `valid-describe` has false positives 😆


SimenB (approved):
thanks!

PR #787, opened 2021-03-08
feat(no-focused-tests): make fixable

alagane:
Already marked as fixable in the README.md

### DIFF
--- a/src/rules/no-focused-tests.ts
+++ b/src/rules/no-focused-tests.ts
@@ -50,11 +50,12 @@ export default createRule({
       category: 'Best Practices',
       description: 'Disallow focused tests',
       recommended: 'error',
+      suggestion: true,
     },
     messages: {
       focusedTest: 'Unexpected focused test.',
+      suggestRemoveFocus: 'Remove focus from test',
     },
-    fixable: 'code',
     schema: [],
     type: 'suggestion',
   },
@@ -71,7 +72,23 @@ export default createRule({
           callee.object.type === AST_NODE_TYPES.Identifier &&
           isCallToFocusedTestFunction(callee.object)
         ) {
-          context.report({ messageId: 'focusedTest', node: callee.object });
+          context.report({
+            messageId: 'focusedTest',
+            node: callee.object,
+            suggest: [
+              {
+                messageId: 'suggestRemoveFocus',
+                fix(fixer) {
+                  return [
+                    fixer.removeRange([
+                      callee.object.range[0],
+                      callee.object.range[0] + 1,
+                    ]),
+                  ];
+                },
+              },
+            ],
+          });
 
           return;
         }
@@ -80,16 +97,50 @@ export default createRule({
           callee.object.type === AST_NODE_TYPES.MemberExpression &&
           isCallToTestOnlyFunction(callee.object)
         ) {
+          const calleeObject: TSESTree.MemberExpression = callee.object;
+
           context.report({
             messageId: 'focusedTest',
-            node: callee.object.property,
+            node: calleeObject.property,
+            suggest: [
+              {
+                messageId: 'suggestRemoveFocus',
+                fix(fixer) {
+                  return [
+                    fixer.removeRange(
+                      calleeObject.property.type ===
+                        AST_NODE_TYPES.Identifier &&
+                        calleeObject.property.name === 'only'
+                        ? [calleeObject.object.range[1], calleeObject.range[1]]
+                        : [calleeObject.range[1], callee.range[1]],
+                    ),
+                  ];
+                },
+              },
+            ],
           });
 
           return;
         }
 
         if (isCallToTestOnlyFunction(callee)) {
-          context.report({ messageId: 'focusedTest', node: callee.property });
+          context.report({
+            messageId: 'focusedTest',
+            node: callee.property,
+            suggest: [
+              {
+                messageId: 'suggestRemoveFocus',
+                fix(fixer) {
+                  return [
+                    fixer.removeRange([
+                      callee.object.range[1],
+                      callee.range[1],
+                    ]),
+                  ];
+                },
+              },
+            ],
+          });
 
           return;
         }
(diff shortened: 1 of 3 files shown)


src/rules/no-focused-tests.ts
```diff
@@ -71,7 +71,18 @@ export default createRule({
           callee.object.type === AST_NODE_TYPES.Identifier &&
           isCallToFocusedTestFunction(callee.object)
         ) {
-          context.report({ messageId: 'focusedTest', node: callee.object });
+          context.report({
+            messageId: 'focusedTest',
+            node: callee.object,
+            fix(fixer) {
```
SimenB, line 77: these should be suggestions as they might break user code

https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions

SimenB: yep - it's not a safe fix to automatically apply, but users can opt-in on a case by case basis

alagane: Then in the README.md, should it be suggest or fixable?

SimenB: for focused test and people having autofix on save this would mess with their workflow as eslint would remove the focus when they save the change to run the tests

SimenB: >Then in the README.md, should it be suggest or fixable?

suggest 🙂

alagane:
Ready to review!

SimenB:
remove `fixable: 'code'` from the rule, and I think the docs generator will agree with you

src/rules/no-focused-tests.ts
```diff
@@ -80,16 +98,50 @@ export default createRule({
...
+                fix(fixer) {
+                  return [
+                    fixer.removeRange(
+                      calleeObject.property.type ===
+                        AST_NODE_TYPES.Identifier &&
+                        calleeObject.property.name === 'only'
+                        ? [calleeObject.object.range[1], calleeObject.range[1]]
+                        : [calleeObject.range[1], callee.range[1]],
+                    ),
+                  ];
```
SimenB, line 118: this was a bit hard to follow

```suggestion
                  if (
                    calleeObject.property.type === AST_NODE_TYPES.Identifier &&
                    calleeObject.property.name === 'only'
                  ) {
                    return fixer.removeRange([
                      calleeObject.object.range[1],
                      calleeObject.range[1],
                    ]);
                  }

                  return fixer.removeRange([
                    calleeObject.range[1],
                    callee.range[1],
                  ]);
```

src/rules/no-focused-tests.ts
```diff
@@ -71,7 +72,23 @@ export default createRule({
...
         ) {
-          context.report({ messageId: 'focusedTest', node: callee.object });
+          context.report({
+            messageId: 'focusedTest',
+            node: callee.object,
+            suggest: [
+              {
+                messageId: 'suggestRemoveFocus',
+                fix(fixer) {
+                  return [
```
SimenB, line 82: no need to return an array when there's only one fix applied

(applies to all of the added fixers I believe)

SimenB (changes requested):

SimenB:
@alagane fyi we'll squash merge when landing, so no need to force push (it's easier to review without force pushes as we can see the changes between each commit. _this_ PR is of course small enough that it doesn't really matter, so keep doing force pushes if it's the workflow you're most comfortable with)

src/rules/no-focused-tests.ts
```diff
@@ -50,11 +50,12 @@ export default createRule({
       category: 'Best Practices',
       description: 'Disallow focused tests',
       recommended: 'error',
+      suggestion: true,
     },
     messages: {
       focusedTest: 'Unexpected focused test.',
+      suggestRemoveFocus: 'Remove focus from test',
```
SimenB, line 57:
```suggestion
      suggestRemoveFocus: 'Remove focus from test.',
```

src/rules/no-focused-tests.ts
```diff
@@ -71,7 +72,23 @@ export default createRule({
...
+            suggest: [
+              {
+                messageId: 'suggestRemoveFocus',
+                fix(fixer) {
+                  return [
+                    fixer.removeRange([
+                      callee.object.range[0],
+                      callee.object.range[0] + 1,
+                    ]),
+                  ];
```
SimenB, line 87:
```suggestion
                  return fixer.removeRange([
                      callee.object.range[0],
                      callee.object.range[0] + 1,
                    ])
```

(prettier won't be happy, so run `yarn lint --fix` after applying all of these)

src/rules/no-focused-tests.ts
```diff
@@ -80,16 +97,55 @@ export default createRule({
...
+            suggest: [
+              {
+                messageId: 'suggestRemoveFocus',
+                fix(fixer) {
+                  return [
+                    fixer.removeRange([
+                      callee.object.range[1],
+                      callee.range[1],
+                    ]),
+                  ];
```
SimenB, line 144:
```suggestion
                  return fixer.removeRange([
                      callee.object.range[1],
                      callee.range[1],
                    ]);
```

src/rules/no-focused-tests.ts
```diff
@@ -99,7 +155,20 @@ export default createRule({
...
+        context.report({
+          messageId: 'focusedTest',
+          node: callee,
+          suggest: [
+            {
+              messageId: 'suggestRemoveFocus',
+              fix(fixer) {
+                return [
+                  fixer.removeRange([callee.range[0], callee.range[0] + 1]),
+                ];
```
SimenB, line 167:
```suggestion
                return fixer.removeRange([callee.range[0], callee.range[0] + 1])
```

SimenB (changes requested):

SimenB (approved):
thanks!

src/rules/no-focused-tests.ts
```diff
@@ -80,16 +95,53 @@ export default createRule({
           callee.object.type === AST_NODE_TYPES.MemberExpression &&
           isCallToTestOnlyFunction(callee.object)
         ) {
+          const calleeObject: TSESTree.MemberExpression = callee.object;
```
SimenB, line 98:
```suggestion
          const calleeObject = callee.object;
```

Is the type annotation needed?

SimenB: can probably move the assignment above the `if` and it should be narrowed properly?

alagane: Well if I do not use the `calleeObject` variable I have some type errors on properties used below, but it works without the type.

PR #788, opened 2021-03-09
fix(no-identical-titles): ignore .each template cases

### LINKED ISSUE #786: [no-identical-title] issue with describe.each
Hi,

This code will trigger a lint error:
```js
describe("my tests", () => {
  describe.each`
    description
    ${'a'}
  `("$description", () => {})

  describe.each`
    description
    ${'b'}
  `("$description", () => {})
});
```

because the `$description` string repeats on both.

I'd expect those cases to actually resolve the title to what will be used on runtime, and then search for duplications.

so for example in this case, we'll have [`a`, `b`] as [describeTitles](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-identical-title.ts#L15), and not `["$description", "$description"]`

idan-at:
Fixes #786

### DIFF
--- a/src/rules/__tests__/no-identical-title.test.ts
+++ b/src/rules/__tests__/no-identical-title.test.ts
@@ -1,4 +1,5 @@
 import { TSESLint } from '@typescript-eslint/experimental-utils';
+import dedent from 'dedent';
 import resolveFrom from 'resolve-from';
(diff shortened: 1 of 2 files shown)


SimenB (approved):
will create false negatives as noted in the issue, but that's better than false positives 👍

PR #792, opened 2021-03-14
fix: support all variations of `describe`, `it`, & `test`

G-Rath:
We've had a bit of flurry of activity in the last few weeks which has resulted in some of our technical debt coming to the surface, so this is (hopefully) the start of me addressing some of that.

There are two main goals with this refactor:
  * we're consistent by only matching if the given `node` is a *call* i.e it'll result in `jest` doing "something"
    * This is primarily important for `.each`, because of`.each()` - while that looks like a call (and technically is), it's not the "right" kind of call we're wanting to match with these methods.
  * we're matching all possible variations both in terms of properties and accessor style (dot property & square brackets)

Because in a perfect world I'd like this to be the last time I have to think twice about this when using these methods, I've implemented a set of tests using an inline eslint rule to ensure we're matching every possible variation of calls and not-calls for these methods.

The end result was `lowercase-name` losing a point of cover due to not having a test that covers when the names are not strings, and a few "valid" cases for `valid-describe` failing as they're now actually supported (🎉).

Because I'd like to land this as a refactor commit by itself for now, I've renamed the methods so that I can keep the old `isDescribe` around for `valid-describe` to use - I'll be refactoring it once this is landed to support the new guards.

It's also technically a more accurate name anyway 🤷 

-----

Because of the interlocking nature of things, I've picked this as a starting point and deliberately ignored other possible refactors as I think it'll be premature to do them in one lump go - this is why I'm not doing any deduplication on the methods (since they now actually share pretty much the exact same implementation) nor touched the types or other util functions.

My goal is to have this as a stable starting point that I can build other refactors & fixes up around :)

### DIFF
--- a/src/rules/__tests__/utils.test.ts
+++ b/src/rules/__tests__/utils.test.ts
@@ -1,0 +1,427 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
+import resolveFrom from 'resolve-from';
+import { createRule, isDescribeCall, isTestCaseCall } from '../utils';
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: resolveFrom(require.resolve('eslint'), 'espree'),
+  parserOptions: {
+    ecmaVersion: 2015,
+  },
+});
+
+const rule = createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Possible Errors',
+      description: 'Fake rule for testing AST guards',
+      recommended: false,
+    },
+    messages: {
+      foundDescribeCall: 'found a call to `describe`',
+      foundTestCaseCall: 'found a call to a test case',
+    },
+    schema: [],
+    type: 'problem',
+  },
+  defaultOptions: [],
+  create: context => ({
+    CallExpression(node) {
+      const messageId =
+        (isDescribeCall(node) && ('foundDescribeCall' as const)) ||
+        (isTestCaseCall(node) && ('foundTestCaseCall' as const));
+
+      if (messageId) {
+        context.report({
+          messageId,
+          node,
+        });
+      }
+    },
+  }),
+});
+
+ruleTester.run('nonexistent methods', rule, {
+  valid: [
+    'describe.something()',
+    'describe.me()',
+    'test.me()',
+    'it.fails()',
+    'context()',
+    'context.each``()',
+    'context.each()',
+    'describe.context()',
+    'describe.concurrent()()',
+    'describe.concurrent``()',
+    'describe.every``()',
+  ],
+  invalid: [],
+});
+
+ruleTester.run('it', rule, {
+  valid: [
+    'it["concurrent"]["skip"]',
+    'it["concurrent"].skip',
+    'it.concurrent["skip"]',
+    'it.concurrent.skip',
+
+    'it["concurrent"]["only"]',
+    'it["concurrent"].only',
+    'it.concurrent["only"]',
+    'it.concurrent.only',
+
+    'it["skip"]["each"]()',
+    'it["skip"].each()',
+    'it.skip["each"]()',
+    'it.skip.each()',
+
+    'it["skip"]["each"]``',
+    'it["skip"].each``',
+    'it.skip["each"]``',
+    'it.skip.each``',
+
+    'it["only"]["each"]()',
+    'it["only"].each()',
+    'it.only["each"]()',
+    'it.only.each()',
+
+    'it["only"]["each"]``',
+    'it["only"].each``',
+    'it.only["each"]``',
+    'it.only.each``',
+
+    'xit["each"]()',
+    'xit.each()',
+
+    'xit["each"]``',
+    'xit.each``',
+
+    'fit["each"]()',
+    'fit.each()',
+
+    'fit["each"]``',
+    'fit.each``',
+
+    'it["skip"]',
+    'it.skip',
+
+    'it["only"]',
+    'it.only',
+
+    'it["each"]()',
+    'it.each()',
+
+    'it["each"]``',
+    'it.each``',
+
+    'fit',
+    'xit',
+    'it',
+  ],
+  invalid: [
+    ...[
+      'it["concurrent"]["skip"]()',
+      'it["concurrent"].skip()',
+      'it.concurrent["skip"]()',
+      'it.concurrent.skip()',
+
+      'it["concurrent"]["only"]()',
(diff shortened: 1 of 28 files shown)


G-Rath:
(Have made this a draft PR for now, but tbh it's ready to go)

src/rules/utils.ts
```diff
@@ -648,33 +648,126 @@ export const getTestCallExpressionsFromDeclaredVariables = (
...
+    return node.callee.object.type === AST_NODE_TYPES.MemberExpression
+      ? isTestCaseName(node.callee.object.object)
+      : isTestCaseName(node.callee.object);
+  }
+
+  if (
+    node.callee.type === AST_NODE_TYPES.TaggedTemplateExpression &&
+    node.callee.tag.type === AST_NODE_TYPES.MemberExpression
+  ) {
+    return node.callee.tag.object.type === AST_NODE_TYPES.MemberExpression
```
G-Rath, line 707: I'm *sure* there's a way to restructure this so we don't have to do it twice, but can't figure it out.

I'm also sure that the same restructure will be usable in `no-focused-tests` (as it has pretty much the exact same conditional structure)

src/rules/utils.ts
```diff
@@ -648,33 +648,126 @@ export const getTestCallExpressionsFromDeclaredVariables = (
...
+      ? isTestCaseName(node.callee.tag.object.object)
+      : isTestCaseName(node.callee.tag.object);
```
G-Rath, line 709: I'm *sure* there's a way to restructure this so we don't have to do it twice, but can't figure it out.

I'm also sure that the same restructure will be usable in `no-focused-tests` (as it has pretty much the exact same conditional structure)

G-Rath:
Just realised I should have tests covering *not* supported methods, i.e `test.items()()`, ```describe.every``()```, etc

src/rules/utils.ts
```diff
@@ -648,33 +648,120 @@ export const getTestCallExpressionsFromDeclaredVariables = (
...
+      : node.callee;
+
+  if (
+    callee.type === AST_NODE_TYPES.MemberExpression &&
+    isTestCaseProperty(callee.property)
+  ) {
+    // if we're an `each()`, ensure we're being called (i.e `.each()()`)
+    if (
+      node.callee.type !== AST_NODE_TYPES.TaggedTemplateExpression &&
+      node.parent?.type !== AST_NODE_TYPES.CallExpression &&
```
SimenB, line 698: there's no issue with `?` potentially making this undefined at the check `true`?

src/rules/__tests__/utils.test.ts
```diff
@@ -0,0 +1,424 @@
+import { TSESLint } from '@typescript-eslint/experimental-utils';
```
SimenB, line 1: great idea to have this test 👍  My only "complaint" is that it will potentially not help us discover dead code, but probably worth it

SimenB (approved):
great stuff 👍

src/rules/__tests__/utils.test.ts
G-Rath: I think we should still be able to find deadcode, just not in these methods. I plan to split the utils up into their own files under `utils/*` at some point, which'll also help.

src/rules/utils.ts
G-Rath: It shouldnt no - `parent` currently never can actually be `undefined` anyway, it's just marked as such because of TypeScript reasons.

There's https://github.com/typescript-eslint/typescript-eslint/issues/1417 which tracks wanting to improve the typings for `parent`.

G-Rath:
I think I might make this a `fix` instead of a `refactor`, and extend the scope a bit to include adjusting `valid-describe`.

There's already a few rules this technically fixes their support for `.each` implicitly, but I've just confirmed explicitly that this fixes `consistent-test-it`.

SimenB:
> There's already a few rules this technically fixes their support for `.each` implicitly, but I've just confirmed explicitly that this fixes `consistent-test-it`.

adding a few tests to the other rules to verify would be sweet

G-Rath:
#798 made me realise this doesn't address a big inconsistency that's not highlighted by the tests: the `CallExpression` that's matched for `.each()()` is not the same as the one matches for all the other cases, including ```.each``()```.

i.e with a rule that has a `create` like this:

```
create: context => ({
  CallExpression(node) {
    if (isTestCaseCall(node)) {
      console.log(node.arguments.length, getNodeName(node));
    }
  },
}),
```

with this source code:

```
it('is true, () => {});
it.only('is true', () => {});
it.each()('is true', () => {});
it.each``('is true', () => {});
```

You would get this output:

```
2 it
2 it.only
0 it.each
2 it.each
```

This is because `.each()()` actually is *two* call expressions, and our handling works from the "outer" CallExpression, which is `.each()`.

This is a really subtle difference that I really want to get rid of because it's something I don't want to have to think about - I think the better behaviour would be to match `.each()()` as the CallExpression we want - that should make it simpler in most situations when working with `.each` as it should mean you don't have to do `const args = isEachCall(node) ? node.parent.args : node.args`.

I think it should be straightforward to implement this behaviour, but will look to do that in a follow PR as I think this is nice by its own, and I could also be wrong 🤷

SimenB:
> I think it should be straightforward to implement this behaviour, but will look to do that in a follow PR as I think this is nice by its own, and I could also be wrong 🤷

👍

G-Rath:
@SimenB I think this is good to go - I've been through most of the rules to find which ones are improved by this refactor, and added a couple of test cases to each that fail with the old guards.

I've also refactored the utils test to have less duplication and be more unit-y - we can always make new tests if we come up with specific cases to test.

Part of the final refactor is also because I've written a patch for including testing `getNodeName`, as well as having implemented [this](https://github.com/jest-community/eslint-plugin-jest/pull/792#issuecomment-803498682) change, which I'll add in follow up PRs.

I'll merge this in a few minutes since I'd like to move on with more refactoring, but happy to discuss and revert/make changes if you've got any review comments :)

src/rules/__tests__/utils.test.ts
G-Rath: iirc `jest` doesn't support any kind of "don't collect coverage from <these> files/tests" ay?

Thinking about it, we should be able to rig CI to run this test specifically in its own job, and skip it when running the coverage-collecting tests

I had a think about this, and think we should be able to do something in CI to cover that?

PR #814, opened 2021-04-05
fix: improve handling of `.each` calls and with tagged literals

### LINKED ISSUE #809: `no-conditional-expect` does not handle `.each()()`
i.e

```
it.each()(foo', () => {
  something || expect(something).toHaveBeenCalled();
})
```

and other conditional variants don't trigger `no-conditional-expect`.

This'll be because the way we track if we're in a test case or not based on `CallExpression` nodes.

I expect it'll be fixed by the refactor I described [here](https://github.com/jest-community/eslint-plugin-jest/pull/792#issuecomment-803498682), which'll change the node we return `true` on.

### LINKED ISSUE #810: `no-if` does not handle `.each()()`
i.e

```
it.each()('foo', () => {
  callExpression()
  if ('bar') {}
})
```

doesn't trigger `no-if`

This'll be because the way we track if we're in a test case or not based on `CallExpression` nodes.

I expect it'll be fixed by the refactor I described [here](https://github.com/jest-community/eslint-plugin-jest/pull/792#issuecomment-803498682), which'll change the node we return `true` on.

### LINKED ISSUE #811: `no-test-return-statement` does not handle `.each()()`
i.e

```
it.each()("one", function () {
  return expect(1).toBe(1);
});
```

doesn't trigger `no-test-return-statement`

This'll be because the way we track if we're in a test case or not based on `CallExpression` nodes.

I expect it'll be fixed by the refactor I described [here](https://github.com/jest-community/eslint-plugin-jest/pull/792#issuecomment-803498682), which'll change the node we return `true` on.

G-Rath:
Implements the change described [here](https://github.com/jest-community/eslint-plugin-jest/pull/792#issuecomment-803498682), so that `isDescribeCall` and `isTestCaseCall` now return `true` for the outer `CallExpression` in `.each()()` calls, meaning matched nodes are now consistent as it'll always be the top-most one.

This has let us shed a bunch of edge-case code across our rules and utils, and also made it easy to refactor `no-focused-tests` so that it's 1/3rd smaller 🎉 

Because of the mix of dependent and independent changes, I'm going to open this PR as a draft, and cherry pick some of the changes out into their own PRs.

Resolves #809
Resolves #810
Resolves #811

### DIFF
--- a/src/rules/__tests__/lowercase-name.test.ts
+++ b/src/rules/__tests__/lowercase-name.test.ts
@@ -191,7 +191,7 @@ ruleTester.run('lowercase-name', rule, {
         {
           messageId: 'unexpectedLowercase',
           data: { method: TestCaseName.it },
-          column: 9,
+          column: 29,
           line: 1,
         },
       ],
@@ -203,7 +203,7 @@ ruleTester.run('lowercase-name', rule, {
         {
           messageId: 'unexpectedLowercase',
           data: { method: DescribeAlias.describe },
-          column: 15,
+          column: 35,
           line: 1,
         },
       ],
--- a/src/rules/__tests__/no-conditional-expect.test.ts
+++ b/src/rules/__tests__/no-conditional-expect.test.ts
@@ -153,6 +153,14 @@ ruleTester.run('logical conditions', rule, {
     },
     {
       code: `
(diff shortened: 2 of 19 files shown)


lion-starsirio (approved):
....

PR #819, opened 2021-04-10
fix(no-conditional-expect): check for expects in `catch`s on promises

### LINKED ISSUE #818: Should `no-conditional-expect` prevent Promise `catch` in tests?
We recently added `no-conditional-expect` to our app, and in trying to eliminate try / catch blocks, discovered that the following is possible. 

```
it('foo', async () => {
    await something
        .then(() => throw 'should have thrown an error')
        .catch((error) => {expect(error).toBe('error')})
})
```

Is that expected or is that a loophole and skirting the intentions of the rule?

G-Rath:
Fixes #818

### DIFF
--- a/docs/rules/no-conditional-expect.md
+++ b/docs/rules/no-conditional-expect.md
@@ -3,6 +3,9 @@
 This rule prevents the use of `expect` in conditional blocks, such as `if`s &
 `catch`s.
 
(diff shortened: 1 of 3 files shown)


manovotny:
Love it! ❤️  This looks great to me! 👍

PR #821, opened 2021-04-11
feat: create `prefer-to-be-for-literals`

### LINKED ISSUE #801: [new-rule] recommend `toBe` over `toEqual` when expecting primitive literals
As far as I know, there's nearly no difference in using `toBe` over `toEqual` when the expected value is a primitive literal, i.e

```
expect(response.statusCode).toBe(200);
// vs
expect(response.statusCode).toEqual(200);
expect(response.statusCode).toStrictEqual(200);

expect(isValid(obj)).toBe(true);
// vs
expect(isValid(obj)).toEqual(true);
expect(isValid(obj)).toStrictEqual(true);

expect(join(['hello', 'world')).toBe('hello world');
// vs
expect(join(['hello', 'world')).toEqual('hello world');
expect(join(['hello', 'world')).toStrictEqual('hello world');
```

IMO `toBe(200)` reads better than `toEqual(200)` (or `toStrictEqual(200)`).

It would be nice to have a rule that checked for `toEqual` matchers that were being passed primitive literal values, and recommended replacing them with `toBe`.

This would *not* apply if you're expecting an object or array literal, nor the value of a variable (even if we could resolve that variable back to a literal, i.e if its `const status = 200;`).

We also probably shouldn't be reporting on floats? (which we could create `prefer-to-be-close-to`) - alternatively I don't think it'd be much of a stretch if `prefer-to-be` supported floats.

<details>
<summary>my quick test setup to compare the three matchers</summary>

```
describe.each<[a: unknown, b: unknown]>([
  [1, 2],
  [1, 1],
  ['', ''],
  ['a', 'b'],
  ['a', 'a'],
  [{ x: 1 }, { x: 2 }],
  [{ x: 1 }, { x: 1 }],
  [{ x: { y: 1 } }, { x: { y: 2 } }],
  [{ x: { y: 1 } }, { x: { y: 1 } }],
  [[1], [2]],
  [[1], [1]],
  [[{ x: { y: 1 } }], [{ x: { y: 2 } }]]
])('%# %s vs %s', (a, b) => {
  test('toBe', () => {
    expect(a).toBe(b);
  });

  test('toEqual', () => {
    expect(a).toEqual(b);
  });

  test('toStrictEqual', () => {
    expect(a).toStrictEqual(b);
  });
});
```
</details>

G-Rath:
Resolves #801

I'm open to bikeshedding on the name:
  * `prefer-to-be` - we've already had a rule named this, but maybe thats ok?
  * `prefer-to-be-for-literals` - a little wordy, arguably a little inaccurate
  * `prefer-to-be-for-primitive-literal` - wordier, but the most accurate

I think this could be good to have in `styles` config when it comes time to do our next major.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ installations requiring long-term consistency.
 | [prefer-hooks-on-top](docs/rules/prefer-hooks-on-top.md)                     | Suggest having hooks before any test cases                      |                  |              |
 | [prefer-spy-on](docs/rules/prefer-spy-on.md)                                 | Suggest using `jest.spyOn()`                                    |                  | ![fixable][] |
 | [prefer-strict-equal](docs/rules/prefer-strict-equal.md)                     | Suggest using `toStrictEqual()`                                 |                  | ![suggest][] |
-| [prefer-to-be-for-literals](docs/rules/prefer-to-be-for-literals.md)         | Suggest using `tooBe()` for primitive literals                  |                  | ![fixable][] |
+| [prefer-to-be-for-literals](docs/rules/prefer-to-be-for-literals.md)         | Suggest using `toBe()` for primitive literals                   |                  | ![fixable][] |
 | [prefer-to-be-null](docs/rules/prefer-to-be-null.md)                         | Suggest using `toBeNull()`                                      | ![style][]       | ![fixable][] |
 | [prefer-to-be-undefined](docs/rules/prefer-to-be-undefined.md)               | Suggest using `toBeUndefined()`                                 | ![style][]       | ![fixable][] |
 | [prefer-to-contain](docs/rules/prefer-to-contain.md)                         | Suggest using `toContain()`                                     | ![style][]       | ![fixable][] |
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ installations requiring long-term consistency.
 | [prefer-hooks-on-top](docs/rules/prefer-hooks-on-top.md)                     | Suggest having hooks before any test cases                      |                  |              |
 | [prefer-spy-on](docs/rules/prefer-spy-on.md)                                 | Suggest using `jest.spyOn()`                                    |                  | ![fixable][] |
 | [prefer-strict-equal](docs/rules/prefer-strict-equal.md)                     | Suggest using `toStrictEqual()`                                 |                  | ![suggest][] |
+| [prefer-to-be-for-literals](docs/rules/prefer-to-be-for-literals.md)         | Suggest using `tooBe()` for primitive literals                  |                  | ![fixable][] |
 | [prefer-to-be-null](docs/rules/prefer-to-be-null.md)                         | Suggest using `toBeNull()`                                      | ![style][]       | ![fixable][] |
 | [prefer-to-be-undefined](docs/rules/prefer-to-be-undefined.md)               | Suggest using `toBeUndefined()`                                 | ![style][]       | ![fixable][] |
 | [prefer-to-contain](docs/rules/prefer-to-contain.md)                         | Suggest using `toContain()`                                     | ![style][]       | ![fixable][] |
--- a/docs/rules/prefer-to-be-for-literals.md
+++ b/docs/rules/prefer-to-be-for-literals.md
@@ -1,4 +1,4 @@
-# Suggest using `tooBe()` for primitive literals (`prefer-to-be-for-literals`)
+# Suggest using `toBe()` for primitive literals (`prefer-to-be-for-literals`)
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -161,6 +161,7 @@ installations requiring long-term consistency.
 | [prefer-hooks-on-top](docs/rules/prefer-hooks-on-top.md)                     | Suggest having hooks before any test cases                      |                  |              |
 | [prefer-spy-on](docs/rules/prefer-spy-on.md)                                 | Suggest using `jest.spyOn()`                                    |                  | ![fixable][] |
 | [prefer-strict-equal](docs/rules/prefer-strict-equal.md)                     | Suggest using `toStrictEqual()`                                 |                  | ![suggest][] |
+| [prefer-to-be-for-literals](docs/rules/prefer-to-be-for-literals.md)         | Suggest using `tooBe()` for primitive literals                  |                  | ![fixable][] |
```
SimenB, line 164:
```suggestion
| [prefer-to-be-for-literals](docs/rules/prefer-to-be-for-literals.md)         | Suggest using `toBe()` for primitive literals                   |                  | ![fixable][] |
```

docs/rules/prefer-to-be-for-literals.md
```diff
@@ -0,0 +1,30 @@
+# Suggest using `tooBe()` for primitive literals (`prefer-to-be-for-literals`)
```
SimenB, line 1:
```suggestion
# Suggest using `toBe()` for primitive literals (`prefer-to-be-for-literals`)
```

src/rules/prefer-to-be-for-literals.ts
```diff
@@ -0,0 +1,51 @@
...
+  isParsedEqualityMatcherCall(matcher) &&
+  followTypeAssertionChain(matcher.arguments[0]).type ===
+    AST_NODE_TYPES.Literal;
+
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Suggest using `tooBe()` for primitive literals',
```
SimenB, line 21:
```suggestion
      description: 'Suggest using `toBe()` for primitive literals',
```

SimenB (approved):
I like it!

LitoMore:
We should support all literals, such as `null`, `undefined`, and `NaN`.

And make its API defaults to:

```js
'jest/prefer-to-be-for-literals': ['error', {
 	literals: ['number', 'string', 'boolean']
}]
```

Its full API:

```js
'jest/prefer-to-be-for-literals': ['error', {
 	literals: ['number', 'string', 'boolean', 'undefined', 'null', 'NaN']
}]
```

What do you think?

razh:
There's already `toBeNull()` and `toBeUndefined()`, the former of which is included as a valid case:

https://github.com/jest-community/eslint-plugin-jest/blob/c5a389d5709600a536ca28180479cf32a581f9cc/src/rules/__tests__/prefer-to-be-for-literals.test.ts#L7-L9

They're handled by the [`prefer-to-be-null`](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-be-null.md) and [`prefer-to-be-undefined`](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-be-undefined.md) rules.

I assume that `NaN` already counts as a primitive literal? Please correct me if I'm wrong!

LitoMore:
Another brainstorm:

Deprecate `prefer-to-be-null` and `prefer-to-be-undefined`.

Use a new rule `prefer-to-be` instead, and make its API to:

```js
'jest/prefer-to-be': ['error', {
  types: ['number', 'string', 'boolean', 'undefined', 'null', 'NaN']
}]
```

G-Rath:
@SimenB cheers, good catch on that silly "too" 😅

G-Rath:
@LitoMore as @razh said, `prefer-to-be-undefined` & `prefer-to-be-null` already cover `undefined` & `null`, but yeah I had totally forgotten about `toBeNaN`.

I think doing it all via one `prefer-to-be` rule makes sense, though I don't think we need the options for now. That'd actually let us support `beDefined` which is something we'd been stuck on because of the name

@SimenB do you have any opinions on that? mainly around having it in our `style` config, since `prefer-to-be-undefined` & `prefer-to-be-null` are there and so would want to be replaced.

G-Rath:
also, I just realised this errors on `toBe` itself 🤦

G-Rath:
I've created `prefer-to-be` in #864 which superseeds this :)

PR #825, opened 2021-04-14
docs(expect-expect): change suggested rule config

callumgare:
Change the suggested rule config for jest/expect-expect when using supertest to use "request.**.expect" in the assertFunctionNames array rather than "request.*.expect". A supertest chained assertion may have more than one function call between `request(app)` and `.expect()`. Without this change the following would raise an error/warning:

```js
describe('Get All Items', () => {
  it('/items/all returns json list of all items', () =>
    request(app)
      .get('/items/all')
      .set('Accept', 'application/json')
      .expect('Content-Type', /json/)
      .expect(200),
  )
})
```


SimenB (approved):
thanks!

PR #831, opened 2021-04-25
Add eslint plugin istanbul

### LINKED ISSUE #222: New rule: no istanbul ignore without reason

antonk52:
This PR is a replacement for #580

Thank you for a ping @G-Rath hopefully this time it gets merged 🙂

Fixes #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,13 @@ ensure consistency and readability in jest test suites.
 
 https://github.com/dangreenisrael/eslint-plugin-jest-formatting
 
+### eslint-plugin-istanbul
+
+A set of rules to enforce good practices for Istanbul, one of the code coverage
+tools used by Jest.
+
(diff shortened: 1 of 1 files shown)


G-Rath (approved):

G-Rath:
@antonk52 you need to apply `prettier` to the README :)

antonk52:
Wow that was quick review!

I've run prettier against the README and updated the commit content.

Thank you @G-Rath

PR #844, opened 2021-05-23
chore(readme): remove `$` from install instructions

sidharthv96:
Github has introduced the Copy Markdown functionality, so users can click the copy button to get the command in their clipboard.

Having the $ in front of the copied command means they can't use this feature.


SimenB (approved):
thanks!

Issue #534, opened 2020-02-11
`expect-expect` & `no-standalone-expect` when using `expect` in afterEach


k-funk:
## Example Code

Sometimes I group my enzyme snapshot tests like this to avoid writing `expect(shallowToJson(wrapper)).toMatchSnapshot()` repeatedly:
```
import React from 'react';
import { shallow } from 'enzyme';
import shallowToJson from 'enzyme-to-json';

import Component1 from './index';


describe('outputs the expected tree when', () => {
  let wrapper;
  
  test('prop1 is foo', () => {
    wrapper = shallow((
      <Component1 prop1="foo" />
    ));
  });
  
  test('prop1 is bar', () => {
    wrapper = shallow((
      <Component1 prop1="bar" />
    ));
  });

  afterEach(() => {
    expect(shallowToJson(wrapper)).toMatchSnapshot();
  });
});
```

The tests are often times more complex than ^ with which props are being passed.

---
## Problem

There are two `eslint-plugin-jest` rules that I'd like to use, `expect-expect` and `no-standalone-expect`, but they are incompatible with the pattern above.

• `expect-expect`: my `test`s don't have an `expect` (but `afterEach` does)
• `no-standalone-expect`: my `expect` is not inside a `test` (but is in an `afterEach` of a `describe` that contains `test`s)

---

## Proposed solution

Could the two rules be modified to allow the above use case?

`no-standalone-expect` could have options for which blocks are valid. Default could remain `test` and `it`. Optional: `"jest/no-standalone-expect": ["error", allowedBlocks: ['test', 'it', 'afterEach']]`

gndelia:
I was about to open a similar ticket, but I found yours. I am having a different scenario, but similar issue.

We added this library to for our tests https://www.npmjs.com/package/jest-theories, so our tests now run inside `theoretically` functions. Though we know about using `jest.each`, this tiny adittion fit our team, given most of them have a C# background and use a very similar library

Because of this we had to disable `jest/no-standalone-expect` in the files were we started to use this library, because now `expect`s run under `theoretically` - it would be great to be able to customize the blocks, as you suggested

rdsedmundo:
+1 for the suggestion of adding the `allowedBlocks`, that would allow compatibility with `jest-in-case`.

mfogel:
Was this implemented by #585? To allow `expect` in the setup and teardown functions, I have in my eslint config:

```json
      "jest/no-standalone-expect": [
        "error",
        {
[...]

PR #850, opened 2021-05-28
feat(expect-expect): support `additionalTestBlockFunctions` option

G-Rath:
Helps with #534

Probably wants to be replaced with just `testBlockFunctions` at some point when we decide we want to support `settings`.
Could be worth doing now - haven't decided 🤔

### DIFF
--- a/docs/rules/expect-expect.md
+++ b/docs/rules/expect-expect.md
@@ -34,7 +34,8 @@ it('should work with callbacks/async', () => {
   "jest/expect-expect": [
     "error",
     {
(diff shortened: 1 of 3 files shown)


G-Rath:
(this is sort of on hold as I started looking at how we can implement this & the other general naming stuff as settings, then learnt that [ESLint handles settings weirdly](https://github.com/eslint/eslint/issues/14728), which will change [in the new config system](https://github.com/eslint/eslint/pull/14321#discussion_r627726368)).

PR #863, opened 2021-07-21
fix(valid-describe): report on concise-body arrow functions

jcfranco:
Jest 27 will throw an error if a describe block returns a value ([changelog](https://github.com/facebook/jest/blob/master/CHANGELOG.md#2700), [test-case](https://replit.com/@jcfranco/jest-playground#test.js)). This PR updates `valid-describe` to report on these instances that would fail tests. 

If deemed useful, I could look at making this fixable in a follow-up PR.

### DIFF
--- a/docs/rules/valid-describe.md
+++ b/docs/rules/valid-describe.md
@@ -43,6 +43,12 @@ describe('myFunction', () => {
     });
   });
 });
+
+// Returning a value from a describe block is not allowed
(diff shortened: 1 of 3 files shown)


G-Rath (approved):
good catch - we should have always supported this tbh

jcfranco:
Awesome, thanks for the quick review and merge!

PR #864, opened 2021-07-22
feat: create `prefer-to-be` rule

### LINKED ISSUE #801: [new-rule] recommend `toBe` over `toEqual` when expecting primitive literals

G-Rath:
An even better version of #821 (plus it doesn't flag `toBe` 🤦).

This deprecates `prefer-to-be-undefined` & `prefer-to-be-null`, as well as supporting `toBeDefined` (vs `toBeUndefined`). (@SimenB let me know if you're happy with this, or if you think they should stay as different rules)

Closes #821
Closes #801

### DIFF
--- a/docs/rules/prefer-to-be.md
+++ b/docs/rules/prefer-to-be.md
@@ -4,7 +4,9 @@ When asserting against primitive literals such as numbers and strings, the
 equality matchers all operate the same, but read slightly differently in code.
 
 This rule recommends using the `toBe` matcher in these situations, as it forms
-the most grammatically natural sentence.
+the most grammatically natural sentence. For `null`, `undefined`, and `NaN` this
+rule recommends using their specific `toBe` matchers, as they give better error
+messages as well.
 
 ## Rule details
 
@@ -27,4 +29,24 @@ expect(getMessage()).toBe('hello world');
 expect(loadMessage()).resolves.toBe('hello world');
 
 expect(catchError()).toStrictEqual({ message: 'oh noes!' });
+```
+
+For `null`, `undefined`, and `NaN`, this rule triggers a warning if `toBe` is
+used to assert against those literal values instead of their more specific
+`toBe` counterparts:
+
+```js
+expect(value).not.toBe(undefined);
+expect(getMessage()).toBe(null);
+expect(countMessages()).resolves.not.toBe(NaN);
+```
+
+The following pattern is not warning:
+
+```js
+expect(value).toBeDefined();
+expect(getMessage()).toBeNull();
+expect(countMessages()).resolves.not.toBeNaN();
+
+expect(catchError()).toStrictEqual({ message: undefined });
 ```
--- a/docs/rules/prefer-to-be.md
+++ b/docs/rules/prefer-to-be.md
@@ -11,7 +11,7 @@ messages as well.
 ## Rule details
 
 This rule triggers a warning if `toEqual()` or `toStrictEqual()` are used to
-assert a primitive literal value such as a string or a number.
+assert a primitive literal value such as numbers, strings, and booleans.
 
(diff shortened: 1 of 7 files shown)


LitoMore:
How about make this rule configurable? As mentioned in https://github.com/jest-community/eslint-plugin-jest/pull/821#issuecomment-880510570.

We could make its API defaults to:

```js
'jest/prefer-to-be': ['error', {
  types: ['number', 'string', 'boolean', 'undefined', 'null', 'NaN']
}]
```

docs/rules/prefer-to-be.md
```diff
@@ -0,0 +1,52 @@
...
+
+This rule triggers a warning if `toEqual()` or `toStrictEqual()` are used to
+assert a primitive literal value such as a string or a number.
+
+The following patterns are considered warnings:
+
+```js
+expect(value).not.toEqual(5);
+expect(getMessage()).toStrictEqual('hello world');
+expect(loadMessage()).resolves.toEqual('hello world');
```
LitoMore, line 21: What about the `toBe(false)`?

G-Rath:
@LitoMore as I said in #821, I don't see a strong need for it to be configurable - why would you want to use it for some but not others?

docs/rules/prefer-to-be.md
G-Rath: That is covered by this rule, but I'll chuck in some examples :)

LitoMore:
> as I said in #821, I don't see a strong need for it to be configurable - why would you want to use it for some but not others? (@G-Rath)

For backward compatibility, I think.

But I agree with you that make this rule is not configurable. I don't want to give any chance to users to write "bad" code.

LitoMore (approved):

PR #872, opened 2021-08-05
Fix regex in mustMatch and mustNotMatch examples

cascornelissen:
The `$` character is used to assert the end of the line and the examples show strings that start with a specific word for which the `^` character should be used.


G-Rath (approved):
Whoops yea thats my bad - I'm always getting them mixed up 😅

cascornelissen:
No problem at all, just confused for me for a little while 😅  I think I've fixed the commit message/description now ✌🏼

PR #889, opened 2021-09-12
fix(no-deprecated-functions): remove `process.cwd` from resolve paths

### LINKED ISSUE #874: [no-deprecated-functions] auto-detect error with monorepo-like project structure
Changing the auto detect from 
```ts
const jestPath = require.resolve('jest/package.json', {
      paths: [process.cwd()]
    });
```
to
```ts
const jestPath = require.resolve('jest/package.json', {
      paths: [process.cwd(), __dirname]
    });
```
fixes the bug introduced 15 months ago with non-root level projects.

### LINKED ISSUE #886: no-deprecated-functions rule probably should try to find jest from its own directory upwards, not from cwd
If we have multiple projects in a parent directory like this:

```
workspace/
├── project1/
│   └── node_modules/
│       ├── eslint/
│       ├── eslint-plugin-jest/
│       └── jest/
└── project2/
    └── node_modules/
        ├── eslint/
        ├── eslint-plugin-jest/
        └── jest/
```

Then if we run eslint in `project1` from `workspace` folder, `require.resolve` throws
because from working directory it can't resolve `jest/package.json`. Now if we remove this line it
works obviously: https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-deprecated-functions.ts#L45

Easy to reproduce with @dbaeumer's plugin for VSCode.
If we open `workspace` folder above in VSCode, no-deprecated-functions rule throws.

Not sure if there's a strong reason to resolve jest from `process.cwd()`. If not, can we remove that line?

G-Rath:
Resolves #874
Resolves #886
Resolves #888

Would be good if people could test this locally on a few projects and in editors - I think it's fine, but want to make sure we've checked that it won't make any editors explode (as thats the primary reason for why this is here).

### DIFF
--- a/src/rules/__tests__/detectJestVersion.test.ts
+++ b/src/rules/__tests__/detectJestVersion.test.ts
@@ -76,11 +76,6 @@ const setupFakeProject = (structure: ProjectStructure): string => {
   return tempDir;
 };
 
-// pin the original cwd so that we can restore it after each test
-// const projectDir = process.cwd();
-
-// afterEach(() => process.chdir(projectDir));
-
 describe('detectJestVersion', () => {
   describe('basic tests', () => {
     const packageJsonFactory = jest.fn<JSONSchemaForNPMPackageJsonFiles, []>();
--- a/src/rules/__tests__/detectJestVersion.test.ts
+++ b/src/rules/__tests__/detectJestVersion.test.ts
@@ -1,0 +1,238 @@
+import { spawnSync } from 'child_process';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { JSONSchemaForNPMPackageJsonFiles } from '@schemastore/package';
+import { create } from 'ts-node';
+import { detectJestVersion } from '../detectJestVersion';
+
+const compileFnCode = (pathToFn: string) => {
+  const fnContents = fs.readFileSync(pathToFn, 'utf-8');
+
+  return create({
+    transpileOnly: true,
+    compilerOptions: { sourceMap: false },
+  }).compile(fnContents, pathToFn);
+};
+const compiledFn = compileFnCode(require.resolve('../detectJestVersion.ts'));
+const relativePathToFn = 'eslint-plugin-jest/lib/rules/detectJestVersion.js';
+
+const runNodeScript = (cwd: string, script: string) => {
+  return spawnSync('node', ['-e', script.split('\n').join(' ')], {
+    cwd,
+    encoding: 'utf-8',
+  });
+};
+
+const runDetectJestVersion = (cwd: string) => {
+  const out = runNodeScript(
+    cwd,
+    `
+      try {
+        console.log(require('${relativePathToFn}').detectJestVersion());
+      } catch (error) {
+        console.error(error.message);
+      }
+    `,
+  );
+
+  console.log('status:', out.status);
+  console.log('stdout:', out.stdout);
+  console.log('stderr:', out.stderr);
+
+  return out;
+};
+
+/**
+ * Makes a new temp directory, prefixed with `eslint-plugin-jest-`
+ *
+ * @return {Promise<string>}
+ */
+const makeTempDir = () =>
+  fs.mkdtempSync(path.join(os.tmpdir(), 'eslint-plugin-jest-'));
+
+interface ProjectStructure {
+  [key: `${string}/package.json`]: JSONSchemaForNPMPackageJsonFiles;
+  [key: `${string}/${typeof relativePathToFn}`]: string;
+  [key: `${string}/`]: null;
+  'package.json'?: JSONSchemaForNPMPackageJsonFiles;
+}
+
+const setupFakeProject = (structure: ProjectStructure): string => {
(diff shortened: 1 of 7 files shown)


G-Rath:
hmm ok maybe that's why I added it, so we could test properly?

I think we could get almost the best of both worlds by instead doing:

```
const jestPath = require.resolve('jest/package.json', {
  paths: [
    process.cwd(),
    ...(require.resolve.paths('jest/package.json') ?? []),
  ],
});
```

But there's one test that still fails.

I'll have to do some thinking and playing around to see what we can do - I'd like to avoid adding a really big e2e-type setup where we copy around a bunch of files if possible but maybe that's what we have to go with?

SimenB:
I think we should just remove it and throw an error if people use this rule without specifying the version of Jest they use. They can easily do this in their own `.eslintrc.js` via `require('jest/package.json').version`

codejedi365:
> They can easily do this in their own `.eslintrc.js` via `require('jest/package.json').version`

Please update the documentation to reflect this concept.  It is uncommon to add a settings object to the `.eslintrc` and I had to go deep into the ESLINT configuration and guess of what to add where.

SimenB:
Our docs are lacking an example of this, yeah

G-Rath:
I've made #897 expanding our documentation to explain this :)

G-Rath:
:(

image

SimenB:
@G-Rath how's this one going? 😀

G-Rath:
@SimenB go for gold on reviewing - there's some minor cleanup I was going to look to do once I got it all green (which it is now) but  it's fine without it

src/rules/__tests__/detectJestVersion.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    fs.writeFileSync(path.join(tempDir, filePath), finalContents);
+  }
+
+  return tempDir;
+};
+
+// pin the original cwd so that we can restore it after each test
+// const projectDir = process.cwd();
+
+// afterEach(() => process.chdir(projectDir));
```
SimenB, line 88:
```suggestion
```

SimenB (changes requested):
this is missing handling of the user passing a string as the version in their settings (via a `require('jest/package.json').version`)

Beyond that, looks good 🙂

G-Rath:
> looks good

Glad you think so - I was pleasantly surprised with how nicely this turned out: while a little annoying we can't collect coverage from the process calls, the tests are fast & cleaner than before + we're not making nearly as many tmp directories as before (only ~4 vs "a whole bunch").

At some point I've been meaning to break our `utils` file up into a few files within `utils/`, which now that we have this I'll probably do soon.

btw this change should mean we can technically switch to using `export =` instead of using `babel-plugin-replace-ts-export-assignment` because that requires there to be only one export which was the case for all but this rule (as it needed to export the `_clearJestVersionCache` function)

G-Rath:
> this is missing handling of the user passing a string as the version in their settings (via a require('jest/package.json').version)

Do you mean you'd like a test that explicitly has `require('jest/package.json').version`? just otherwise I was pretty sure we already had a test in the rule for if the setting is present, but I'll double check.

src/rules/__tests__/detectJestVersion.test.ts
G-Rath: hehe yeah this was the main "minor cleanup" I was going to do - I think we should be able to use changing the directory to reduce a couple of lines from each test.

G-Rath:
ugh dang @SimenB I see what you mean now about the jest version: `require('jest/package.json').version` will return the _full_ version, which isn't what we currently expect when it's passed in explicitly. That also means #897 shouldn't be landed as-is as it has the same problem.

src/rules/no-deprecated-functions.ts
```diff
@@ -1,64 +1,26 @@
...
 
-let cachedJestVersion: JestVersion | null = null;
-
-/** @internal */
-export const _clearCachedJestVersion = () => (cachedJestVersion = null);
-
-const detectJestVersion = (): JestVersion => {
-  if (cachedJestVersion) {
-    return cachedJestVersion;
+const parseJestVersion = (rawVersion: number | string): JestVersion => {
```
G-Rath, line 16: Technically this is duplicated in `detectJestVersion`, but I'm fine with that as it's only a few lines and otherwise it calls into question a bit the `JestVersion` & underlying function name which I don't think is worth doing right now.

It'll be easy enough to refactor in the future if we find more uses for this value and/or a use for having minor or patch versions.

src/rules/__tests__/detectJestVersion.test.ts
G-Rath: So using `process.chdir` does work, but tbh I don't know which is actually nicer: using it means we don't have to pass around the project directory argument in all of our tests, but then we'll be magically changing directories and so stuff "just works" 🤔

SimenB: `process.chdir` actually changes the dir, so I'd prefer not to

SimenB (approved):

PR #891, opened 2021-09-13
fix: stop testing on Node 10 and 15

SimenB:
BREAKING CHANGE: Drop support for Node 10 and 15


package.json
```diff
@@ -131,7 +131,7 @@
     }
   },
   "engines": {
-    "node": ">=10"
+    "node": "^12.13.0 || ^14.15.0 || >=16.0.0"
```
SimenB, line 134: First LTS release of v12 and v14

PR #896, opened 2021-09-14
fix(no-deprecated-functions): do not attempt to detect version of Jest automatically

### LINKED ISSUE #874: [no-deprecated-functions] auto-detect error with monorepo-like project structure

### LINKED ISSUE #886: no-deprecated-functions rule probably should try to find jest from its own directory upwards, not from cwd

SimenB:
BREAKING CHANGE: `no-deprecated-functions` will require a Jest version specified in shared settings

Resolves #874
Resolves #886
Resolves #888
Resolves #889

### DIFF
--- a/docs/rules/no-deprecated-functions.md
+++ b/docs/rules/no-deprecated-functions.md
@@ -6,6 +6,37 @@ either been renamed for clarity, or replaced with more powerful APIs.
 While typically these deprecated functions are kept in the codebase for a number
 of majors, eventually they are removed completely.
 
+## Jest version
+
+This rule requires configuration to tell it which version of Jest is running.
+You provide this through `eslint`'s
+["shared settings"](https://eslint.org/docs/user-guide/configuring/configuration-files#adding-shared-settings).
+
+Example:
+
+```json
+{
+  "settings": {
+    "jest": {
+      "version": 27
+    }
+  }
+}
+```
+
+To avoid hard-coding a number, you can also fetch it from the installed version
+of Jest if you use a JavaScript config file such as `.eslintrc.js`:
+
+```js
+module.exports = {
+  settings: {
+    jest: {
+      version: require('jest/package.json').version,
+    },
+  },
+};
+```
+
 ## Rule details
 
 This rule warns about calls to deprecated functions, and provides details on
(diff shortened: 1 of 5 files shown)


G-Rath:
@SimenB I would _really_ like to keep this functionality as I think it's really useful - keep in mind as well that this rule is in our `recommended`, so this change would mean that users would have to have an extra bit in their configs to use the recommended.

Likewise for that same reason, I think this rule is being used without issue by the majority of our consumers; I think instead we should just amend the documentation to include how to "automatically" get the jest version if you're trying to run the linting in a monorepo-like (but not actually since shared packages are not at the toplevel) setup.

Alternatively, we should be able to abstract the `require.resolve` call into its own file which we could then mock - I'd be happy shipping that if it allowed keeping this functionality.

SimenB:
@G-Rath if you feel strongly about this I won't argue 🙂 I just wanna simplify the logic and remove any "magic" that causes confusion or errors for users. But again, your view of the situation weighs heavier in this case

G-Rath:
@SimenB yeah that's understandable, and I agree with the principle but think we can get this sorted without needing to remove the feature.

For the short-term I'll close this and make a PR updating the readme to mention that you'll need to provide the jest version in the config using `require('jest/package.json').version` in some conditions, while I have a go at seeing if we can get it working for this other case automatically.

PR #897, opened 2021-09-17
docs: expand on how jest version is determined and common workarounds

G-Rath:
This is something of a documentation stop-gap while I work on #889.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,20 +60,38 @@ This is included in all configs shared by this plugin, so can be omitted if
 extending them.
 
 The behaviour of some rules (specifically `no-deprecated-functions`) change
-depending on the version of `jest` being used.
+depending on the version of Jest being used.
 
-This setting is detected automatically based off the version of the `jest`
-package installed in `node_modules`, but it can also be provided explicitly if
-desired:
+By default, this plugin will attempt to determine the version of Jest
+automatically by looking for Jest in the `node_modules` of _the working directly
+that ESLint is being run from_.
+
+If you're working with a project structure that means Jest is not installed at
+the same directory level as where you run ESLint (such as if you're using a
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -60,22 +60,40 @@ This is included in all configs shared by this plugin, so can be omitted if
 extending them.
 
 The behaviour of some rules (specifically `no-deprecated-functions`) change
```
SimenB, line 62: we should create a section here and link to it from https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-deprecated-functions.md

SimenB (changes requested):

G-Rath:
Currently this is actually incorrect, since we assume that the version being passed in `settings` is the jest major version, which `require('jest/package.json').version` won't be - so we should either land the `parseJestVersion` function from #889 as a fix or otherwise merge this after #889 (which'd mean this should be adjusted to explain the new logic).

docs/rules/no-deprecated-functions.md
```diff
@@ -6,6 +6,10 @@ either been renamed for clarity, or replaced with more powerful APIs.
 While typically these deprecated functions are kept in the codebase for a number
 of majors, eventually they are removed completely.
 
+This rule requires knowing which version of Jest you're using - see
+[this section of the readme](README.md) for details on how that is obtained
```
SimenB, line 10: This link doesn't work, does it?

G-Rath: No it doesn't - IntelliJ seems to have weird handling of relative links in nested docs :/

G-Rath:
Have pulled these changes into #889

PR #898, opened 2021-09-17
fix: mark rules that suggest fixes with `hasSuggestion` for ESLint v8

G-Rath:
We can land this ahead of time to reduce the overhead a little.


SimenB (approved):
👍

PR #903, opened 2021-09-20
feat: support ESLint 8.x

MichaelDeBoey:
ESLint has released the first beta versions of v8 🎉
https://eslint.org/blog/2021/08/eslint-v8.0.0-beta.0-released
https://eslint.org/blog/2021/08/eslint-v8.0.0-beta.1-released
https://eslint.org/blog/2021/09/eslint-v8.0.0-beta.2-released

Dependencies should be compatible with ESLint 8 too before we can merge this one:

- [ ] [`@typescript-eslint/experimental-utils`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/experimental-utils) (https://github.com/typescript-eslint/typescript-eslint/issues/3738)
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3737
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3767
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3768
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3800
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3806
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3808
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3809
  - [ ] Release

devDependency compatibility with ESLint 8:

- [ ] [`@typescript-eslint/eslint-plugin`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin) (https://github.com/typescript-eslint/typescript-eslint/issues/3738)
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3737
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3767
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3768
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3800
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3806
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3808
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3809
  - [ ] Release
- [ ] [`@typescript-eslint/parser`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser) (https://github.com/typescript-eslint/typescript-eslint/issues/3738)
  - [x] https://github.com/typescript-eslint/typescript-eslint/pull/3737
[...]

PR #904, opened 2021-09-20
chore(ci): specify only majors for GH Actions

MichaelDeBoey:
- Set action versions to major
  This will create less PRs to update their versions
- ~Set node version to 16~

---

Taken from #903

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,6 +18,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Danger
-        uses: danger/danger-js@10
+        uses: danger/danger-js@v10
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,6 +18,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Danger
-        uses: danger/danger-js@10.6.6
+        uses: danger/danger-js@10
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/nodejs.yml
```diff
@@ -18,9 +18,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-node@v2.4.0
+      - uses: actions/setup-node@v2
         with:
-          node-version: 14.x
+          node-version: 16
```
SimenB, line 23: 16 is not LTS (yet), so I'd prefer to keep this on 14

.github/workflows/nodejs.yml
```diff
@@ -18,9 +18,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-node@v2.4.0
+      - uses: actions/setup-node@v2
```
SimenB, line 21: dependabot is what changed from just specifying major: #620. But if renovate leaves it alone, I'm happy. I was unable to find any docs specifying the behavior - do you know?

SimenB (changes requested):

.github/workflows/nodejs.yml
SimenB: I pushed this now

SimenB (approved):
thanks!

.github/workflows/lint.yml
```diff
@@ -18,6 +18,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Danger
-        uses: danger/danger-js@10.6.6
+        uses: danger/danger-js@v10
```
SimenB, line 21: @orta it doesn't work to specify a range - is this something you could fix? 😀

PR #905, opened 2021-09-20
Release v25

SimenB:
Just to keep track.

I believe for `semantic-release` to be happy we need to merge this (rather than squash or rebase), so I'll toggle the setting allowing it when we're ready to roll this.

Missing:

- [ ] #889
- [ ] #903


G-Rath:
@SimenB are you wanting to land #905 as a breaking change, or are you wanting to get it landed _before_ the new major so that folks can get that fix without needing to do the major upgrade?

I'm fine with either, but if it's the latter then I think we should apply the `parseJestVersion` code + #897 (which really should be done together) first as tbh that is a bug fix.

SimenB:
Current thinking is that #903 (which I assume you meant to link to) is a semver minor, so the plan is to merge this after #889 lands

G-Rath:
Actually I meant to link to #889 🤦

SimenB:
Ah. Plan is to land that as the last patch of 24 then move on with 25 🙂 Just missing the docs thing in #897, then I'll land all three

G-Rath:
@SimenB I just remembered about #864 😅 

That rule opens up the possibility of deprecating `prefer-to-be-undefined` & `prefer-to-be-null` (which the PR doesn't do because they're in our `style` config), so if you're happy with that we could land it now and remove those two rules in this major, replacing them with the new rule in our `style` config.

The PR is all good to go if you're happy with that :)

G-Rath:
@SimenB have opened #910 which I think we should merge before the new major too.

G-Rath:
@SimenB also random thought: should we rename `lowercase-name` to something like `prefer-lowercase-title`? I'm not too fussed, but it struck me as I was looking over some issues that it's sort of an odd duck vs our other rules (e.g. `no-identical-title`, `valid-title`).

G-Rath:
Another naming thought: might make sense to rename `valid-describe` to `valid-describe-callback`, since that is actually what the rule is checking vs e.g. `valid-title` which applies to `describe`s.

SimenB:
> @SimenB I just remembered about #864 😅
> 
> That rule opens up the possibility of deprecating `prefer-to-be-undefined` & `prefer-to-be-null` (which the PR doesn't do because they're in our `style` config), so if you're happy with that we could land it now and remove those two rules in this major, replacing them with the new rule in our `style` config.
> 
> The PR is all good to go if you're happy with that :)

Let's do it 👍 

---

> @SimenB also random thought: should we rename `lowercase-name` to something like `prefer-lowercase-title`? I'm not too fussed, but it struck me as I was looking over some issues that it's sort of an odd duck vs our other rules (e.g. `no-identical-title`, `valid-title`).

Seems reasonable to me 🙂 

---

> Another naming thought: might make sense to rename `valid-describe` to `valid-describe-callback`, since that is actually what the rule is checking vs e.g. `valid-title` which applies to `describe`s.

Yeah, agreed

PR #912, opened 2021-09-25
feat(require-top-level-describe): support enforcing max num of describes

### LINKED ISSUE #598: [new-rule] only-one-top-describe
Do we have a rule which enforces only 1 top-level `describe` in a test file?

G-Rath:
Resolves #598

The argument could be made that maybe this should be part of a more general "max-stuff" rule, but for now the logic is simple enough so happy to include it in this rule, since it's already checking for top-level describes anyway.


SimenB (approved):
I like it

PR #918, opened 2021-10-03
Deprecate rules

G-Rath:
(this should be merged via rebase, so we have separate changelog entries)

Follow up from #864


SimenB (changes requested):
I think we should just delete the old ones rather than keeping them around.

If we want to be nice to people migrating, we can add some sort of `throw 'renamed to foobar'` in the old rules - I don't think we should make them actually work

G-Rath:
Yeah I was going to delete them in `next` - this way folks could update their configs (and shared config packages) to use the new names, so that it's less breaking.

But I'm not fussed - happy to just mark them as deprecated in this PR or do nothing with them 🤷

SimenB:
Ah, this is targeting `main` - didn't notice. I think we can just skip the extra step of deprecating in the current release - hopefully people look at changelogs when bumping majors, so they'll see the rename

SimenB:
I wanna get next major out now. Can you change this PR to just remove/rename? As long as the changelog entry is good I don't think juggling a migration is worth it. It's a simple config change for people

G-Rath:
@SimenB yeah I agree - sorry for the delay, got tied up with some work stuff.

I'm keen to get the new major shipped too, so will get all the ducks in a row right now.

PR #919, opened 2021-10-03
feat: add `prefer-to-be` to style ruleset

G-Rath:
BREAKING CHANGE: recommend `prefer-to-be` rule


src/__tests__/__snapshots__/rules.test.ts.snap
```diff
@@ -91,6 +91,7 @@ Object {
     ],
     "rules": Object {
       "jest/no-alias-methods": "warn",
+      "jest/prefer-to-be": "error",
       "jest/prefer-to-be-null": "error",
       "jest/prefer-to-be-undefined": "error",
```
SimenB, line 96: these should be removed first (or at the same time)

G-Rath: Yeah I wanted to make the PR without setting it up to conflict when #918 is merged in

PR #929, opened 2021-10-10
feat: create `require-hook` rule

### LINKED ISSUE #752: [new rule] Prevent calling functions in tests outside of it & before/after blocks
The following is problematic
```
setup();

it('works', () => {
   // test
});
```

Because setup() is not correctly part of the test lifecycle, so you can get issues with state being cleared between tests etc.

It should be written like this:

```
beforeEach(() => {
    setup();
});

it('works', () => {
   // test
});
```

also could be problematic (arguably?)
```
const x = new Setup();
```

not problematic
```
const x = { test: 'data' };
```

G-Rath:
tbh this was a bit of a no-brainer to have as a rule - the more I think about it and check my existing tests, the more I think "yeah there's no reason that can't/shouldn't be wrapped in a before hook" 😅 

The docs should explain the important stuff - namely, that this rule flags pretty much all common "action" statements like re-assignments and function calls that are not within a hook.

I have excluded function calls on the `jest` global, like `jest.mock` but suspect some of those should actually also always be within a hook function?

I also think we should probably make this a recommended rule, but understand if you don't want to delay the new major any further.

Closes #752

### DIFF
--- a/src/rules/__tests__/require-hook.test.ts
+++ b/src/rules/__tests__/require-hook.test.ts
@@ -12,6 +12,14 @@ const ruleTester = new TSESLint.RuleTester({
 
 ruleTester.run('require-hook', rule, {
   valid: [
+    'describe()',
+    'describe("just a title")',
+    dedent`
+      describe('a test', () =>
+        test('something', () => {
+          expect(true).toBe(true);
+        }));
+    `,
     dedent`
       test('it', () => {
         //


G-Rath:
I'm wondering if this rule should have a `strict` or `loose` option (depending on what we want as the default) to adjust how it handles cases like:

```
const myValue = getValue();

  ...
});
```

For the above you could argue that `myValue` should be a `let` that gets assigned in a `beforeEach`, based on it being assigned the return from a function call (vs if it was a primitive literal, in which case it is fine).

But I wouldn't be surprised if some folks found it too noisy, due to having builder functions that are returning static values? 🤔

PR #940, opened 2021-10-11
feat: support `eslint@8`

SimenB:
@G-Rath this is not meant to be merged, just to see where we're at.

- [ ] I have to use resolutions for other usage of `@typescript-eslint/experimental-utils` as some use eslint internals
- [x] typecheck fails 
image
- [ ] `src/rules/__tests__/unbound-method.test.ts` fails, so that rule needs to be tweaked to work with ~v5 of `@typescript-eslint/eslint-plugin`~ v8 of `eslint`
- [x] need to patch `eslint-plugin-import` until https://github.com/import-js/eslint-plugin-import/pull/2194 is released

Closes #903

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
       {
         "displayName": "test",
         "moduleNameMapper": {
-          "@eslint/eslintrc/universal": "@eslint/eslintrc/dist/eslintrc-universal.cjs",
           "eslint/use-at-your-own-risk": "eslint/lib/unsupported-api.js"
         },
         "testPathIgnorePatterns": [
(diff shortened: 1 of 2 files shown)


SimenB:
Seems v4 of the plugin doesn't support eslint v8, so we need to exclude that combination from CI. I still think it's worthwhile for _us_ to attempt to support both versions, tho

SimenB:
Ok, the issue (beyond missing update to `eslint-plugin-import`) just seems to be that the unbound rule is broken on ESLint v8. Beyond that we're good to go

PR #970, opened 2021-10-30
fix(no-deprecated-functions): mark jest as an optional peer dependency

jdanil:
`jest/no-deprecated-functions` requires `jest` if a version isn't explicitly set in the config to automatically detect it from jest's `package.json` manifest file.

As `jest` isn't declared as a dependency of `eslint-plugin-jest`, stricter node linkers like pnp and pnpm throw errors when trying to run the rule.

I opted to add it as an optional peer dependency as some consumers may explicitly declare their jest version in their eslint config, may not be using this specific rule, or `eslint-plugin-jest` may be included in a shareable eslint config that runs on some packages that may not include jest tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,9 @@
   "peerDependenciesMeta": {
     "@typescript-eslint/eslint-plugin": {
       "optional": true
+    },
+    "jest": {
+      "optional": true
     }
   },
   "engines": {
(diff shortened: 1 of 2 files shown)


erwindetesters:
Great to see a fix for this @jdanil as we are running into this with our shareable setup 👍

G-Rath (approved):
Thanks!

jdanil:
Thanks for updating the lockfile! Forgot the `yarn install` 😅

Issue #961, opened 2021-10-16
[require-hook] what should we be doing for `const` variables?


G-Rath:
Currently `require-hook` considers `const` valid to have outside of hooks and test methods; this is to allow this sort of code:

```
const { S3 } = require('@aws-sdk/client-s3');
const { mocked } = require('ts-jest/utils');

jest.mock('@aws-sdk/client-s3');

const mockS3 = mocked(S3, true);

const buildApiGatewayEvent = (message: string): APIGatewayProxyEvent => {
  return {
    body: JSON.stringify({ message }),
    headers: {}
  } as APIGatewayProxyEvent;
};
```

However, this also allows doing things like this:

```
describe('when the body is not valid json', () => {
  const event = buildApiGatewayEvent('');

  event.body = ''; // mutation!

  it('returns Bad Request', async () => {
    await expect(
      handler(event, fakeContext, console.log)
    ).resolves.toHaveProperty('statusCode', 400);
  });
});
```

Overall this isn't the worst thing in the world but it can be a nasty bug and part of what `require-hook` is meant to be helping you avoid; it is made worse if you consider that something in a _test_ might be mutating the value:

```
const event = buildApiGatewayEvent('');

describe('when the body is not valid json', () => {
  it('returns Bad Request', async () => {
    event.body = ''; // mutation!

    await expect(
      handler(event, fakeContext, console.log)
    ).resolves.toHaveProperty('statusCode', 400);
  });
});

describe('when the body is valid json', () => {
  it('returns Ok', async () => {
    await expect(
      handler(event, fakeContext, console.log)
    ).resolves.toHaveProperty('statusCode', 200);
  });
});
```

In the above, the second test will fail due to `event.body` having been made blank in the previous test (_assuming that test runs_, which makes it even more confusing since if you stick a `.only` it'll suddenly start passing); ideally instead the `event` variable should be replaced with this:

```
let event;

beforeEach(() => {
  event = buildApiGatewayEvent('');
});
```

That way there's a new `event` every run.

-----

I'm reasonably confident in our ability to detect if a `const` assignment should be in a hook (or specifically, that it should not be)  due to the reasonably low number of patterns that actually should be allowed, so currently am proposing we do one of the following:

1. Continue to always ignore `const` variables, and assume the developer understands the implications (or lack of) for objects & arrays
2. Continue to always ignore `const` variables, with an option to consider `const` to be an error (excluding "special cases")
[...]

PR #983, opened 2021-11-13
Adding optional settings to require-hook rule

Kreeg:
### The problem
Some testing libs provide methods which takes hook as an argument and should be executed outside of a hook. The greatest example is using [`enableAutoDestroyHook` from `@vue/test-utils` lib ](https://vue-test-utils.vuejs.org/api/#enableautodestroy-hook). 
By default `jest/require-hook` rule throws an error in such correct code
```js
import {
  enableAutoDestroy,
  resetAutoDestroyState,
  mount
} from '@vue/test-utils'
import Foo from './Foo.vue'

// calls wrapper.destroy() after each test
enableAutoDestroy(afterEach)
// resets auto-destroy after suite completes
afterAll(resetAutoDestroyState)

describe('Foo', () => {
  it('renders a div', () => {
    const wrapper = mount(Foo)
    expect(wrapper.contains('div')).toBe(true)
    // no need to call wrapper.destroy() here
  })
})
```

To deal with it you must ignore `jest/require-hook` rule (or mark it as warn). Which is absolutely wrong and causing all the problems described in original rule's docs.

### The solution
In this PR I added addtional settings for the rule to exclude such tools as `enableAutoDestroy` with additional settings to reduce or maybe extend allowed hooks.

### DIFF
--- a/docs/rules/require-hook.md
+++ b/docs/rules/require-hook.md
@@ -148,3 +148,40 @@ afterEach(() => {
   clearCityDatabase();
 });
 ```
+
+## Options
+
+If there are methods that you want to call outside of hooks and tests, you can
+mark them as allowed using the `allowedFunctionCalls` option.
+
+```json
+{
+  "jest/require-hook": [
+    "error",
+    {
+      "allowedFunctionCalls": ["enableAutoDestroy"]
+    }
+  ]
+}
+```
+
+Examples of **correct** code when using
+`{ "allowedFunctionCalls": ["enableAutoDestroy"] }` option:
+
+```js
+/* eslint jest/require-hook: ["error", { "allowedFunctionCalls": ["enableAutoDestroy"] }] */
+
+import { enableAutoDestroy, mount } from '@vue/test-utils';
+import { initDatabase, tearDownDatabase } from './databaseUtils';
+
+enableAutoDestroy(afterEach);
+
+beforeEach(initDatabase);
+afterEach(tearDownDatabase);
+
+describe('Foo', () => {
+  test('always returns 42', () => {
+    expect(global.getAnswer()).toBe(42);
+  });
+});
+```
--- a/src/rules/__tests__/require-hook.test.ts
+++ b/src/rules/__tests__/require-hook.test.ts
@@ -152,6 +152,18 @@ ruleTester.run('require-hook', rule, {
         });
       });
     `,
+    {
+      code: dedent`
+        enableAutoDestroy(afterEach);
+        
+        describe('some tests', () => {
+          it('is false', () => {
+            expect(true).toBe(true);
+          });
+        });
+      `,
+      options: [{ allowedFunctionCalls: ['enableAutoDestroy'] }],
+    },
   ],
   invalid: [
     {
@@ -370,6 +382,25 @@ ruleTester.run('require-hook', rule, {
         {
           messageId: 'useHook',
           line: 50,
+          column: 1,
+        },
+      ],
+    },
+    {
+      code: dedent`
+        enableAutoDestroy(afterEach);
+        
+        describe('some tests', () => {
+          it('is false', () => {
+            expect(true).toBe(true);
+          });
+        });
+      `,
+      options: [{ allowedFunctionCalls: ['someOtherName'] }],
+      errors: [
+        {
+          messageId: 'useHook',
+          line: 1,
           column: 1,
         },
       ],
--- a/src/rules/require-hook.ts
+++ b/src/rules/require-hook.ts
@@ -27,12 +27,18 @@ const isNullOrUndefined = (node: TSESTree.Expression): boolean => {
   );
 };
 
-const shouldBeInHook = (node: TSESTree.Node): boolean => {
+const shouldBeInHook = (
+  node: TSESTree.Node,
+  allowedFunctionCalls: readonly string[] = [],
+): boolean => {
   switch (node.type) {
     case AST_NODE_TYPES.ExpressionStatement:
-      return shouldBeInHook(node.expression);
+      return shouldBeInHook(node.expression, allowedFunctionCalls);
     case AST_NODE_TYPES.CallExpression:
-      return !isJestFnCall(node);
+      return !(
+        isJestFnCall(node) ||
+        allowedFunctionCalls.includes(getNodeName(node) as string)
+      );
     case AST_NODE_TYPES.VariableDeclaration: {
       if (node.kind === 'const') {
         return false;
@@ -48,7 +54,10 @@ const shouldBeInHook = (node: TSESTree.Node): boolean => {
   }
 };
 
-export default createRule({
+export default createRule<
+  [{ allowedFunctionCalls?: readonly string[] }],
+  'useHook'
+>({
   name: __filename,
   meta: {
     docs: {
@@ -60,13 +69,30 @@ export default createRule({
       useHook: 'This should be done within a hook',
     },
     type: 'suggestion',
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allowedFunctionCalls: {
+            type: 'array',
+            items: { type: 'string' },
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
   },
-  defaultOptions: [],
+  defaultOptions: [
+    {
+      allowedFunctionCalls: [],
+    },
+  ],
   create(context) {
+    const { allowedFunctionCalls } = context.options[0] ?? {};
+
     const checkBlockBody = (body: TSESTree.BlockStatement['body']) => {
       for (const statement of body) {
-        if (shouldBeInHook(statement)) {
+        if (shouldBeInHook(statement, allowedFunctionCalls)) {
           context.report({
             node: statement,
             messageId: 'useHook',


G-Rath:
I'm heading to bed so will give it a proper review tomorrow, but I'm thinking we should just have an `allowedFunctions` option, as I think functions taking hooks as their first argument is reasonably specific (I've never come across it myself until now), and if we check for this arguably there could be other patterns people could like us to support (like functions that take the hook as the last argument, if they exist)

G-Rath (changes requested):
I'm heading to bed so will give it a proper review tomorrow, but I'm thinking we should just have an `allowedFunctions` option, as I think functions taking hooks as their first argument is reasonably specific (I've never come across it myself until now), and if we check for this arguably there could be other patterns people could like us to support (like functions that take the hook as the last argument, if they exist)

G-Rath:
Btw I'm pretty sure you can call the method in a `beforeAll` and that that would technically be better as Jest would ensure its actually called in the right order.

But I think it's worth us supporting `allowedFunctions` anyway :)

Kreeg:
Hey @G-Rath! Thanks for commenting.
Yes, I think that's reasonable to allow hooks to be not in specific order. But I'm sure it's good to validate that hooks are passed in arguments list, because I have not yet encountered with any valid function called outside a hook

The first idea came to me is
```js
// helpers.js
export default doAllMocks = () => {
  jest.mock('oneModule');
  jest.mock('anotherModule');
  // list of 20 mocks
}

// test.js
import doAllMocks from './helpers';

doAllMocks();
```

But I'm not sure it is a good practice.

> Btw I'm pretty sure you can call the method in a `beforeAll` and that that would technically be better as Jest would ensure its actually called in the right order.

Yes, technically it is good but for me it looks like a little overengineered, like calling `afterAll` hook inside `beforeAll` (for example using resetAutoDestroyState in `@vue/test-utils`) just for the linting to be passed. 

Goodnight! I'll be waiting for any decisions 👍

G-Rath:
@Kreeg I still think we should go with a more generic option as it means less complexity on our side and if we wanted to make it more generic later that'd be a breaking change since we'd be removing an option.

e.g. it's perfectly valid (and easy) to do something like this:

```
export const setupAutoDestroy = () => {
  beforeAll(() => {
    enableAutoDestroy(afterEach);
  });
};
```

This would also make it easier to reuse e.g. for #961, when checking `const` assignments.

Kreeg:
@G-Rath it seems reasonable for me. 
I'll update PR tomorrow, now it's my time to go to bed :D

Kreeg:
@G-Rath hey! I updated the branch. I personally don't like meaningless commits so I force pushed it.

docs/rules/require-hook.md
```diff
@@ -148,3 +148,74 @@ afterEach(() => {
...
 });
 ```
+
+## Options
+
+Some test utils provides methods which takes hook as an argument 
+and should be executed outside a hook.
+
+For example https://vue-test-utils.vuejs.org/api/#enableautodestroy-hook 
+which takes the hook as an argument. To exclude them you can update settings
```
G-Rath, line 158: The fact that the methods take hooks as an argument is irrelevant, so this should be say something along the lines of "if there are methods that you want to call outside of hooks and tests, you can mark them as allowed using the `allowedFunctionCalls` option" :)

docs/rules/require-hook.md
```diff
@@ -148,3 +148,74 @@ afterEach(() => {
...
+
+```js
+/* eslint jest/require-hook: ["error", { "excludedFunctions": ["enableAutoDestroy"] }] */
+
+import {
+  enableAutoDestroy,
+  resetAutoDestroyState,
+  mount
+} from '@vue/test-utils';
+import {initDatabase, tearDownDatabase} from './databaseUtils';
```
G-Rath, line 208: Personally I think it's good to try and keep the "correct" and "incorrect" examples similar, so think we should either stick with just having `initDatabase` or have `tearDownDatabase` in both examples.

docs/rules/require-hook.md
```diff
@@ -148,3 +148,74 @@ afterEach(() => {
...
+
+Examples of **incorrect** code for the `{ "excludedFunctions": ["enableAutoDestroy"] }`
+option:
+
+```js
+/* eslint jest/require-hook: ["error", { "excludedFunctions": ["enableAutoDestroy"] }] */
+
+import {
+  enableAutoDestroy,
+  resetAutoDestroyState,
```
G-Rath, line 181: nit: this isn't actually used in this code 😅

src/rules/__tests__/require-hook.test.ts
```diff
@@ -152,6 +152,18 @@ ruleTester.run('require-hook', rule, {
         });
       });
     `,
+    {
+      code: dedent`
```
G-Rath, line 156: Could you fix up the indenting on these please? the code itself should be indented one level, and the closing backtick should be at the same level as the properties, e.g.

```
{
  code: dedent`
    enableAutoDestroy(forEach);

      it('is false', () => {
        expect(true).toBe(true);
      });
    });
  `,
  options: [{ excludedFunctions: ['enableAutoDestroy'] }],
}
```

src/rules/require-hook.ts
```diff
@@ -27,12 +31,28 @@ const isNullOrUndefined = (node: TSESTree.Expression): boolean => {
...
+  node: TSESTree.CallExpression,
+  options: RequireHooksOptions,
+): boolean => {
+  const nodeName = getNodeName(node);
+
+  if (nodeName === null) {
+    return false;
+  }
+
+  return !!options.excludedFunctions?.includes(nodeName);
```
G-Rath, line 44: You should be able to just cast this to a `string`, which makes the code a little smaller without any real loss (since this is a known TS limitation that we've got already in a few other places)

```suggestion
  return !!options.excludedFunctions?.includes(getNodeName(node) as string));
```

src/rules/require-hook.ts
```diff
@@ -27,12 +31,28 @@ const isNullOrUndefined = (node: TSESTree.Expression): boolean => {
...
+};
+
+const shouldBeInHook = (
+  node: TSESTree.Node,
+  options: RequireHooksOptions,
```
G-Rath, line 49: Since we currently don't have any other options, I think we might as well just pass in an string array as that's more expressive and it's not a lot of work to refactor if we add more options in the future

(it also means you don't have to build an object when calling this method)

G-Rath (changes requested):
Looks good - got a couple of minor changes I'd like to see; also, what do you think about using "allow" instead of exclude? e.g. `allowedFunctionCalls`.

I feel like that might be a better name since we're _allowing_ functions to be _called_ - exclude to me feels like we're bailing out of checking completely, which isn't the case (we're just not reporting them).

Kreeg:
@G-Rath Saw your review, I will answer and update later tonight, thank you!

src/rules/require-hook.ts
Kreeg: Thanks! This is my first encounter with TS.

src/rules/require-hook.ts
Kreeg: @G-Rath thanks. Also I got rid of `RequireHooksOptions` interface.

docs/rules/require-hook.md
Kreeg: I updated the documentation but I still don't like it. Because the misleading option `allowedFunctionCalls`, which contain correct function, but the code example is still incorrect because of another function calls. Is it good?

src/rules/__tests__/require-hook.test.ts
```diff
@@ -374,6 +386,25 @@ ruleTester.run('require-hook', rule, {
...
+        enableAutoDestroy(afterEach);
+        
+        describe('some tests', () => {
+          it('is false', () => {
+            expect(true).toBe(true);
+          });
+        });
+    `,
+      options: [{ allowedFunctionCalls: ['someOtherName'] }],
```
G-Rath, line 399:
```suggestion
        });
      `,
      options: [{ allowedFunctionCalls: ['someOtherName'] }],
```

src/rules/require-hook.ts
```diff
@@ -60,13 +75,35 @@ export default createRule({
...
   create(context) {
+    const { allowedFunctionCalls } = context.options[0] ?? {};
+
     const checkBlockBody = (body: TSESTree.BlockStatement['body']) => {
       for (const statement of body) {
-        if (shouldBeInHook(statement)) {
+        if (
+          shouldBeInHook(
+            statement,
+            Array.isArray(allowedFunctionCalls) ? allowedFunctionCalls : [],
```
G-Rath, line 104: This should always be an array due to the schema - if you're doing this because `allowedFunctionCalls` is being marked as possibly undefined by TS, you can provide a default value in the assignment

src/rules/require-hook.ts
```diff
@@ -48,7 +60,10 @@ const shouldBeInHook = (node: TSESTree.Node): boolean => {
   }
 };
 
-export default createRule({
+export default createRule<
+  [Partial<{ allowedFunctionCalls?: readonly string[] }>],
```
G-Rath, line 64:
```suggestion
  [{ allowedFunctionCalls?: readonly string[] }],
```

nit: `Partial` marks all properties as optional, so we don't need it if the properties already are optional and likewise we don't need to mark the properties as optional if we're using `Partial` 

If we had more properties, using `Partial` could be nicer but for now since we've just got the one we can make the code a little smaller by marking the sole property as optional.

docs/rules/require-hook.md
```diff
@@ -148,3 +148,61 @@ afterEach(() => {
...
+    "error",
+    {
+      "allowedFunctionCalls": ["enableAutoDestroy"]
+    }
+  ]
+}
+```
+
+Examples of **incorrect** code for the
+`{ "allowedFunctionCalls": ["enableAutoDestroy"] }` option:
```
G-Rath, line 169: nit: I think it might make more sense to say:

```suggestion
Examples of **incorrect** code when using the `allowedFunctionCalls` option:
```

since we've got the config straight after in a comment

docs/rules/require-hook.md
```diff
@@ -148,3 +148,61 @@ afterEach(() => {
...
+
+describe('Foo', () => {
+  test('always returns 42', () => {
+    expect(global.getAnswer()).toBe(42);
+  });
+});
+```
+
+Examples of **correct** code for the
+`{ "allowedFunctionCalls": ["enableAutoDestroy"] }` option:
```
G-Rath, line 190: nit: I think it might make more sense to say:

```suggestion
Examples of **correct** code when using the `allowedFunctionCalls` option:
```

since we've got the config straight after in a comment

G-Rath (approved):
Looking good! Got a few minor nits/recommendations, but nothing blocking - will merge sometime this week :)

src/rules/require-hook.ts
G-Rath: No problem, you're doing good!

src/rules/__tests__/require-hook.test.ts
Kreeg: Got it. Is there any automatic linting for that? Neither prettier nor eslint fixed this.

G-Rath: Sadly not because technically it affects the content (since you're adding more spaces inside the string) but we strip it away using `dedent`, which eslint & prettier don't understand.

Kreeg:
@G-Rath I will fix the code due to your suggestions tomorrow, thank you :)

docs/rules/require-hook.md
G-Rath: > Because the misleading option allowedFunctionCalls,

I'm not sure I follow you on what you're finding misleading?

I'm happy if you want to play with the documentation some more - do you think it would be better if we didn't have _any_ config in our "incorrect" example (e.g. "Example of incorrect code without using `allowedFunctionCalls`"), and left the "correct" example as-is?

Kreeg: @G-Rath yes, the problem is just with "incorrect" example. So in docs we are saying "incorrect usage of allowed functions" but there is no incorrect use of these functions in the example. This makes some confusion so I'll remove it

G-Rath (approved):
Thanks for this!

G-Rath:
(I used `fix` instead of `feat`, so manually re-worded the commit: https://github.com/jest-community/eslint-plugin-jest/commit/9d9336a7624c53c0bb3ee899b8cc336a0b3349cb 😅 )

PR #994, opened 2021-12-02
feat(expect-expect): support using wildcards in `additionalTestBlockFunctions`

### LINKED ISSUE #993: Allow additionalTestBlockFunctions to take a regex
I would like to do something like this

```
  "jest/expect-expect": [
    "error",
    {
        assertFunctionNames: ['expect', 'expect*', 'assert*'],
        additionalTestBlockFunctions: ['expect*', 'assert*']
    }
  ]
```

So that I can modularize assertion functions but still get the `expect-expect` checking on it. Seems like this would be easy to do by duplicating what assertionFunctionNames is doing

matchesAssertFunctionName 

https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/expect-expect.ts#L25

and using it here: 

https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/expect-expect.ts#L118

NickBolles:
TODO
- [x] - update `no-standalone-expect`
- [x] - update tests
- [x] - update docs
- [ ] - fix tests 😦 

Fixes #993

### DIFF
--- a/src/rules/__tests__/expect-expect.test.ts
+++ b/src/rules/__tests__/expect-expect.test.ts
@@ -88,6 +88,20 @@ ruleTester.run('expect-expect', rule, {
       `,
       options: [{ additionalTestBlockFunctions: ['theoretically'] }],
     },
+    {
+      code: dedent`
+        const assertFoo = (input) => {
+          expect(input).toBe("foo");
+        }
+        it("should pass", () => assertFoo("foo"))
+      `,
+      options: [
+        {
+          assertFunctionNames: ['assert*'],
+          additionalTestBlockFunctions: ['assert*'],
+        },
+      ],
+    },
   ],
 
   invalid: [
@@ -144,6 +158,26 @@ ruleTester.run('expect-expect', rule, {
         })
       `,
       options: [{ additionalTestBlockFunctions: ['theoretically'] }],
+      errors: [
+        {
+          messageId: 'noAssertions',
+          type: AST_NODE_TYPES.CallExpression,
+        },
+      ],
+    },
+    {
+      code: dedent`
+        const assertFoo = (input) => {
+          var valid = input === "foo"
+        }
+        it("should pass", () => assertFoo("foo"))
+      `,
+      options: [
+        {
+          assertFunctionNames: ['assert*'],
+          additionalTestBlockFunctions: ['assert*'],
+        },
+      ],
       errors: [
         {
           messageId: 'noAssertions',
--- a/docs/rules/expect-expect.md
+++ b/docs/rules/expect-expect.md
@@ -107,14 +107,15 @@ describe('GET /user', function () {
 ### `additionalTestBlockFunctions`
 
 This array can be used to specify the names of functions that should also be
-treated as test blocks:
+treated as test blocks. Function names can use wildcards i.e `expect*`,
+`expectUtils.**`, `*.expect*`
 
 ```json
 {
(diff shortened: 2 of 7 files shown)


src/rules/__tests__/expect-expect.test.ts
```diff
@@ -151,6 +165,26 @@ ruleTester.run('expect-expect', rule, {
...
+    {
+      code: dedent`
+        const assertFoo = (input) => {
+          var valid = input === "foo"
+        }
+        it("should pass", () => assertFoo("foo"))
+      `,
+      options: [
+        {
+          assertFunctionNames: ['assert*'],
```
G-Rath, line 177: We should be able to have a test case without this option being provided, to keep things simple (having _both_ test cases is good too).

e.g. like this https://github.com/jest-community/eslint-plugin-jest/pull/994/files#diff-be15fe30b6db731ff0d3ddcc2746228fd05cad63180b95c497b00869c2e7e601R144-R153

src/rules/__tests__/expect-expect.test.ts
```diff
@@ -88,6 +88,20 @@ ruleTester.run('expect-expect', rule, {
...
+    {
+      code: dedent`
+        const assertFoo = (input) => {
+          expect(input).toBe("foo");
+        }
+        it("should pass", () => assertFoo("foo"))
+      `,
+      options: [
+        {
+          assertFunctionNames: ['assert*'],
```
G-Rath, line 100: We should be able to have a test case without this option being provided, to keep things simple (having _both_ test cases is good too).

e.g. like this https://github.com/jest-community/eslint-plugin-jest/pull/994/files#diff-be15fe30b6db731ff0d3ddcc2746228fd05cad63180b95c497b00869c2e7e601R144-R153

Additionally, this test is currently failing because `assertFunctionNames` is meant to be *all* of the assertion functions to allow - i.e. it's default is `['expect']`, meaning if you pass a custom value you need to explicitly include `'expect'`.

G-Rath (changes requested):
Overall looks good (minus the failing tests), but I think it would be good to have a changelog entry for each rule which means either doing a PR for each rule or having one PR with only two commits (that follow the conventional commit format), one for `expect-expect` & another for `no-standalone-expect`.

Let me know if you'd like me to help, or if you'd prefer I take care of that side of things. 

(I'd say it's probably best to do two different PRs, as it looks like the change for `no-standalone-expect` has broken other tests, which could be because the regex is not suitable).

G-Rath:
@NickBolles are you still working on this?

NickBolles:
@G-Rath totally forgot about it. I'll come back to it sometime in the next week

G-Rath:
@NickBolles are you willing to work on this, or are you happy if I pick it up?

NickBolles:
> @NickBolles are you willing to work on this, or are you happy if I pick it up?

You're welcome to pick it up, otherwise I'd still like to get to it.sorry for the delay.

SimenB:
is this one still alive? 🙂

G-Rath:
I don't think this PR is no, but I started to implement myself and after looking at the rules source code for like half an hour I'm pretty sure its super broken and needs re-implementing (i.e. its not checking for expects using `parseJestFnCall`).

Based on the tests I also think that this change is actually not quite what the OP wanted, but its hard to know for sure because they didn't give a code example in the issue 🤔

PR #1003, opened 2021-12-26
ci: properly configure smoke test

G-Rath:
whoops

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -4,9 +4,6 @@ on:
   schedule:
     - cron: '0 0 * * SUN'
   workflow_dispatch:
-  pull_request:
-    branches:
-      - main
 
 jobs:
   test:
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -4,6 +4,9 @@ on:
   schedule:
     - cron: '0 0 * * SUN'
   workflow_dispatch:
+  pull_request:
+    branches:
+      - main
 
 jobs:
   test:
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -24,4 +24,4 @@ jobs:
       - uses: AriPerkkio/eslint-remote-tester-run-action@v3
         with:
           issue-title: 'Results of weekly scheduled smoke test'
-          eslint-remote-tester-config: smoke-test/eslint-remote-tester.config.js
+          eslint-remote-tester-config: eslint-remote-tester.config.js
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -24,4 +24,4 @@ jobs:
       - uses: AriPerkkio/eslint-remote-tester-run-action@v3
         with:
           issue-title: 'Results of weekly scheduled smoke test'
-          eslint-remote-tester-config: eslint-remote-tester.config.js
+          eslint-remote-tester-config: smoke-test/eslint-remote-tester.config.ts
--- a/package.json
+++ b/package.json
@@ -111,6 +111,8 @@
     "eslint-plugin-import": "^2.25.1",
     "eslint-plugin-node": "^11.0.0",
     "eslint-plugin-prettier": "^3.4.1",
+    "eslint-remote-tester": "^2.0.1",
+    "eslint-remote-tester-repositories": "^0.0.3",
     "husky": "^7.0.2",
     "is-ci": "^3.0.0",
     "jest": "^27.0.0",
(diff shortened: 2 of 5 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -21,4 +21,3 @@ jobs:
       - uses: AriPerkkio/eslint-remote-tester-run-action@v3
         with:
           issue-title: 'Results of weekly scheduled smoke test'
-          eslint-remote-tester-config: smoke-test/eslint-remote-tester.config.js
```
AriPerkkio, line 24: The action doesn't actually pick these `.ts` configurations by default since action parameter has default value for `.js` config: https://github.com/AriPerkkio/eslint-remote-tester-run-action/blob/e72eaaa36788a76ef9044271bebbbd83377e7d6d/action.yml#L14-L17

So this needs to be specified here: 
```yml
    eslint-remote-tester-config: smoke-test/eslint-remote-tester.config.ts
```

package.json
```diff
@@ -111,6 +111,8 @@
     "eslint-plugin-import": "^2.25.1",
     "eslint-plugin-node": "^11.0.0",
     "eslint-plugin-prettier": "^3.4.1",
+    "eslint-remote-tester": "^2.1.0",
```
AriPerkkio, line 114:
```suggestion
    "eslint-remote-tester": "^2.1.1",
```

.github/workflows/smoke-test.yml
G-Rath: Ah yeah I wondered if that might be the case 👍

G-Rath:
@AriPerkkio thanks for the review - tbh I'm sort of on the fence about sticking with this now that I know you've actually been running the same workflow for our plugin and from what I can tell we've never actually had an error; it seems like a bit of a waste of power to be running the same ~6 hour workflow twice...

G-Rath:
(I'm going to merge now just so it's fixed, but still thinking about maybe reverting)

G-Rath:
@AriPerkkio random other question: does the remote tester support anyway to handle linting repositories multiple times with different rule configurations? My thinking is about how rule logic can be changed with options, so running against e.g. `jest/all` wouldn't cover _all_ of the possible code the rules could run.

I can't find anything in the docs that look like a way to do this - if I've not missed anything, let me know if you think this could be useful and I can open a new issue to continue discussing.

.github/workflows/smoke-test.yml
```diff
@@ -21,4 +21,4 @@ jobs:
...
+          eslint-remote-tester-config: smoke-test/eslint-remote-tester.config.ts
```
AriPerkkio, line 24: @G-Rath this path is still incorrect 😄 

```diff
-          eslint-remote-tester-config: smoke-test/eslint-remote-tester.config.ts
+          eslint-remote-tester-config: eslint-remote-tester.config.ts
```

https://github.com/jest-community/eslint-plugin-jest/issues/1002#issuecomment-1003642151

G-Rath: Yup

AriPerkkio:
> does the remote tester support anyway to handle linting repositories multiple times with different rule configurations?

At the moment, no. I've also had this idea before but didn't proceed with implementing it. I think the implementation wouldn't be that difficult though.
This is the spot where we are reading the `config.eslintrc` which can be either a single configuration, or function returning one. https://github.com/AriPerkkio/eslint-remote-tester/blob/30803c6b628da08515c6fbbe49065153dd727982/lib/engine/worker-task.ts#L212-L218
It could simply return a list of configuration which would then be looped in the rest of `workerTask` below.

Currently it is only possible to initialize `config.eslintrc` based on repository, e.g. checking whether project has `tsconfig.json` and enabling certain rules: https://github.com/AriPerkkio/eslint-remote-tester/blob/30803c6b628da08515c6fbbe49065153dd727982/ci/plugin-configs/typescript-eslint-eslint-plugin.config.js#L10.

AriPerkkio:
> tbh I'm sort of on the fence about sticking with this now that I know you've actually been running the same workflow for our plugin and from what I can tell we've never actually had an error

I've been running the `jest/all` for long time and this has been one of the most stable plugin I've been testing. Though there have been some other plugins as well which have been running successfully for months and then suddenly caused random crashes. [`eslint-plugin-regexp` is a good example here](https://github.com/ota-meshi/eslint-plugin-regexp/issues?q=author%3AAriPerkkio).

Running these tests in repository where main branch contains unreleased changes makes a lot sense, e.g. [`eslint-plugin-react`](https://github.com/yannickcr/eslint-plugin-react/issues?q=author%3Aapp%2Fgithub-actions). However this project uses automatic releasing so that's not perfect reasoning here.

I think you'll get most of these tests if `eslintrc.rules` are configured extensively to cover as much code blocks as possible. This is something I've imagined to happen in projects themselves.
Also bringing the tests closer to the actual plugin projects is always better since I don't have to be building minimal repros and reporting the bugs anymore. :smile:
You'll get issues reported by Github bot automatically in the project.

PR #1025, opened 2022-01-10
perf(utils): reduce complexity of getTestCallExpressionsFromDeclaredV…

Connormiha:
I refactored `getTestCallExpressionsFromDeclaredVariables` to avoid performance impact.

- `concat` creates new array on each iteration(It means squad complexity). `acc.push()` more preferable, because it doesn't create new collection.
- Avoided creating an intermediate collection with chain `map`+`filter`


G-Rath:
I'm not really convinced that this is a problem - do you have any benchmarks that show the performance and memory wins from this change?

G-Rath:
Closing per my previous comment

PR #1027, opened 2022-01-14
feat: create `no-conditional-in-test` rule

### LINKED ISSUE #1026: no-if rule doesn't work in many cases
The `jest/no-if` rule has a bug which results in it failing to find conditions in many cases.

[This pops from the stack whenever a call expression ends](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-if.ts#L102-L104) yet [this only adds to the stack when specific call expression start](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-if.ts#L77-L85).

This results in more items being popped off than pushed.

Example:

```
test("shows error", () => {
    if (1 === 2) {
        expect(true).toBe(false);
    }
});

test("does not show error", () => {
    setTimeout(() => console.log("noop"));
    if (1 === 2) {
        expect(true).toBe(false);
    }
});
```

The inner function in the second example causes the stack to be popped off one too many times and so by the time it arrives at the condition, the stack is empty when it shouldn't be.

I believe there are also other cases where this can occur as when debugging I often see it attempting to pop off the stack several times despite the stack being empty.

G-Rath:
This replace is meant to replace `no-if`: in addition to making the name and documentation consistent (since the rule supports more than just `if` statements), it greatly simplifies it by reporting _any_ condition in a test state.

If developers have a good reason for a condition within a test, then they can disable the rule with an inline disable.

This fixes #1026

I'll do a follow-up PR deprecating `no-if`, so that there's a changelog entry for visibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ installations requiring long-term consistency.
 | [no-alias-methods](docs/rules/no-alias-methods.md)                           | Disallow alias methods                                              | ![style][]       | ![fixable][] |
 | [no-commented-out-tests](docs/rules/no-commented-out-tests.md)               | Disallow commented out tests                                        | ![recommended][] |              |
 | [no-conditional-expect](docs/rules/no-conditional-expect.md)                 | Prevent calling `expect` conditionally                              | ![recommended][] |              |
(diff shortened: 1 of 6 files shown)


SimenB (approved):
I like it

PR #1040, opened 2022-01-31
chore: update semantic-release

SimenB:
We need to force npm to be on v7 due to some change they've made which uses wildcards for all dependencies, leading to pulling in incompatible versions in yarn since they don't respect `bundleDependencies`.

I've reported this to yarn (discord, so no link), and I think they're gonna fix it 🙂 But this fixes a "vulnerability" (which, as always, is complete bullshit, but it is what it is)

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "pinst": "^2.0.0",
     "prettier": "^2.0.5",
     "rimraf": "^3.0.0",
-    "semantic-release": "^18.0.0",
+    "semantic-release": "^19.0.0",
     "semver": "^7.3.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -159,7 +159,9 @@
     ]
   },
   "resolutions": {
-    "@typescript-eslint/experimental-utils": "^5.0.0"
+    "@typescript-eslint/experimental-utils": "^5.0.0",
+    "fsevents/node-gyp": "^7.0.0",
```
SimenB, line 163: added this as well to avoid yarn choosing `latest`

PR #1072, opened 2022-03-24
change node engine version

r04922101:
Follow dependency "@typescript-eslint/utils" node engine version


SimenB (approved):
thanks! also matches eslint v8

PR #1094, opened 2022-04-25
feat: support `@jest/globals`

### LINKED ISSUE #556: Support `@jest/globals`
See https://github.com/facebook/jest/pull/9801.

We need to support finding `expect`, `test`, `it` etc when imported from `@jest/globals`. Should be relatively simple for ESM syntax, possibly a bit more pain for `require`. We need to support aliasing as well `import { test as testFunction } from '@jest/globals'`

Note that the feature hasn't been released yet (blocked by https://github.com/facebook/jest/pull/9806), but we can add support before it lands.

/cc @G-Rath

G-Rath:
I think the core logic is pretty much done - just need to shore it up with tests (including coverage for TS-specific imports), switch over all the rules to use it, and implement support for `require`.

I'm considering landing this in two parts: 1. just switching us over to checking if references are globals, then 2. (re)implementing the special casing of imports from `@jest/globals`, since they are really two different features and that might make it a bit nicer to list in changelogs etc; but will see how the PR ends up looking.

Resolves #556


SimenB:
Yay! Will review tomorrow (currently at a concert) if you want @G-Rath. I see it's draft and red ci, so might not be ready for review?

G-Rath:
Yeah it's not quite ready for a complete review - I tagged you so you knew it was coming incase you were planning on doing a new major soon (since Jest v28 is out) and we spoke about dropping the `jest` globals from `recommended` once we support this; + you might have some ideas/snippets for getting complete coverage (e.g. `ref.defs` is defined as an array, though I'm pretty sure it should always have 0 or 1 elements, but there might be a edge-case I've forgotten about 🤷).

Now that I know you've seen it, I'll remove your review and re-add once it is actually ready for review 😄

SimenB:
Haha, sounds good!

SimenB:
(no offense, but a tool concert is more fun than code 🤘)

G-Rath:
@SimenB I'm getting some weird behaviour when trying to generate a coverage report for `utils.ts` - if I do `yarn run test utils --coverage`, it seems to be generating a report with the files in `coverage/lcov-report/rules/...`, and the actual content of `coverage/lcov-report/rules/utils.ts.html` has this error:

```
Cannot read property 'start' of undefined
TypeError: Cannot read property 'start' of undefined
    at /c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-reports/lib/html/annotator.js:177:41
    at Array.forEach ()
    at annotateBranches (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-reports/lib/html/annotator.js:128:33)
    at annotateSourceCode (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-reports/lib/html/annotator.js:252:9)
    at HtmlReport.onDetail (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-reports/lib/html/index.js:414:33)
    at LcovReport. [as onDetail] (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-reports/lib/lcov/index.js:28:23)
    at Visitor.value (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-lib-report/lib/tree.js:38:38)
    at ReportNode.visit (/c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-lib-report/lib/tree.js:88:21)
    at /c/Users/G-Rath/workspace/projects-oss/eslint-plugin-jest/node_modules/istanbul-lib-report/lib/tree.js:92:19
    at Array.forEach ()
```

If I run the whole test suite with coverage (`yarn run test --coverage`) then the report is fine, with the contents in `coverage/lcov-report/src/...`.

I can reproduce this on `main` - could you see if happens for you too? and if so what'd be the best way to tackle this - should I make a new issue on the jest repo? I don't know if the issue is actually with `jest`, `istanbul` or something else and I don't know if it happens on Jest v28 either.

SimenB:
The changing path is https://github.com/facebook/jest/issues/7176, would be great to fix that.

Unsure of the error, tho...

G-Rath:
Ok I think this should ready for review - couple of things:
  1. Currently this is just about answering the question of "is this a jest function?", not "_what_ jest function is this?" - that means e.g. `import { xit as it } from '@jest/globals';` is correctly determined to be a jest function, but rules will treat it as `it` rather than `xit`
    * All of the info for supporting this is present, we just need to change the function signatures of the utils to be something like `parseIfJestFunction` which returns `JestFunction | null` instead of a boolean
    * ~A small part of me wonders if we should maybe not do this, as it could serve as a way of bailing out~ we should do it, because this is a core feature of ESM imports + you can "bailout" by just assigning to a variable
  2. Similarly, we don't attempt to resolve variables e.g. `const it = jest.it` won't work
    * this is something I think we should support less, but at the same time I actually don't think it'll be as hard as it might first seem (we're already doing this kind of logic in `valid-expect-in-promise` - it'd mainly just be a huge refactor)
  3. Similarly similarly, this means we don't support `jest.*` (and `import jest from '@jest/globals'`)
    * I assume we actually will need to support this at some point, because that's how you get access to other methods e.g. `jest.spyOn`; that also means we need to have the same logic for handling checking that...
  4. This PR also re-writes `isHook` to work the same way as `isTestCaseCall` and `isDescribeCall` - that could easily be cherry-picked into another PR though

----

Having written all that up, I think really this is only half of the work required to really close #556, because supporting `@jest/globals` should also mean we handle other jest methods like `jest.spyOn` and `jest.mock`.

This actually makes me think it might be best to go with my original plan of trying to land support for "true" global scope/reference first, and then follow it up with adding support for `@jest/globals` as an exception to that logic.

@SimenB what do you think? I'm happy to split this up into a couple of different PRs and also land it via `next` (even if it's not a breaking change, having a few changelog entries could be good).

G-Rath:
> Having written all that up, I think really this is only half of the work required to really close https://github.com/jest-community/eslint-plugin-jest/issues/556, because supporting @jest/globals should also mean we handle other jest methods like jest.spyOn and jest.mock.

huh maybe we don't have any rules that actually check for those - just ones that suggest using them....

src/rules/__tests__/valid-describe-callback.test.ts
```diff
@@ -20,6 +20,7 @@ ruleTester.run('valid-describe-callback', rule, {
     'fdescribe("foo", () => {})',
     'describe.only("foo", () => {})',
     'describe.skip("foo", () => {})',
+    '([]).skip("foo", () => {})',
```
SimenB, line 23: seems separate from this pr?

src/rules/__tests__/utils.test.ts
```diff
@@ -1,12 +1,35 @@
...
+    const eslintPackageJson =
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      require(eslintPath) as JSONSchemaForNPMPackageJsonFiles;
+
+    if (eslintPackageJson.version) {
+      const [majorVersion] = eslintPackageJson.version.split('.');
+
+      return parseInt(majorVersion, 10);
+    }
+  } catch {}
```
SimenB, line 26: would be nice to just remove the `try` - if something inside here fails, that error is useful in debugging

src/rules/__tests__/utils.test.ts
```diff
@@ -299,3 +324,561 @@ testUtilsAgainst(
   ],
   'describe',
 );
+
+const hooks = ['beforeAll', 'beforeEach', 'afterEach', 'afterAll'];
+
+ruleTester.run('hooks', rule, {
+  valid: [...hooks, 'beforeAll.each(() => {})'],
```
SimenB, line 331: `beforeAll.each` isn't valid. But I guess the eslint plugin can just ignore it

src/rules/__tests__/utils.test.ts
```diff
@@ -299,3 +324,561 @@ testUtilsAgainst(
...
+  invalid: hooks.map(code => ({
+    code: `${code}(() => {})`,
```
SimenB, line 333:
```suggestion
  invalid: hooks.map(hook => ({
    code: `${hook}(() => {})`,
```

src/rules/utils.ts
```diff
@@ -669,39 +679,41 @@ export const getTestCallExpressionsFromDeclaredVariables = (
...
+  }
+
+  name = resolveToJestFn(scope, name);
+
+  return !!name && TestCaseName.hasOwnProperty(name);
+};
+
+const findFirstCallPropertyName = (
+  node: TSESTree.CallExpression,
+  properties: string[],
```
SimenB, line 713:
```suggestion
  properties: readonly string[],
```

src/rules/utils.ts
```diff
@@ -713,96 +725,193 @@ export const isTestCaseCall = (
...
 
-  const callee =
-    node.callee.type === AST_NODE_TYPES.TaggedTemplateExpression
-      ? node.callee.tag
-      : node.callee.type === AST_NODE_TYPES.CallExpression
-      ? node.callee.callee
-      : node.callee;
+  name = resolveToJestFn(scope, name);
+
+  return !!name && DescribeAlias.hasOwnProperty(name);
```
SimenB, line 772:
```suggestion
  return name !== null && DescribeAlias.hasOwnProperty(name);
```

src/rules/utils.ts
```diff
@@ -649,14 +649,24 @@ export const isFunction = (node: TSESTree.Node): node is FunctionExpression =>
...
+): node is JestFunctionCallExpressionWithIdentifierCallee<HookName> => {
+  let name = findFirstCallPropertyName(node, []);
+
+  if (!name) {
+    return false;
+  }
+
+  name = resolveToJestFn(scope, name);
+
+  return !!name && HookName.hasOwnProperty(name);
```
SimenB, line 664:
```suggestion
  return name != null && HookName.hasOwnProperty(name);
```

src/rules/utils.ts
```diff
@@ -669,39 +679,41 @@ export const getTestCallExpressionsFromDeclaredVariables = (
             (node): node is JestFunctionCallExpression<TestCaseName> =>
               !!node &&
```
SimenB, line 680:
```suggestion
              node != nulll &&
```

src/rules/utils.ts
```diff
@@ -669,39 +679,41 @@ export const getTestCallExpressionsFromDeclaredVariables = (
...
-    return true;
+  let name = findFirstCallPropertyName(node, Object.keys(TestCaseProperty));
+
+  if (!name) {
+    return false;
+  }
+
+  name = resolveToJestFn(scope, name);
+
+  return !!name && TestCaseName.hasOwnProperty(name);
```
SimenB, line 708:
```suggestion
  return name != null && TestCaseName.hasOwnProperty(name);
```

src/rules/utils.ts
```diff
@@ -713,96 +725,193 @@ export const isTestCaseCall = (
...
+};
+
+interface ImportDetails {
+  source: string;
+  local: string;
+  imported: string;
+}
+
+const describeImportDefAsImport = (
+  def: TSESLint.Scope.Definitions.ImportBindingDefinition,
```
SimenB, line 782:
```suggestion
  definition: TSESLint.Scope.Definitions.ImportBindingDefinition,
```

src/rules/utils.ts
```diff
@@ -713,96 +725,193 @@ export const isTestCaseCall = (
...
+  }
 
-    return callee.object.type === AST_NODE_TYPES.MemberExpression
-      ? isDescribeAlias(callee.object.object)
-      : isDescribeAlias(callee.object);
+  return null;
+};
+
+const describeVariableDefAsImport = (
+  def: TSESLint.Scope.Definitions.VariableDefinition,
```
SimenB, line 834:
```suggestion
  definition: TSESLint.Scope.Definitions.VariableDefinition,
```

src/rules/utils.ts
```diff
@@ -713,96 +725,193 @@ export const isTestCaseCall = (
...
+ * Attempts to describe a definition as an import if possible.
+ *
+ * If the definition is an import binding, it's described as you'd expect.
+ * If the definition is a variable, then we try and determine if it's either
+ * a dynamic `import()` or otherwise a call to `require()`.
+ *
+ * If it's neither of these, `null` is returned to indicate that the definition
+ * is not describable as an import of any kind.
+ */
+const describePossibleImportDef = (def: TSESLint.Scope.Definition) => {
```
SimenB, line 872:
```suggestion
const describePossibleImportDef = (definition: TSESLint.Scope.Definition) => {
```

SimenB (approved):
fantastic stuff! 👏

src/rules/utils.ts
G-Rath: A little on the fence about this - I went with `def` because that's what the underlying runtime property that's passed in is called

src/rules/__tests__/valid-describe-callback.test.ts
G-Rath: it's required for coverage, but I can cover it in the `utils` tests (I'd actually done the refactoring to `findFirstCallPropertyName` a while ago, so forgot it also added this 😅)

SimenB (approved):
great stuff 👍

G-Rath:
Turns out this isn't completely right because `context.getScope` is actually stateful:

> returns the [scope](https://eslint.org/docs/developer-guide/scope-manager-interface#scope-interface) of the currently-traversed node. This information can be used to track references to variables.

So while the logic is correct, it only works correctly in `utils` because the test rule does this:

```
const rule = createRule({
  // ...
  create: context => ({
    CallExpression(node) {
      const scope = context.getScope();
      // ...
    },
  }),
});
```

whereas in the rules we're doing:

```
const rule = createRule({
  // ...
  create: context => {
    const scope = context.getScope();

    return {
      CallExpression(node) {
        // ...
      },
    };
  },
});
```

I think for now I'm just going to replace all uses of `scope` with `context.getScope()` to be safe - I want to avoid over saturating our tests with the same extensive subset of tests, but will also chuck a few explicit ones into some of the rules just to be safe.

G-Rath:
Also @SimenB I just realised we should be doing this for `expect` as well right?

SimenB:
Yeah, all globals plus `jest`

G-Rath:
Yeah but I'm pretty sure these + expect are the only ones we actually check for right now 😁

PR #1098, opened 2022-05-05
feat: create `prefer-hooks-in-order` rule

### LINKED ISSUE #992: [prefer-hooks-on-top]: allow enforcing order of hooks
I using this rule https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-hooks-on-top.md 

Would be nice have possibility to specify exact order for hooks, for example:
`order: ['beforeAll', 'beforeEach', 'afterEach', 'afterAll']`

aaron012:
The tests on your branch were a major help for implementing this @G-Rath. Thanks!

Resolves #992


SimenB:
there's a conflict, could you resolve? 🙂

G-Rath (changes requested):
Looks good - could you throw in some more complex examples (the correct and incorrect samples from the doc would be a good start - we want to make sure that the tracking works correctly around nested cases; also can you chuck in an example with multiple hooks of the same name?), and you need to apply formatting + regenerate the docs (`yarn run tools:regenerate-docs`).

src/rules/prefer-hooks-in-order.ts
```diff
@@ -0,0 +1,74 @@
...
+          // Ignore everything that is passed into a hook
+          return;
+        }
+
+        if (!isHookCall(node, context.getScope())) {
+          // Reset the previousHookIndex when encountering something different from a hook
+          previousHookIndex = -1;
+        }
+
+        if (isHookCall(node, context.getScope())) {
```
G-Rath, line 41: We already know that `node` is a hook at this point, because we're just tested if it _isn't_ so if you add an early return to the above `if !isHookCall` you should be able to drop this check and remove a level of indentation

src/rules/prefer-hooks-in-order.ts
```diff
@@ -0,0 +1,74 @@
...
+          inHook = true;
+          const currentHook = node.callee.name;
+          const currentHookIndex = HooksOrder.findIndex(
+            name => name === currentHook,
+          );
```
G-Rath, line 46: nit: you should be able to use `indexOf` here right?

```suggestion
          const currentHookIndex = HooksOrder.indexOf(currentHook);
```

src/rules/prefer-hooks-in-order.ts
```diff
@@ -0,0 +1,74 @@
...
+          }
+        }
+      },
+      'CallExpression:exit'(node) {
+        if (!isHookCall(node, context.getScope()) && !inHook) {
+          // Reset the previousHookIndex when encountering something different from a hook
+          previousHookIndex = -1;
+        }
+
+        if (isHookCall(node, context.getScope())) {
```
G-Rath, line 68: Again, we're doubling up our conditional check here - if you use an early return we should be able to just do the check once.

src/rules/prefer-hooks-in-order.ts
```diff
@@ -0,0 +1,74 @@
...
+          if (currentHookIndex < previousHookIndex) {
+            context.report({
+              messageId: 'reorderHooks',
+              node,
+              data: {
+                previousHook: HooksOrder[previousHookIndex],
+                currentHook,
+              },
+            });
+          } else {
```
G-Rath, line 57: nit: personally I avoid using `else` as I think early returns are better than having an extra level of indenting

G-Rath (commented):
Looks good - I've got a few suggestions & nitpicks which'd be good to be addressed before merging, but I don't think need to _stop_ merging if it comes to it.

Let me know if you're happy to action them or if you'd like me to :)

aaron012:
> Looks good - I've got a few suggestions & nitpicks which'd be good to be addressed before merging, but I don't think need to _stop_ merging if it comes to it.
> 
> Let me know if you're happy to action them or if you'd like me to :)

Thanks! I have fixed those 😄

G-Rath (approved):

Issue #1118, opened 2022-05-19
Support `test.failing`


G-Rath:
https://jestjs.io/docs/api#testfailingname-fn-timeout

Looks like this is only on `test` & `it` (not `describe`). Should be a matter of updating the enum.

PR #1122, opened 2022-05-28
feat: improve how jest function calls are resolved to account for import aliases

### LINKED ISSUE #1106: Support resolving underlying `@jest/globals` method from import
While we now support `@jest/globals` via #1094, we currently are still doing our checks against the "local name" of whatever is being imported, e.g.

```
import { xit as it } from '@jest/globals';

it('is skipped', () => {
  //
});
```

For the above rules see this as importing `it` rather than `xit` - we're already resolving this information as part of our reference scope checking, we just need to expose it.

This will make the API a bit more cumbersome, because we have to change from a boolean type guard check to an "extract" based method (similar to `parseExpectCall`) but I think it's worth it.

This is required for #1101

G-Rath:
Resolves #1106 and #1119, and has highlighted a few other improvements (like that `prefer-todo` ignores `x` prefixed methods even though it applies to `.skip`'d methods, and the ability to support `context` as an alias of `describe`).

This overhauls how we parse jest methods to leverage the redesign of how to check if something is a jest method that was landed via #1094 so that rules can handle aliased functions.

Because of how this turned out I've ended up doing away with most of the `utils` tests as they're now covered by tests in rules - I have only brought them back to cover cases that are not specific to one particular rule.

I've kept this to being just focused on implementing #1106 even though I could also do a few more things like add support for `failing` (#1118), support parsing `expect`, remove support for invalid `concurrent` call chains, etc. I'll be doing follow up PRs straight away for these so that they have their own changelog.

Support for parsing `expect` calls in particular will be my next focus as I've got that mostly done already but realised a little bit of extra thinking needs to be done because `expect` usage is generally inverse to other methods since it goes `expect(..).property.call(...)` - I would like to merge the existing parser/api into this new one, but need to decide on a pattern to represent the difference between `expect` calls where `expect` itself is called (which is what the current parser handles) and `expect` calls where it's not (which is what the new parser is really looking for, in all cases except for `expect`).

I also realised while doing this that `spyOn` and co are still accessed via `jest` even when using `@jest/globals` - you just import `jest` (e.g. it's `import { jest } from '@jest/globals';`, not `import { spyOn } from '@jest/globals';`)

Also also while not super important, this makes `lib` ~4kb smaller which I think is a good sign that it's on the right track :D


src/rules/utils.ts
```diff
@@ -6,6 +6,7 @@ import {
   TSESTree,
 } from '@typescript-eslint/utils';
 import { version } from '../../package.json';
+import { isTypeOfJestFnCall } from './utils/parseJestFnCall';
```
SimenB, line 9: Rename _this_ file to `utils/index.ts`?

SimenB (approved):
Woah, amazing!

src/rules/utils.ts
G-Rath: I was planning on moving everything else in a follow-up PR, as I'll split things across a few more files but wanted to wait until I'd refactored `expect` parsing since that represents about 1/3rd of the file.

PR #1125, opened 2022-05-28
fix(no-disabled-tests): switch to using jest function call parser

### LINKED ISSUE #1115: [no-disabled-tests] does not support `@jest/globals`
Currently this rule uses AST selectors & `getNodeName` to target Jest functions, which means it doesn't support `@jest/globals` - it should be switched to using our scoped utils.

G-Rath:
To be merged after #1122 is landed

Resolves #1115

### DIFF
--- a/src/rules/no-disabled-tests.ts
+++ b/src/rules/no-disabled-tests.ts
@@ -1,4 +1,9 @@
-import { createRule, getNodeName, scopeHasLocalReference } from './utils';
+import {
+  createRule,
+  getAccessorValue,
+  parseJestFnCall,
+  scopeHasLocalReference,
+} from './utils';
 
 export default createRule({
   name: __filename,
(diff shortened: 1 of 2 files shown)


src/rules/no-disabled-tests.ts
```diff
@@ -27,40 +30,49 @@ export default createRule({
...
     return {
-      'CallExpression[callee.name="describe"]'() {
-        suiteDepth++;
-      },
-      'CallExpression[callee.name=/^(it|test)$/]'() {
-        testDepth++;
-      },
       'CallExpression[callee.name=/^(it|test)$/][arguments.length<2]'(node) {
         context.report({ messageId: 'missingFunction', node });
       },
```
G-Rath, line 35: ugh I realised I missed this somehow - will do a follow-up PR

Issue #1033, opened 2022-01-22
[unbound-method] Allow custom assert function names


SchroederSteffen:
Thanks for the plugin! We're happily using it in multiple TypeScript projects.

Though, we recently came across the issue that the `unbound-method` rule also complains if we're passing a method into a custom expect function like this one:
```
  const expectCallbackToReturn = (
    method: Class['method'],
    expectedReturnValue: string
  ) => {
    const mock = (method as jest.Mock).mock;
    const [callback] = mock.calls[0] as Parameters<typeof method>;
    expect(callback()).toBe(expectedReturnValue);
  };

[...]

    expectCallbackToReturn(
      // Error: Avoid referencing unbound methods
      mockInstance.method,
      'expected-value'
    );
```

We wrote this custom expect function to reuse it across a test file, but writing a custom matcher seems like overhead because we need it only in a single file.

A possible solution could be if we could specify custom `assertFunctionNames` like for `expect-expect`. Then, we would simply allow `expect*`.

G-Rath:
I'm really on the fence about this one because this is actually dependent on more than just the name of the function, and I'm pretty sure the example you've given could be better structured using the mock types, e.g. something like:

```
  const expectCallbackToReturn = (
    mock: jest.Mocked<Class['method']>,
    expectedReturnValue: string
  ) => {
    const [callback] = mock.calls[0];
    expect(callback()).toBe(expectedReturnValue);
  };

[...]

    expectCallbackToReturn(
      mockInstance.method.mock,
      'expected-value'
    );
```

We want to be careful with this rule because the more we customize it the harder it becomes to maintain with the base rule, and I'm not sure if there's enough value here vs the alternatives to be worth it.

SchroederSteffen:
Thanks for the response. I understand the concern.
Do you agree that it should be possible to write convenience assert/expect function to avoid code duplication? If there would be a working pattern for this, we wouldn't need to adjust the rule.

Unfortunately, I couldn't get your suggestion to work, because the property `calls` doesn't exist on type `jest.Mocked<Class['method']>`, at least on my machine.
Though, I was able to improve the function as follows: 👍 
```ts
  const expectCallbackToReturn(
    mockedFunction: jest.MockedFunction<Class['method']>,
    expectedReturnValue: string
  ) => {
    const [callback] = mockedFunction.mock.calls[0];
    expect(callback()).toBe(expectedReturnValue);
  };
```
[...]

PR #1129, opened 2022-05-28
feat: support aliases for jest globals (e.g. `context`)

### LINKED ISSUE #83: Are you open to support `context` as `describe` alias?
Up until around 1 year ago, `context` was an officially supported `describe` alias, but later [got removed](https://github.com/facebook/jest/issues/2468). Still, because of its value, some people (myself included) still want to use it, and as explained on that issue [it's very easy to add it back](https://github.com/facebook/jest/issues/2468#issuecomment-269718083).

Unfortunately, it's not compatible with using this ESLint plugin (and I really like this plugin :) ). Would you be open to merge a PR adding `context` (and derived `context.only`, …) to the [describe aliases](https://github.com/jest-community/eslint-plugin-jest/blob/master/rules/util.js#L89) and other places where it would be needed?

I understand maybe you wouldn't want to have defaults that are different to what jest itself has in terms of globals, but if so, maybe adding it as a configurable option instead? Afaik, the plugin itself cannot receive a configuration though, so apparently that would need to be done by passing options to each applicable rule.

If you are open to this I can give a PR a try. Or if you have any other suggestion on how to use ´context´ and this plugin at the same time, I would love to hear it.

Thanks for this plugin!

G-Rath:
I'm a bit on the fence about this: it ended up being a pretty straightforward addition but I'm worried it's catering to a very small niche because: 
1. it requires you to set up the globaled aliases with a custom `setupTestFrameworkScriptFile`
2. you can get this already by importing the function you want to alias from `@jest/globals`
    * in addition to being a requirement if you're using ESM (since no globals), I think it's straight up easier than 1 as devs will already know how to import but would need to readup on jest configuration to know how to do 1

@SimenB do you think this is worth landing?

Resolves #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,25 @@ doing:
 This is included in all configs shared by this plugin, so can be omitted if
 extending them.
 
+#### Aliased Jest globals
+
+You can tell this plugin about any global Jests you have aliased using the
+`globalAliases` setting:
+
+```json
+{
+  "settings": {
+    "jest": {
+      "globalAliases": {
+        "describe": ["context"],
+        "fdescribe": ["fcontext"],
+        "xdescribe": ["xcontext"]
+      }
+    }
+  }
+}
+```
+
 ### Running rules only on test-related files
 
 The rules provided by this plugin assume that the files they are checking are
--- a/src/rules/__tests__/no-focused-tests.test.ts
+++ b/src/rules/__tests__/no-focused-tests.test.ts
@@ -45,6 +45,23 @@ ruleTester.run('no-focused-tests', rule, {
       ],
     },
     {
+      code: 'context.only()',
+      errors: [
+        {
+          messageId: 'focusedTest',
+          column: 9,
+          line: 1,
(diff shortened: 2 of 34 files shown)


G-Rath:
hmmm having this could allow us to deprecate the `additionalTestBlockFunctions` option that we've got on a few tests, which should reduce the logic of our rules and maybe improve some other rules for people (e.g. #534) without us having to do anything; so that could be another reason to land it.

SimenB (approved):
I think this seems reasonable 👍

G-Rath:
@SimenB do you think it's worth us checking if we can remove some of our rule options in favor of this?

G-Rath:
double hmmm: actually if we made this support checking local declarations (_just_ for aliases - jest native functions would still be expected to be either globals or imported from `@jest/globals`), I think it should allow all of our rules to be customisable when combined with `overrides`.

This'd definitely let us drop a few rule-specific options, and resolve e.g. #1033

PR #1132, opened 2022-05-29
fix: don't consider method calls on literals or `new` to be jest functions

### LINKED ISSUE #1128: v26.4.2 is treating any file including `/regex/.test(arg)` as a test file
Updated to 26.4.2 through renovate bot and noticed my CI failed because this linter was running on non-test files.

I came across this issue https://github.com/jest-community/eslint-plugin-jest/issues/555 which explains that anything that uses jest testing functions is treated as a test file.

The non-test files in question are simply doing regex matching of the kind e.g. `/regex/.test(arg)`, which I would expect to not trigger this heuristic to treat it as a test file.

I reverted to 26.3.0 for now without any other changes and it's working. Just wanted to give you a heads up.

Thanks for working on this linter!

### LINKED ISSUE #1130: false positive lint error introduced between 26.1.5 and 26.4.2
When I upgraded from 26.1.5 to 26.4.2 I started to get this false positive:
image

G-Rath:
I was trying to be a little clever, but there's a reason why `getNodeName` does this...

I've also included removing allowing `NewExpression` because that's also not correct.

Resolves #1128
Resolves #1130

### DIFF
--- a/src/rules/utils/__tests__/parseJestFnCall.test.ts
+++ b/src/rules/utils/__tests__/parseJestFnCall.test.ts
@@ -114,6 +114,10 @@ ruleTester.run('nonexistent methods', rule, {
     'describe.concurrent()()',
     'describe.concurrent``()',
     'describe.every``()',
+    '/regex/.test()',
+    '"something".describe()',
+    '[].describe()',
+    'new describe().only()',
   ],
   invalid: [],
 });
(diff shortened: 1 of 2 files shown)


G-Rath:
I've also just found we're matching on `test.only``()` - will fix that in another PR since it's a different thing

PR #1166, opened 2022-07-13
feat: create `max-expects` rule

### LINKED ISSUE #1163: New rule: `max-expects`
Like [`ava/max-asserts`](https://github.com/avajs/eslint-plugin-ava/blob/main/docs/rules/max-asserts.md), I thought it would be nice to have a rule recommending smaller tests. (Couldn't figure out if there are any existing rules that could be substituted.)

makotot:
Closes #1163.
I have added the new rules proposed in #1163. The default value of `max` option is the same value as in [eslint-plugin-ava](https://github.com/avajs/eslint-plugin-ava/blob/main/docs/rules/max-asserts.md#options).

### DIFF
--- a/src/rules/max-expects.ts
+++ b/src/rules/max-expects.ts
@@ -43,8 +43,6 @@ export default createRule({
 
       if (isTestFn) {
         count = 0;
-
-        return;
       }
     };
 


src/rules/max-expects.ts
```diff
@@ -0,0 +1,89 @@
...
+        node.parent.type !== AST_NODE_TYPES.CallExpression ||
+        isTypeOfJestFnCall(node.parent, context, ['test']);
+
+      if (isTestFn) {
+        count = 0;
+
+        return;
+      }
+    };
+    const onFunctionExpressionExit = (node: FunctionExpression) => {
```
G-Rath, line 54: We don't need this, since we're already resetting the count whenever we enter a test function and test functions cannot be nested so we don't need to track a stack.

src/rules/max-expects.ts
```diff
@@ -0,0 +1,89 @@
...
+      'FunctionExpression:exit': onFunctionExpressionExit,
+      ArrowFunctionExpression: onFunctionExpressionEnter,
+      'ArrowFunctionExpression:exit': onFunctionExpressionExit,
+      CallExpression(node) {
+        if (isExpectCall(node)) {
+          count += 1;
+        }
+        if (count > max && node.parent) {
+          context.report({
+            node: node.parent,
```
G-Rath, line 81: nit: I'm not sure if we should be reporting on the parent or the node itself - it doesn't matter too much except it lets us drop an impossible condition (TS doesn't know that `parent` is always present except for `Program` type nodes).

src/rules/max-expects.ts
```diff
@@ -0,0 +1,89 @@
...
+        'Too many assertion calls ({{ count }}). Maximum allowed is {{ max }}.',
+    },
+    type: 'suggestion',
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          max: {
+            type: 'integer',
+            minimum: 0,
```
G-Rath, line 28: What do you think about having 1 as the minimum? I can't think of a situation where you'd ever want to disallow _any_ exceptions (even if you were wanting to not use `expect`, there are better rules for enforcing that)

```suggestion
            minimum: 1,
```

src/rules/max-expects.ts
```diff
@@ -0,0 +1,89 @@
...
+  isExpectCall,
+  isTypeOfJestFnCall,
+} from './utils';
+
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Enforces a maximum assertion calls in a test body',
```
G-Rath, line 14:
```suggestion
      description: 'Enforces a maximum number assertion calls in a test body',
```

G-Rath (changes requested):
Looks good - got a few tweaks to make, and we need a few more test cases to get full coverage e.g. used with `describe`, multiple `test` blocks in the same "file", and one with `test.each` would probably good too.

src/rules/max-expects.ts
makotot: Fixed in https://github.com/jest-community/eslint-plugin-jest/pull/1166/commits/b9e4ca486fe43e334ba3e8a869a0d6e7aba6668b

src/rules/max-expects.ts
makotot: Fixed in https://github.com/jest-community/eslint-plugin-jest/pull/1166/commits/29174fd4aae37610c9b869ec836d336cb7bd174f and https://github.com/jest-community/eslint-plugin-jest/pull/1166/commits/589b7d48f0abe0f70fe8d37d01f8c3eee5d17d0a.

src/rules/max-expects.ts
makotot: Fixed in https://github.com/jest-community/eslint-plugin-jest/pull/1166/commits/8bba714d62e447a1680ae4df59017c2fefd55a18. I too think 1 is the appropriate minimum value.
But I can't figure out if something would be needed that would stop the process when a value less than 1 is passed.

src/rules/max-expects.ts
makotot: Fixed in https://github.com/jest-community/eslint-plugin-jest/pull/1166/commits/d96cca7a682d639b072ea3097912178f90cbb610

makotot:
@G-Rath 
Thanks for the review! 
When I added more tests to meet coverage, I found that the implementation for multiple "test" blocks was incorrect, so have added 79bb73484f6ce6c26df82c9a7ca726c864a84091 to fix it.

src/rules/max-expects.ts
makotot: I didn't understand how validation works. Thanks for explaining it to me!

Issue #1101, opened 2022-05-10
[new-rule] prefer importing from @jest/globals


ecraig12345:
It would be nice to have a rule (with fixer) enforcing that APIs such as `expect`, `describe`, `it`, `jest`, etc should be imported from `@jest/globals`. If there's interest in this rule, I'd probably be willing to implement it.

Motivation:
- Reduce global type pollution by not needing to put `"types": ["jest"]` in a tsconfig
- Allow removing dependency on `@types/jest` which can be outdated, mismatched, and/or pull in extra versions of packages (this is an issue as of writing--jest 28 has released but the types are still on 27 and pull in some outdated deps from 27)
- Unblock adoption of [future mode with globals unavailable](https://jestjs.io/blog/2020/05/05/jest-26#a-new-way-to-consume-jest---jestglobals) once that's implemented (I'm not sure where that is on their roadmap)

Here's an example of code with the rule enabled, and an error it might catch:
```js
import { describe, it, expect, jest } from '@jest/globals';
import { foo } from './foo';

describe('foo', () => {
  // ERROR - you forgot to import beforeAll
  beforeAll(() => {
    jest.useFakeTimers();
  });

  it('does stuff', () => {
    expect(foo()).toBe(0);
  });
});
```

(I assume this would be blocked by #556, but looks like there's an approved PR #1094 addressing that.)

G-Rath:
I think you should be able to handle this with the existing tooling and education - e.g. so long as you don't have `@types/jest` installed, Typescript will error (you just have to ignore it's suggestion to install `@types/jest`).

Having said that the core of the work for this is now already done via the PR you mentioned, so a rule shouldn't be expensive (though I don't know if we could have a autofix to add the import without adding a bunch of complexity).

What is missing first is the follow up work I'm going to do: moving from a conditional check to returning information about the reference (which is needed to handle `import { xit as skip } from ...` sort of stuff) - once that's done this rule should be a matter of adding a `isImported` boolean prop to that return and then have a rule which checks for that...

Im planning to merge that PR and then do this follow up work over the next couple of weeks :) (I've had a few things pop up at work and need to do some stuff on my `osv-detector` project, which is why I've not merged yet)
It will be best to wait until I've done the follow to that PR, as what

SimenB:
[...]

PR #1169, opened 2022-07-19
fix: make no-jest-import error message more accurate

mmkal:
It isn't quite accurate to say that Jest doesn't export anything. It exports the `run` and `runCLI` functions, even though it isn't generally recommended to use them ([yet](https://github.com/facebook/jest/issues/5048#issuecomment-470693540)).

This updates the message to match the language in https://jestjs.io/docs/jest-object ("The `jest` object is automatically in scope within every test file").

There's also a similar update to the no-jest-import.md docs file.

### DIFF
--- a/docs/rules/no-jest-import.md
+++ b/docs/rules/no-jest-import.md
@@ -2,7 +2,7 @@
 
 The `jest` object is automatically in scope within every test file. The methods
 in the `jest` object help create mocks and let you control Jest's overall
-behavior. It is therefore completely unnecessary to import in `jest`, as Jest
+behavior. It is therefore usually unnecessary to import in `jest`, as Jest
 doesn't export anything in the first place.
 
 ### Rule details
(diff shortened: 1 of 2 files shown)


SimenB:
I'm fine with either wording, @G-Rath wdyt?

G-Rath:
tbh I think we should be tossing the rule all together as it's no longer accurate:
  * `require('jest')` doesn't give you an object with the same properties as what most people want (e.g. `test`, `spyOn`, `fn`, etc) so it should be very obvious something wrong is being done
  * now that we have `@jest/globals`, it is valid to import `jest` from there (and my understanding is that's required in ESM)
  * you can use [`no-restricted-imports`](https://eslint.org/docs/latest/rules/no-restricted-imports) to enforce `@jest/globals` isn't used, and you can set `injectGlobals` to false to enforce using `@jest/globals`
      * though we want a new rule that enforces using `@jest/globals` (tracked by #1101, and should be coming soon)

That middleone at least means the proposed new message isn't actually true, because it depends on if you have `injectGlobals` set.

PR #1173, opened 2022-07-24
feat: resolve `expect` based on scope

G-Rath:
This switches our handling of `expect` over to match the rest of our we now resolve jest functions, meaning it correctly respects `@jest/globals` & the related patterns that come with that, done through `parseJestFnCall`.

This required a lot more work than the others as currently we handle `expect` calls from the `expect(...)` call itself whereas for e.g. `test` we do it at the "top" call expression (`test.only(...)`) - inverting this for `expect` had some interesting impacts, the biggest of which is to `valid-expect` because now the act of parsing effectively enforces a "valid" `expect` statement, which now is effectively a wrapper for exposing the "reason" that `parseJestFnCall` internally now returns (which is to keep `valid-expect` usable in the first place, and is only exposed with `parseJestFnCallWithReason`).

Elsewhere some places have become easier (`unbound-method` & `no-restricted-matchers` especially) but it's also gotten more technical, so overall I think there's no net loss or gain (which is fine).

This does come with a minor performance hit since we're doing a decent amount more work - as a basic test I checked the times of running `jest/all` on all file types in the `jest` monorepo, and we're looking at about an extra 10 seconds; also importantly `parseJestFnCall` has a significant enough overhead that we need to be careful about calling it multiple times unless we really have to (I was able to clearly optimize the time of `no-standalone-expect` & `prefer-expect-assertions` this way).

While I think the current performance hit doesn't bother me (it's still less than 45 seconds on average, and when you include other rules that gets to 60s+) but that second point does as it's something we have to watch out for - luckily it looks like we can easily fix this by just chucking in a cache, which brings the times down to _10 seconds_; I'm planning to land that in another PR to keep things clean (+ we'd have a cacheless version we can baseline against).

I'm opening this as a draft because I want to do some cleanup before it's merged and there are edge-cases to think about (especially in `valid-expect`), but it should be reviewable already. I focused first on just replacing `parseExpectCall` with the low level API, to avoid premature optimizing - I plan to go back through the rules and identity common usage/patterns that we can expose on the parsed call, such as accessing the arguments for expect and checking modifiers nodes.

As expected this has also identified a few edge-case bugs which I might try and land ahead of time to make things a bit easier.

I'm also going to run the smoke test against this branch.


G-Rath:
I toyed around with adding some helpers to the parsed jest call e.g. `jestFnCall.hasMember('each')` which were neat but I've not included them for now because our rules are pretty stable these days and they didn't reduce the overall compiled size so don't think they're worth landing right now.

I might add them in the future if we get some more rules that would benefit from them.

src/rules/__tests__/no-restricted-matchers.test.ts
```diff
@@ -70,7 +70,7 @@ ruleTester.run('no-restricted-matchers', rule, {
       ],
     },
     {
-      code: 'expect(a).not',
+      code: 'expect(a).not[x]()',
```
SimenB, line 73: shouldn't we keep the old test?

src/rules/__tests__/no-standalone-expect.test.ts
```diff
@@ -20,7 +20,6 @@ ruleTester.run('no-standalone-expect', rule, {
     'it("an it", () => expect(1).toBe(1))',
     'const func = function(){ expect(1).toBe(1); };',
     'const func = () => expect(1).toBe(1);',
-    'expect.hasAssertions()',
```
SimenB, line 23: ?

src/rules/__tests__/no-standalone-expect.test.ts
```diff
@@ -59,27 +58,31 @@ ruleTester.run('no-standalone-expect', rule, {
   ],
   invalid: [
     {
-      code: "(() => {})('testing', () => expect(true))",
-      errors: [{ endColumn: 41, column: 29, messageId: 'unexpectedExpect' }],
+      code: "(() => {})('testing', () => expect(true).toBe(false))",
+      errors: [{ column: 29, messageId: 'unexpectedExpect' }],
```
SimenB, line 62: why no `endColumn`

src/rules/__tests__/prefer-equality-matcher.test.ts
```diff
@@ -32,6 +32,10 @@ const expectSuggestions = (
 
 ruleTester.run('prefer-equality-matcher: ===', rule, {
   valid: [
+    'expect.hasAssertions',
+    'expect.hasAssertions()',
+    'expect.assertions(1)',
+    'expect(true).toBe(...true)',
```
SimenB, line 38: looks like a syntax error to me 😅

SimenB (approved):
exciting stuff!

src/rules/__tests__/no-standalone-expect.test.ts
G-Rath: [This has become an error](https://github.com/jest-community/eslint-plugin-jest/pull/1173/files#diff-4d19ed2eeb02828caa2177ad7b043e207d6d773e56452b5dde658bfe5f7a17b6R65), since it's a standalone `expect`.

src/rules/__tests__/no-standalone-expect.test.ts
G-Rath: Eh I typically don't include it as I don't think its super valuable since we can't control how the results are used in tools & editors so it just creates more work for us without much gain, but yeah I should include it here for consistency since the other tests have it 😅

src/rules/__tests__/no-restricted-matchers.test.ts
G-Rath: The new parser doesn't consider the old test to be a valid expect call, and we've already got `expect(a).rejects` as a `valid` test which covers this change 🤷

PR #1181, opened 2022-07-29
chore: sort package.json

G-Rath:
It's been a while since we've run this, and so there's a few new keys supported now.


SimenB:
@G-Rath can we run this as part of CI?

PR #1182, opened 2022-07-30
ci: adjust caching to apply to all OSs

G-Rath:
@SimenB the OS is included in the cache key, so currently we're only generating a cache for `ubuntu` even though we're running on MacOS & Windows too.


SimenB (changes requested):
We should do same as in jest, with separate cache building runs.

https://github.com/facebook/jest/pull/12840

SimenB (approved):
👍

PR #1196, opened 2022-08-15
chore: rollback eslint-plugin-eslint-plugin

SimenB:
Plugin doesn't report them anymore


SimenB:
Oh, v5 doesn't support node 12, so #1165 should never have landed

https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/releases/tag/v5.0.0

PR #1202, opened 2022-08-20
ci: remove `#` from label colors

G-Rath:
I'm pretty sure this is why labels are being created without the color - I've gone with `replace` because having the `#` means IDEs like IntelliJ/WebStorm will still show their colorpickers for the actual colors.


G-Rath:
Confirmed:

image

Issue #1210, opened 2022-08-22
Next major


SimenB:
I'm thinking it's about time to make a new release dropping node 12 (and 17) and eslint 6 so that we can update our dev deps. @G-Rath any other breaking changes you wanna get it at the same time?

G-Rath:
Yup:
  * https://github.com/jest-community/eslint-plugin-jest/issues/1172: I don't think it's breaking per say, but if we're doing a new major might as well include it
  * https://github.com/jest-community/eslint-plugin-jest/issues/1156#issuecomment-1160825857: I think changing the behaviour to be an entire matcher makes better sense in the long run because it gives more flexibility and it just means you need to repeat a couple of lines which can easily be done with `.map` and shouldn't need updating so not really an extra burden
    * technically we could then remove `prefer-called-with`, though I'm hesitant as it makes it a little harder to use in shared configs (since if teams decide to use `no-restricted-matchers` they have to make sure to apply the config from the shared config too)
  * we should talk about what we do with `no-jest-import` now that we've got support for `@jest/globals` - I think tossing it is probably the right thing to do? https://github.com/jest-community/eslint-plugin-jest/pull/1169#issuecomment-1193380757 (we could always reuse the name later)

I think that's all the breaking stuff, but will do a double check this week.

Also how would you feel if I made an issue over on Jest pushing (again) for removing the alised matchers, given we've got `no-alias-methods` as an autofix? then we could remove or deprecate that rule here in favor of `no-restricted-matchers` or even `no-deprecated-functions` (either in this major or the next).

SimenB:
All those sounds good to me!

Happy to remove aliased matchers in _next_ major of jest (30), but not the imminent 29 (this week)

G-Rath:
Cool - 30 is a nice round number anyway 😄 I'll make an issue on Jest and do follow up stuff later this week. Do you want to get a `next` branch going?

SimenB:
https://github.com/jest-community/eslint-plugin-jest/tree/next

G-Rath:
Have created https://github.com/facebook/jest/issues/13164

G-Rath:
@SimenB I think we should add `no-alias-methods` to recommended since we're going to remove from Jest "soon" (technically they're deprecated functions but iirc `no-deprecated-functions` is not yet setup for `expect` matchers - I guess we should think about what we want to do for that)

SimenB:
works for me 👍

[...]

PR #1211, opened 2022-08-23
feat: drop support for Node versions 12 and 17

SimenB:
As discussed in #1210


package.json
```diff
@@ -153,6 +153,6 @@
   },
   "packageManager": "yarn@3.2.2",
   "engines": {
-    "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
+    "node": "^14.15.0 || ^16.10.0 || >=18.0.0"
```
SimenB, line 156: matches upcoming Jest 29: https://github.com/facebook/jest/blob/187566a70aa4b6aa5f74952b504bbeddb5854aef/package.json#L169-L171

PR #1218, opened 2022-08-27
feat(no-restricted-matchers): match based on start of chain, requiring each permutation to be set

### LINKED ISSUE #1156: [no-restricted-matchers] allow restricting matchers without specific modifier
Hy guys, I have an issue

If I want to restrict one of matchers except it is negated, I can't tell the config that.

For example: `toHaveBeenCalled` except `not.toHaveBeenCalled`

Can you add this little feature to `no-restricted-matchers` rule?

G-Rath:
BREAKING CHANGE

Resolves #1156

This messed my head up a little trying to come up with an intuitive interface that let you shortcut the config a bit i.e. by allowing `*` * `**`, but it always came out feeling brittle and hard to explain or dropping support for restricting modifiers; in the end I realised changing the check to just be `startsWith`  gets us what we want while being easy to understand.

### DIFF
--- a/docs/rules/no-restricted-matchers.md
+++ b/docs/rules/no-restricted-matchers.md
@@ -8,8 +8,9 @@ alternatives.
 Bans are expressed in the form of a map, with the value being either a string
 message to be shown, or `null` if the default rule message should be used.
 
-Both matchers, modifiers, and chains of the two are checked, allowing for
-specific variations of a matcher to be banned if desired.
+Bans are checked against the start of the `expect` chain - this means that to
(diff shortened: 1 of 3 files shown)


cascornelissen:
@G-Rath, before creating an issue. I feel like this change has unintended side effects. With the following item in the map we're also getting errors on `toBeDefined`, `toBeUndefined`, etc. Could this be related now that the check is done with startsWith?

```
'toBe': "Use `toEqual()` instead.",
```

G-Rath:
@cascornelissen yeah that's a big - create an issue and I'll try to get to it in the morning

PR #1226, opened 2022-08-28
fix: consistent rule doc notices and sections

bmish:
This PR adds the following standardized rule doc notices:

* 💼 This rule is enabled in the following [configs](https://github.com/jest-community/eslint-plugin-jest#shareable-configurations): `all`, `recommended`.
* 🔧 This rule is automatically fixable using the `--fix` [flag](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) on the command line.
* 💡 This rule provides [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) that can be applied manually.
* ❌ This rule is deprecated.

This improves usability by enabling users to clearly and easily find out if a rule is applying to their codebase given their configuration as well as other key rule details like the availability of auto-fixers or auto-suggestions that can help with addressing violations. 

All these notices are tested to ensure they are in the correct position and not forgotten. `yarn tools:regenerate-docs` will also automatically add or update these notices in each rule doc (some inspiration taken from eslint-plugin-unicorn which has a similar command). 

It's common for ESLint and many popular ESLint plugins to include notices like these. Some examples (note: I am a maintainer on some of these plugins and have been working to standardize rule documentation):
* https://eslint.org/docs/latest/rules/no-unused-vars
* https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unknown-property.md
* https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-destructuring.md
* https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md
* https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-schema.md

I also added tests about a few other aspects of rule docs:
* The title matches the rule description
* Has a `Rule details` section
* Has an `Options` section only if the rule has options and mentions any named options


G-Rath (changes requested):
I'll give this a more detailed review later but we've already got a tool for updating our docs which runs in CI so it already covers at least some of what you've added tests for, and automatically does the fixing so want to favor that - would you be interested in expanding that to generate what you've manually added to these docs?

I love the idea of standardizing rule documentation but am wondering if it would be best to try and submit some sort of loose RFC/template so that we could see all the different parts + make it easy to write a tool to help generate them.

@MichaelDeBoey this feels like the sort of thing that would tie in fantastically with the new ESLint community stuff that got landed, so interested in if you have thoughts.

bmish:
@G-Rath thanks for your initial thoughts.

1. Yes, I can update the rule doc generator to generate these notices as well. That will take some refactoring to make the generator a bit more sophisticated, so I'll work on that over the next week. If you like the updates to the rule docs, it could be worth be merging this first to avoid potential merge conflicts, and then I could create a separate dedicated follow-up PR to focus only on switching to automatic notice generation.
2. I'm definitely a fan of standardizing the rule documentation as I called out in the PR description. But I'm not sure an ESLint RFC would be appropriate as the ESLint team isn't really in the business of enforcing a documentation standard for plugins. However, we do have [generator-eslint](https://github.com/eslint/generator-eslint), and I like the idea of updating its [rule doc template](https://github.com/eslint/generator-eslint/blob/main/rule/templates/_doc.md) to include notices, so I will plan to do that.

src/__tests__/rules.test.ts
```diff
@@ -76,4 +77,138 @@ describe('rules', () => {
...
+
+      return configNames.sort();
+    }
+
+    function configNamesToList(configNames: string[]) {
+      return `\`${configNames.join('`, `')}\``;
+    }
+
+    ruleNames.forEach(ruleName => {
+      it(ruleName, () => {
```
SimenB, line 123: should use `it.each`: https://jestjs.io/docs/api#testeachtablename-fn-timeout

bmish: Neat, updated to use that.

bmish:
@G-Rath good news, I went ahead and updated the `yarn tools:regenerate-docs` command to automatically add or update the rule notices in each rule doc.

README.md
```diff
@@ -215,6 +215,7 @@ installations requiring long-term consistency.
 | [no-focused-tests](docs/rules/no-focused-tests.md)                           | Disallow focused tests                                              | ![recommended][] | ![suggest][] |
 | [no-hooks](docs/rules/no-hooks.md)                                           | Disallow setup and teardown hooks                                   |                  |              |
 | [no-identical-title](docs/rules/no-identical-title.md)                       | Disallow identical titles                                           | ![recommended][] |              |
+| [no-if](docs/rules/no-if.md)                                                 | Disallow conditional logic                                          | ![deprecated][]  |              |
```
G-Rath, line 218: @SimenB I guess we forgot to remove this in the new major 😅

docs/rules/consistent-test-it.md
```diff
@@ -1,14 +1,26 @@
 # Have control over `test` and `it` usages (`consistent-test-it`)
 
+<!-- RULE_NOTICE -- Generated by `yarn tools:regenerate-docs` -->
+
+<!-- prettier-ignore -->
```
G-Rath, line 5: It should be safe to apply prettier to these and I think it's better to be consistent as much as possible

docs/rules/consistent-test-it.md
```diff
@@ -1,14 +1,26 @@
...
+💼 This rule is enabled in the following [configs](https://github.com/jest-community/eslint-plugin-jest#shareable-configurations): `all`.
+
+<!-- prettier-ignore -->
+🔧 This rule is automatically fixable using the `--fix` [option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) on the command line.
+
+<!-- /RULE_NOTICE -->
```
G-Rath, line 11: I'd really love it if we could somehow not need these markers all together, but I can't think of a good way to match that - still, since we're wanting these to always be at top and everything until this line is generated I think we can just use a "end rule notices" marker (I'd like that name too, because its consistent with the format of the markers in the readme)

I think it would also make sense to have the tool add this if it can't find the marker, which would make it a bit smoother if you're making a new rule.

docs/rules/expect-expect.md
```diff
@@ -1,8 +1,15 @@
...
+
+<!-- prettier-ignore -->
+💼 This rule is enabled in the following [configs](https://github.com/jest-community/eslint-plugin-jest#shareable-configurations): `all`.
+
+<!-- /RULE_NOTICE -->
+
 Ensure that there is at least one `expect` call made in a test.
 
-## Rule details
+## Rule Details
```
G-Rath, line 12: I think we should actually be ensuring `details` is lowercase because that's more consistent with other headers (and technically more grammatically correct)

docs/rules/no-done-callback.md
```diff
@@ -1,5 +1,15 @@
 # Avoid using a callback in asynchronous tests and hooks (`no-done-callback`)
 
+<!-- RULE_NOTICE -- Generated by `yarn tools:regenerate-docs` -->
+
+<!-- prettier-ignore -->
+💼 This rule is enabled in the following [configs](https://github.com/jest-community/eslint-plugin-jest#shareable-configurations): `all`, `recommended`.
+
+<!-- prettier-ignore -->
+💡 This rule is manually fixable by editor [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
```
G-Rath, line 9: nit: what do you think of "This rule provides suggestions that can be applied manually"? That feels a bit more grammatically natural to me

docs/rules/no-if.md
```diff
@@ -1,6 +1,10 @@
 # Disallow conditional logic (`no-if`)
 
-## Deprecated
+<!-- RULE_NOTICE -- Generated by `yarn tools:regenerate-docs` -->
+
+❌ This rule is deprecated.
```
G-Rath, line 5: nit: doesn't have to be done in this PR, but a good followup could be adding "in favor of ..." if there's a suggested replacement

README.md
bmish: Yeah, luckily with this change it will be clearly visible that it still exists but is deprecated.

src/__tests__/rules.test.ts
```diff
@@ -76,4 +82,54 @@ describe('rules', () => {
       expect(() => require(`../rules/${ruleName}`)).not.toThrow();
     });
   });
+
+  describe('rule documentation files have the correct content', () => {
+    it.each(ruleNames)('%s', ruleName => {
```
G-Rath, line 87: I don't really think we need this test because we're able to re-run the docs tool in ci and check if any files have changed, whereas we'd have to update this test if the output changes and its relatively complex.

The only thing we won't catch from the tool is if we've got or not got a header for options, but I'm not super fussed about that because that's a pretty obvious one and the more important thing is ensuring all the options are documented which is a lot harder to automate. (it'd also be pretty easy to have the docs tool check & fail on anyway).

src/__tests__/rules.test.ts
```diff
@@ -76,4 +82,54 @@ describe('rules', () => {
...
+      if (
+        (Array.isArray(rule.meta.schema) && rule.meta.schema.length > 0) ||
+        (typeof rule.meta.schema === 'object' &&
+          Object.keys(rule.meta.schema).length > 0)
+      ) {
+        // Should have an options section header:
+        expect(documentContents).toContain('## Options');
+      } else {
+        // Should NOT have any options section header:
+        expect(documentContents.includes('## Options')).toStrictEqual(false);
```
G-Rath, line 130: it doesn't matter since we'll be axing this test, but fwiw you should be able to use `not.toContain` here

tools/regenerate-docs.ts
```diff
@@ -126,8 +160,11 @@ details.forEach(({ name, description }) => {
 
   const contents = fs.readFileSync(pathToDoc).toString().split('\n');
 
+  // Replace the title.
   contents[0] = `# ${description} (\`${name}\`)`;
 
+  updateRuleNotices(contents, name);
```
G-Rath, line 166: Relating to my comment above since we're always working with the very start of the file, I think we should replace this with `removeHeader` + `generateHeader` functions, and then the `generateHeader` function can be simplified by just being an array:

```
const generateHeader = () => {
  const header = [`# ${description} (\`${name}\`)`];

  if (/* rule is in a config */) {
    header.push('', '$ this rule is in configs ...');
  }

  // etc

  return header.join('\n');
}
```

G-Rath (changes requested):
Looks like a good start - I've put some comments around on how I'd like the output to change; I'll hold off reviewing the `tools/` stuff in detail until we've gotten the actual output sorted because that'll impact the actual implementation :)

README.md
G-Rath: I guess - alternatively you could say now its clearly visible that it exists at all even though we don't want people to be using it 🤷

G-Rath:
> I'm not sure an ESLint RFC would be appropriate as the ESLint team isn't really in the business of enforcing a documentation standard for plugins

No but they have rules themselves that have documentation, and this is also exactly why @eslint-community has been created. I don't think it needs to be a strictly large or full-blown RFC but I do like the idea of something like this going through the RFC process in terms of visibility and discussion as I think it could help unify the community and especially help people that are looking to create their own plugins (I think especially for non-native English speakers, having a template like that would make it a lot easier).

cc @nzakas cause you're probably the best person to give a steer.

SimenB:
This is awesome @bmish, thanks for taking the time! Is the plan to upstream this into some shared eslint tool?

README.md
bmish: My personal preference is to continue documenting deprecated features while clearly indicating they are deprecated. People may still be using the deprecated features and need to find their documentation, and omitting any link to them can cause confusion.

docs/rules/no-done-callback.md
bmish: That's fine, updated.

docs/rules/expect-expect.md
bmish: Sounds good, updated.

docs/rules/consistent-test-it.md
bmish: I was able to remove the prettier ignore comments.

docs/rules/consistent-test-it.md
bmish: I switched to using an "end rule header" marker only. And the tool will update the header or create it if it doesn't exist.

src/__tests__/rules.test.ts
bmish: I removed the tests for the content that is automatically generated.

I do want to continue to advocate for testing that:
* incorrect notices are not present
* section title consistency like "Rule details"
* "Options" section is included only for rules with options

Not only did these tests catch many issues with the jest rule docs, but they also save a lot of time during PR reviews, since many contributors will often just copy and paste an existing rule doc, and then PR reviewers will have to remind them to remove all the outdated content (and outdated content/notices can easily slip through). These tests have saved me a lot of time in my other plugins.

I also have some tests in other plugins that ensure that each configuration option is mentioned in the rule doc like [this example](https://github.com/ember-cli/eslint-plugin-ember/blob/cb145d8f1b37cfa940f27ce57b0ffe45dc85ab4b/tests/rule-setup.js#L133). That would be a good follow-up to this PR, so I do think we can and should have at least basic tests that rule docs document their rule options.

I would also argue that having a few actual jest tests for rule docs is preferable to having the rule doc generator fail. Jest tests/assertions are standardized, easy/convenient to write, and will provide more detailed and consistent failure output than some custom failure in the rule doc generator.

tools/regenerate-docs.ts
bmish: I went ahead and refactored to something like this -- `generateRuleHeaderLines()` + `replaceOrCreateHeader()`

bmish:
> > I'm not sure an ESLint RFC would be appropriate as the ESLint team isn't really in the business of enforcing a documentation standard for plugins
> 
> No but they have rules themselves that have documentation, and this is also exactly why @eslint-community has been created. I don't think it needs to be a strictly large or full-blown RFC but I do like the idea of something like this going through the RFC process in terms of visibility and discussion as I think it could help unify the community and especially help people that are looking to create their own plugins (I think especially for non-native English speakers, having a template like that would make it a lot easier).
 
> This is awesome @bmish, thanks for taking the time! Is the plan to upstream this into some shared eslint tool?

Here's a summary of my plans to standardize ESLint plugin documentation:

1. Submitting PRs like this one to ensure the top ESLint plugins conform to basic documentation standards (as mentioned in the PR description, I've already completed this for many of the top plugins). The documentation of ESLint and its top plugins will naturally serve as examples for other plugin authors to follow. 
2. Updating the documentation templates in [generator-eslint](https://github.com/eslint/generator-eslint) which can be used for generating a new plugin or rule.
3. Exploring creating a lint rule(s) in [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) for enforcing documentation consistency (title, notices, etc), as most ESLint plugins already use this plugin and would thus easily pull in the new checks. We could also use a lint rule autofixer to generate the auto-generated parts of the rule doc (title, notices).
4. I'm open to working on some ESLint community recommendations around this, perhaps to point people to the generator-eslint templates or examples of high-quality / standardized documentation.

src/__tests__/rules.test.ts
G-Rath: For now I disagree: most of what you've put is already covered [by us running the docs tool in CI](https://github.com/jest-community/eslint-plugin-jest/blob/main/.github/workflows/nodejs.yml#L103-L121), with the resulting output telling people what they need to do to resolve the problem (run the tool).

My dislike is with how brittle this still feels + you needing to export stuff from `tools/`, vs our current CI setup.

bmish: I agree `tools/` isn't the best place to export stuff from so I can definitely move `rule-notices.ts` to `lib/` instead if you prefer.

But I'm confused by "most of what you've put is already covered [by us running the docs tool in CI](https://github.com/jest-community/eslint-plugin-jest/blob/main/.github/workflows/nodejs.yml#L103-L121)" as I've explicitly removed the tests for auto-generated content and only left in place tests that basic details are included or not included in the rule doc (which cannot be resolved by running the tool).

tools/regenerate-docs.ts
```diff
@@ -89,6 +99,36 @@ const updateRulesList = (
...
+ */
+const generateRuleHeaderLines = (description, name) => {
+  const lines = [`# ${description} (\`${name}\`)`];
+
+  lines.push(...getRuleNoticeLines(name));
+
+  lines.push(END_RULE_HEADER_MARKER);
+
+  return lines;
+};
```
G-Rath, line 116: We might as well just return an array here:

```suggestion
const generateRuleHeaderLines = (description, name): string[] => [
  `# ${description} (\`${name}\`)`,
  ...getRuleNoticeLines(name),
  END_RULE_HEADER_MARKER,
];
```

src/__tests__/rules.test.ts
G-Rath: I would expect the tool to be able to handle the notices completely, since if we can detect them we should be able to remove them as well.

As the other stuff it feels very low gain because the important part is actually _where_ in the document they are - i.e. you can make the tests pass by just adding those headers at the start or end. Personally what I'd rather see from this is actually further evolving the docs tool (e.g. have it output warnings about what you're testing for here) because that's more portable - we actually should already be almost at a point where you could have it as its own package (we'd just need to sort out loading the rules).

But that's out of scope for this I think - the key thing for me here is that we're still checking about the notices when we should be able to handle that in the tool.

tools/regenerate-docs.ts
```diff
@@ -5,7 +5,12 @@ import * as path from 'path';
 import { TSESLint } from '@typescript-eslint/utils';
 import prettier, { Options } from 'prettier';
 import { prettier as prettierRC } from '../package.json';
-import config from '../src/index';
+import plugin from '../src/index';
+import { getRuleNoticeLines } from './rule-notices';
+
+// Marker so that rule doc header (title/notices) can be automatically updated.
+export const END_RULE_HEADER_MARKER =
+  '<!-- end rule header - generated by `yarn tools:regenerate-docs` -->';
```
G-Rath, line 13: I'm in favor of omitting the `generated by..` because I think its easy to locate the tool by searching for the header and the command might change (i.e. if we extracted it into a cli package, then what do we have here?)

tools/rule-notices.ts
```diff
@@ -0,0 +1,109 @@
...
+    '💼 This rule is enabled in the following [configs](https://github.com/jest-community/eslint-plugin-jest#shareable-configurations):',
+  [MESSAGE_TYPE.DEPRECATED]: '❌ This rule is deprecated.',
+  [MESSAGE_TYPE.FIXABLE]:
+    '🔧 This rule is automatically fixable using the `--fix` [option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) on the command line.',
+  [MESSAGE_TYPE.HAS_SUGGESTIONS]:
+    '💡 This rule provides [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) that can be applied manually.',
+};
+
+type RuleName = keyof typeof plugin.rules;
+type Rule = typeof plugin.rules[RuleName];
```
G-Rath, line 21: I'd rather not have the `@ts-ignore` below instead of this as I don't think there's a lot of value vs using `string` & co; it might be possible to make it work but I'd have to play around with the types a bit (which I don't have the time for right now) so if you can't make it work lets just go with `string`.

tools/regenerate-docs.ts
```diff
@@ -89,6 +99,36 @@ const updateRulesList = (
...
+};
+
+/**
+ * Replace the header of a doc up to and including the specified marker.
+ * Insert at beginning if header doesn't exist.
+ * @param lines - lines of doc
+ * @param newHeaderLines - lines of new header including marker
+ * @param marker - marker to indicate end of header
+ */
+const replaceOrCreateHeader = (lines, newHeaderLines, marker) => {
```
G-Rath, line 125: It looks like maybe our typechecking is not running properly on these files - I'll look into that later, but you need to have explicit types on these params as currently they're implicitly `any`.

(You can make typechecking work by adding `tools/**/*` to the `include` array in `tsconfig.json`, but I need to check if that will impact the compiled output before actually committing that)

tools/rule-notices.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Get the lines for the notice section at the top of a rule doc.
+ */
+export function getRuleNoticeLines(ruleName: RuleName) {
+  const lines: string[] = [];
+
+  const { expectedNotices } = getNoticesForRule(plugin.rules[ruleName]);
+
+  expectedNotices.forEach(expectedNotice => {
```
G-Rath, line 92: nit: I'd prefer using `for ... of`

tools/rule-notices.ts
```diff
@@ -0,0 +1,109 @@
...
+ * Convert list of configs to formatted string list.
+ */
+function configNamesToList(configNames: string[]) {
+  return `\`${configNames.join('`, `')}\``;
+}
+
+/**
+ * Determine which notices should and should not be included at the top of a rule doc.
+ */
+export function getNoticesForRule(rule: Rule) {
```
G-Rath, line 56: I'm hoping that by moving the handling of notices entirely into the tool we can re-implement this to leverage native data structures a bit more rather than need all these elses i.e. we could have a map of `{ message: expected }`:

```
{
  [MESSAGE_TYPE.DEPRECATED]: rule.meta.deprecated,
}
```

which you could iterate over with something like:

```
for (const [notice, expected] of Object.entries(notices)) {
  if (expected) {
    header.push(notice);

    continue;
  }

  // or maybe linesToRemove.push(notice), and then do that in the outer function
  content.replace(notice, '');
}
```

(this is very psudo-y; let me know if that doesn't make sense 😅)

tools/regenerate-docs.ts
bmish: Good point, fixed.

tools/rule-notices.ts
bmish: Fixed. I recommend we enable eslint-plugin-unicorn recommended rules including [unicorn/no-array-for-each](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md) to catch stuff like this.

tools/regenerate-docs.ts
bmish: Good catch, added types and updated tsconfig.json. We have `"noEmit": true` so I don't think this will cause any output to be produced?

tools/rule-notices.ts
G-Rath: I'm not committed to it enough yet to bring in a whole new plugin especially just for one rule (+ I don't think its worth the energy bikeshedding over all the rules we could have, given how stable this plugin is) - its more something I've been evaluating for the last couple of months to see how I feel about it :)

tools/regenerate-docs.ts
G-Rath: Thanks, but I'd really like to avoid changing the `tsconfig.json` in anyway in this PR just to be on the safe side - its also an indicator that I should do a look over the whole codebase for a bit of a tidy up

tools/rule-notices.ts
bmish: Well this is just one example. Of course I think there are many valuable rules in eslint-plugin-unicorn. Something to consider separately.

tools/regenerate-docs.ts
bmish: Okay removed change to `tsconfig.json`.

tools/rule-notices.ts
bmish: This `RuleName` type I have already evaluates to a `string`. I went ahead and removed this type since it's not really necessary, but this does not solve the `@ts-ignore` issue below.

```
Element implicitly has an 'any' type because expression of type '`jest/${string}`' can't be used to index type 'Record<string, RuleLevel> | { 'jest/no-alias-methods': string; 'jest/prefer-to-be': string; 'jest/prefer-to-contain': string; 'jest/prefer-to-have-length': string; }'.ts(7053)
```

This could be solved if we had a static type containing all the possible rule names, but I don't think we have that, and having to define a type containing all the rule names would be a maintenance burden, so it seems best to just use `@ts-ignore` for this situation.

tools/regenerate-docs.ts
bmish: Removed this part of the marker.

tools/rule-notices.ts
bmish: I switched over to having `getNoticesForRule()` return an object. I agree this is much nicer.

src/__tests__/rules.test.ts
bmish: Okay, I went ahead and moved these tests (checking for `## Rule details` and `## Options`) into the generator. If these sections are incorrectly present, we will output the issue and the script will exit with failure.

docs/rules/no-if.md
bmish: Definitely like this idea so I went ahead and updated the notice to link to the replacement rule(s).

docs/rules/consistent-test-it.md
```diff
@@ -1,14 +1,26 @@
 # Have control over `test` and `it` usages (`consistent-test-it`)
 
+💼 This rule is enabled in the following
+[configs](https://github.com/jest-community/eslint-plugin-jest#shareable-configurations):
```
SimenB, line 4: can we link to the `README.md` file? Otherwise the link is broken on mobile

README.md
SimenB, line 218: nice, listing out all rules even if deprecated makes sense to me 👍

docs/rules/consistent-test-it.md
SimenB, line 4: also, should it say `config` instead of `configs` when only in a single one?

package.json
```diff
@@ -81,6 +81,9 @@
           "<rootDir>/lib/.*",
           "<rootDir>/src/rules/__tests__/fixtures/*",
           "<rootDir>/src/rules/__tests__/test-utils.ts"
+        ],
+        "coveragePathIgnorePatterns": [
+          "<rootDir>/tools/"
```
SimenB, line 86: what isn't covered?

G-Rath: It's the tools - becuase they're not imported by anything that is involved in the tests, they're not expected to be covered. When @bmish introduced tests that imported from there, they got included in coverage.

Now that those tests have been axed again, this can go too

docs/rules/consistent-test-it.md
bmish: Good point, updated.

package.json
bmish: Removed.

docs/rules/consistent-test-it.md
bmish: I could go either way on this. My first thought was that it was grammatically correct to use `configs` regardless, since this sentence is separated from the serial list with a colon, and it's clear that the list of configs is variable. If I'm wrong or you still prefer the other way, let me know.

SimenB: I'm not a native speaker, so happy to leave it 👍

SimenB (approved):

G-Rath:
@bmish I've just landed #1239 which updates the `tsconfig.json` to check tools - could you rebase to include that, as `typecheck` is failing now.

tools/rule-notices.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+function getConfigsForRule(ruleName: string) {
+  const { configs } = plugin;
+  const configNames: Array<keyof typeof configs> = [];
+  let configName: keyof typeof configs;
+
+  for (configName in configs) {
+    const config = configs[configName];
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- we don't have a static type for rule names
+    // @ts-ignore
```
G-Rath, line 33: I'm hoping to get this resolved via #1241 - in the meantime, disabling this rule is not the appropriate thing to do; instead you should use `@ts-expect-error` with a reason:

```suggestion
    // @ts-expect-error -- we don't have a static type for rule names
```

G-Rath (changes requested):
This is looking awesome! If it's ok with you, I'm keen to toy with this a bit once its landed to see if it can be expanded into a self-standing tool.

We're just got a few type related things to resolve before landing this :)

tools/rule-notices.ts
SimenB: Merged in `main`, is this still right?

G-Rath: na now it can go completely

SimenB, line 33:
```suggestion
```

SimenB:
Thanks @bmish, this is awesome!

bmish:
@G-Rath @SimenB thanks a lot for helping to get this in! As part of my goal to standardize ESLint plugin documentation, I agree we should work to share this tool, and I'd be interested to collaborate on that. 

If we went with the CLI command here, it could possibly make sense for it to live inside [generator-eslint](https://github.com/eslint/generator-eslint), although that tool currently only has Yeoman generators in it.

I also plan to experiment with whether I could create a lint rule in [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) for enforcing documentation consistency (title, notices, options, etc), as most ESLint plugins already use this plugin and would thus easily pull in the new checks. We could also use a lint rule autofixer to generate the auto-generated parts of the rule doc (title, notices).

docs/rules/consistent-test-it.md
ljharb: The link works fine for me on mobile, as-is.

PR #1230, opened 2022-09-04
docs: update removal details in `no-deprecated-functions`

mrazauskas:
Some removal details could be update in `no-deprecated-functions` rule details.

Methods which were scheduled to be removed in Jest 27 are already gone. See #9853

And `jest.genMockFromModule` is now scheduled to be removed in Jest 30: https://github.com/facebook/jest/blob/0cfc2ad2991373324306c8daf962c41d0f37793c/packages/jest-environment/src/index.ts#L132-L142


SimenB (approved):
thanks!

PR #1239, opened 2022-09-09
chore: make sure that tools are typechecked

G-Rath:
Found during #1226

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,5 +11,5 @@
     "resolveJsonModule": true,
     "forceConsistentCasingInFileNames": true
   },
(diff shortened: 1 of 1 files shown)


G-Rath:
Really we should also have `*.ts` but that'd mean having `danger` as a dev dependency - @SimenB are you ok with that?

SimenB:
Yeah, no real reason not tom

G-Rath:
Sweet - I'll merge this as is because its important for #1226, and then look into adding `*.ts` (as it looks like we also need to do something for `eslint-remote-tester`

PR #1242, opened 2022-09-10
chore: use `@tsconfig/node14` and stricter TS config

SimenB:
#1226 fails CI due to "missing" `flatMap` - instead of manually defining `lib` etc., extending `@tsconfig/node14` seems cleaner

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,15 +1,25 @@
 {
+  "extends": "@tsconfig/node14/tsconfig.json",
   "compilerOptions": {
-    "target": "es2015",
-    "module": "commonjs",
-    "moduleResolution": "node",
-    "lib": ["es2018"],
     "noEmit": true,
+    "importsNotUsedAsValues": "error",
+    "stripInternal": true,
+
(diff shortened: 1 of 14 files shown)


tsconfig.json
```diff
@@ -1,15 +1,25 @@
 {
+  "extends": "@tsconfig/node14/tsconfig.json",
   "compilerOptions": {
-    "target": "es2015",
-    "module": "commonjs",
     "moduleResolution": "node",
-    "lib": ["es2018"],
     "noEmit": true,
+    "importsNotUsedAsValues": "error",
```
G-Rath, line 6: Curious why this? I can live with it, but my understanding is that its "maybe" needed in some rare cases (e.g. using something that isn't smart enough to automatically figure it out like tsc and Babel can), and it means extra noise in diffs when you need to switch between imports.

I'm pretty sure it doesn't help us as I've looked at our compiled outside in the past and feel this is the sort of thing I'd have done already if I found it was adding unneeded imports

PR #1263, opened 2022-10-10
docs: use eslint-doc-generator for rule documentation

bmish:
I built this CLI tool [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) for automating the generation of the README rules list and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of our custom rules through better documentation and streamline the process of adding new rules).

This is a follow-up to the work I did to add the original notices in #1226 (and thanks @G-Rath and @SimenB for helping to inspire it). eslint-doc-generator is significantly more robust compared to the previous documentation generator script/tests which I have now removed. It has much more functionality (for example, the rules list legend is auto-generated, and the rules list will show additional columns of information when applicable) and 100% test coverage.

There are some discrepancies between the old docs and the new docs. Most of these are intentional and likely improvements. If you don't like any of the changes, let me know and I can tweak eslint-doc-generator or add a config option to support the desired behavior.

Some differences include:
* The `all` config is hidden everywhere as a result of the `--ignore-config all` option
* Removed the `### Default configuration` section header from some rule docs which was confusing because it was used for rules with no config/options (and causes the options section check to fail)


G-Rath:
This sounds really cool! I'll have a proper look when I've got some time, but two initial thoughts:

  * would it be possible to support having a different table for the type based rules? the reason I did that originally is because it seemed distracting to have a whole other column for something that only applies for one rule, and that we're probably not going to see any more type-based rules anytime soon (I expect that's probably true for a lot of plugins outside of `@typescript-eslint`)
  * it looks like the script entry in `package.json` is pretty large - maybe it would make sense to have a lightweight config json file?

bmish:
@G-Rath yes, I am planning to support both of those, see these issues:

* https://github.com/bmish/eslint-doc-generator/issues/22
* https://github.com/bmish/eslint-doc-generator/issues/95

SimenB (approved):
yay, this is awesome! stoked to see this land when Gareth is happy 🙂

SimenB:
Hmm, weird CI failure needs to be investigated

bmish:
I implemented a feature to separate the type-checking rules list with `--split-by meta.docs.requiresTypeChecking` and then was able to hide the type-checking column from the lists using the `--rule-list-columns` option.

Still need to investigate the CI failure.

docs/rules/expect-expect.md
```diff
@@ -1,10 +1,9 @@
 # Enforce assertion to be made in a test body (`expect-expect`)
 
-💼 This rule is enabled in the following
-[configs](https://github.com/jest-community/eslint-plugin-jest/blob/main/README.md#shareable-configurations):
-`all`.
+✅ This rule will _warn_ in the `recommended`
+[config](https://github.com/jest-community/eslint-plugin-jest/blob/main/README.md#shareable-configurations).
```
bmish, line 4: This comes from a new feature to indicate when a rule is set to *warn* in a config.

bmish:
The CI failure has been fixed by ensuring we build before running the generator.

This PR is ready for review.

Note that I plan to implement this feature eventually but would prefer not to block the PR on it:
* https://github.com/bmish/eslint-doc-generator/issues/95

bmish:
Just a heads up that this PR is still ready. I'm just updating it whenever I published a new version of the tool.

G-Rath:
sweet - I'll try and have a look over the weekend; thanks again for your work on this!

package.json
```diff
@@ -29,7 +29,7 @@
     "prettier:write": "prettier --write 'docs/**/*.md' README.md '.github/**' package.json tsconfig.json src/globals.json .yarnrc.yml",
     "postpublish": "pinst --enable",
     "test": "jest",
-    "tools:regenerate-docs": "ts-node -T tools/regenerate-docs",
+    "tools:regenerate-docs": "yarn prepack && eslint-doc-generator --ignore-config all --rule-doc-title-format desc-parens-name --rule-doc-section-include \"Rule details\" --rule-list-columns name,description,configsError,configsWarn,configsOff,fixable,hasSuggestions,deprecated --split-by meta.docs.requiresTypeChecking --url-configs \"https://github.com/jest-community/eslint-plugin-jest/blob/main/README.md#shareable-configurations\" && yarn prettier:write",
```
G-Rath, line 32: Can I vote again for having a config file? 😅

G-Rath (approved):
Great stuff! I have some mild feelings about if we really need so many columns in the table (i.e. the "Warn" and "fix vs suggest" specifically), but other maintainers don't seem mind and hey this is exactly what I was hoping for in the long run with my `regenerate-docs` script 😂

package.json
bmish: Haha yeah a config file (https://github.com/bmish/eslint-doc-generator/issues/61) will be one of my top priorities now that I've successfully introduced this tool into ~10 top ESLint plugins (a month of work was needed to implement all the baseline functionality necessary for that).

bmish:
> I have some mild feelings about if we really need so many columns in the table (i.e. the "Warn" and "fix vs suggest" specifically)

The good news is that the columns are [configurable](https://github.com/bmish/eslint-doc-generator#configuration-options). I filed an issue to consider a consolidated column for fixable/suggestions: https://github.com/bmish/eslint-doc-generator/issues/207.

For the configuration columns, I've gone through multiple iterations of it and the current design of a separate column per configured rule severity is by far the most scalable and robust solution I've reached.

MichaelDeBoey (approved):

package.json
bmish: Implemented: https://github.com/bmish/eslint-doc-generator/releases/tag/v0.20.0

PR #1264, opened 2022-10-14
Fix typo: possessive it's should be its

sirreal:
Fix a typo in `valid-expect-in-promise` that uses [`it's` as a possessive instead of the correct `its`](https://www.merriam-webster.com/words-at-play/when-to-use-its-vs-its):

```
   17:5  error  This promise should either be returned or awaited to ensure the expects in it's chain are called  jest/valid-expect-in-promise
```

### DIFF
--- a/src/rules/valid-expect-in-promise.ts
+++ b/src/rules/valid-expect-in-promise.ts
@@ -347,7 +347,7 @@ export default createRule({
     },
     messages: {
       expectInFloatingPromise:
-        "This promise should either be returned or awaited to ensure the expects in it's chain are called",
+        'This promise should either be returned or awaited to ensure the expects in its chain are called',
     },
     type: 'suggestion',
     schema: [],


G-Rath (approved):

nchevsky:
Sad that this fix ended up with a changelog entry of _"valid-expect-in-promise: adjust **grammer** [sic] in rule message."_ 🥲

G-Rath:
ugh crap yeah sorry for some reason my dyslexia always messes that specific word for me without fail - I'll see if I can change it

G-Rath:
I've updated the GitHub release notes and opened https://github.com/jest-community/eslint-plugin-jest/pull/1265 updating the changelog.

I'd like to redo the commit but won't for obvious reasons.

sirreal:
No worries and thank you. I had several misspellings myself as I crafted this small PR 😅

PR #1275, opened 2022-11-04
perf: don't collect more info than needed when resolving jest functions

G-Rath:
Huge thanks to @SLKnutson for pointing this out - this is mainly a win for larger codebases, since we still have to loop until we find the information we need (or don't), but it is definitely faster

Before, completed in ~30 seconds:

```
Rule                            | Time (ms) | Relative
:-------------------------------|----------:|--------:
jest/consistent-test-it         |   919.565 |     5.3%
jest/prefer-expect-assertions   |   714.352 |     4.1%
jest/no-conditional-expect      |   626.779 |     3.6%
jest/no-standalone-expect       |   612.223 |     3.5%
jest/prefer-lowercase-title     |   608.857 |     3.5%
jest/no-identical-title         |   599.639 |     3.4%
jest/no-disabled-tests          |   587.099 |     3.4%
jest/prefer-each                |   586.834 |     3.4%
jest/prefer-snapshot-hint       |   585.044 |     3.4%
jest/require-top-level-describe |   579.505 |     3.3%
```

After, completed in ~15 seconds:

```
Rule                            | Time (ms) | Relative
:-------------------------------|----------:|--------:
jest/consistent-test-it         |   370.588 |     6.8%
jest/prefer-expect-assertions   |   248.168 |     4.6%
jest/expect-expect              |   232.857 |     4.3%
jest/no-standalone-expect       |   209.712 |     3.9%
jest/no-conditional-expect      |   193.040 |     3.6%
jest/no-disabled-tests          |   173.617 |     3.2%
jest/no-identical-title         |   167.683 |     3.1%
jest/prefer-snapshot-hint       |   161.561 |     3.0%
jest/require-top-level-describe |   159.859 |     3.0%
jest/prefer-each                |   159.230 |     2.9%
```

Resolves #1274

Issue #1274, opened 2022-11-02
Rules run a lot slower now


SLKnutson:
Hi,
I've noticed that recently this plugin has gotten a lot slower. I use many eslint rules, and still this plugin is over 50% of the runtime. It looks like it spends the majority of its time in `parseJestFnCall`.

Here is a sampling of the output from TIMING=all
```
jest/consistent-test-it                                    | 12047.834 |     5.5%
jest/no-disabled-tests                                     |  8272.639 |     3.8%
jest/no-duplicate-hooks                                    |  8047.621 |     3.7%
jest/no-identical-title                                    |  8024.471 |     3.7%
jest/require-top-level-describe                            |  7910.195 |     3.6%
jest/no-alias-methods                                      |  4233.073 |     1.9%
jest/no-export                                             |  4037.647 |     1.9%
jest/valid-expect                                          |  4012.385 |     1.8%
jest/valid-title                                           |  4009.165 |     1.8%
jest/prefer-hooks-on-top                                   |  3998.323 |     1.8%
jest/prefer-comparison-matcher                             |  3997.734 |     1.8%
jest/prefer-to-contain                                     |  3995.129 |     1.8%
jest/prefer-todo                                           |  3993.870 |     1.8%
jest/no-large-snapshots                                    |  3992.662 |     1.8%
jest/prefer-equality-matcher                               |  3988.052 |     1.8%
jest/prefer-strict-equal                                   |  3982.349 |     1.8%
jest/no-interpolation-in-snapshots                         |  3981.740 |     1.8%
jest/prefer-expect-resolves                                |  3981.178 |     1.8%
jest/no-focused-tests                                      |  3972.438 |     1.8%
jest/prefer-to-have-length                                 |  3968.205 |     1.8%
jest/require-to-throw-message                              |  3964.580 |     1.8%
jest/no-test-prefixes                                      |  3957.872 |     1.8%
jest/valid-describe-callback                               |  3945.417 |     1.8%
```

Thanks!

G-Rath:
Can you provide some more details about your setup, including the version of node, `eslint-plugin-jest`, OS, etc?

`parseJestFnCall` definitely has a higher performance cost as its doing a lot more work, but I don't think there's anything that can really be optimized and the rules should still be very fast compared to others (i.e. they're all only less than 4% each in your timings) - the main performance improvement we could make is to cache the results, which was landed in v28.8.3 via #1187.

[...]

PR #1276, opened 2022-11-04
refactor: inline `collectReferences` utility into `scopeHasLocalReference`

G-Rath:
Follow on to #1274 & #1275 - in theory this is faster but it has no noticeable impact on performance so I'm marking this as a refactor rather than `perf`.

The actual reason I'm doing it in the first place is because it makes the codebase a little smaller and we don't really need the two functions anymore.


G-Rath:
huh so apparently we didn't need the `unresolved` check at all - that's nice

SimenB (approved):
Big fan of 20 less lines of code

PR #1277, opened 2022-11-04
ci: adjust permissions so that comments can be put on prs and issues

G-Rath:
The hardening done in #1249 was incomplete - this should resolve that.


SimenB:
👍

PR #1281, opened 2022-11-08
ci: remove git credentials after checkout

Fdawgs:
This PR removes Git credentials/SSH keys after checkout as a security precaution by setting `persist-credentials` to false, they are not used after the initial checkout.

Due to git being used in the `docs` job in the nodejs workflow, it has not been removed there:
https://github.com/jest-community/eslint-plugin-jest/blob/ad04fcc97b5dcc676c5bd2152f84608c08e0c50a/.github/workflows/nodejs.yml#L125-L128

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,6 +11,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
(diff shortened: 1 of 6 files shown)


SimenB (approved):
sweet, thanks!

PR #1282, opened 2022-11-10
fix: ensure rule fixes produce valid code when function params and args have trailing commas

G-Rath:
The fix that no one asked for 😄 props to @berkinanik for bring this to my attention over in `eslint-plugin-jest-dom`.

Weirdly ESLint doesn't seem to think some of the result syntax before this is invalid (namely in `no-done-callback` which'd give `((), ) => {}`), but didn't feel like digging into that further right now.

### DIFF
--- a/src/rules/__tests__/no-done-callback.test.ts
+++ b/src/rules/__tests__/no-done-callback.test.ts
@@ -60,6 +60,24 @@ ruleTester.run('no-done-callback', rule, {
       ],
     },
     {
+      code: 'test("something", (done,) => {done();})',
+      errors: [
+        {
+          messageId: 'noDoneCallback',
+          line: 1,
+          column: 20,
+          suggestions: [
+            {
+              messageId: 'suggestWrappingInPromise',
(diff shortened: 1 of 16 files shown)


src/rules/__tests__/no-done-callback.test.ts
```diff
@@ -89,7 +107,7 @@ ruleTester.run('no-done-callback', rule, {
               messageId: 'suggestWrappingInPromise',
               data: { callback: 'done' },
               output:
-                'test("something", () => {return new Promise((done) => {done();})})',
+                'test("something", () => {return new Promise(done => {done();})})',
```
SimenB, line 110: it should probably preserve the parens if they were there originally?

G-Rath: Best practice for fixers are to avoid doing stylistic fixes since that should be enforced by another rule (which will be automatically fixed if possible), to help keep fixes as small as possible.

PR #1291, opened 2022-11-19
chore: switch to using config file to configure `eslint-doc-generator`

G-Rath:
I've created https://github.com/bmish/eslint-doc-generator/issues/260 about supporting TypeScript for config, but for now JS will do.

Currently we don't actually check JS files with TS, but will do that in another PR since it affects a few other files.

### DIFF
--- a/.eslint-doc-generatorrc.js
+++ b/.eslint-doc-generatorrc.js
@@ -1,0 +1,20 @@
+/** @type {import('eslint-doc-generator/dist/lib/options').GenerateOptions} */
+const config = {
+  ignoreConfig: ['all'],
+  ruleDocTitleFormat: 'desc-parens-name',
+  ruleDocSectionInclude: ['Rule details'],
+  ruleListColumns: [
+    'name',
+    'description',
+    'configsError',
+    'configsWarn',
+    'configsOff',
+    'fixable',
+    'hasSuggestions',
+    'deprecated',
+  ].join(),
+  splitBy: 'meta.docs.requiresTypeChecking',
+  urlConfigs: `https://github.com/jest-community/eslint-plugin-jest/blob/main/README.md#shareable-configurations`,
+};
+
+module.exports = config;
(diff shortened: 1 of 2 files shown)


bmish (approved):

.eslint-doc-generatorrc.js
```diff
@@ -0,0 +1,20 @@
+/** @type {import('eslint-doc-generator/dist/lib/options').GenerateOptions} */
```
bmish, line 1: Nice! I'll export this type in the future: https://github.com/bmish/eslint-doc-generator/issues/261

bmish: FYI the public type is now available:

* https://github.com/bmish/eslint-doc-generator/releases/tag/v0.26.0
* https://github.com/bmish/eslint-doc-generator#configuration-file
* https://github.com/bmish/eslint-doc-generator/pull/264

G-Rath: sweet - I'm going to leave it as-is for now but will address when https://github.com/bmish/eslint-doc-generator/issues/260 is done

bmish: Okay well you might want to switch to the public type sooner since:
* The type no longer exists at the location you're using it--would cause an issue with any upgrade
* I don't think I'll be able to work on the TypeScript config support any time soon

bmish: Looks like this is now broken for you after the update.

G-Rath: Nope nothing to do with this because

> Currently we don't actually check JS files with TS

🙂

bmish: Regardless of that, I find it valuable for the hover-over documentation / auto-complete too.

G-Rath: As do I, but that's primarily useful when you edit the config file so I'll either update the type then or when I setup type checking for JS files.

bmish: Sounds good!

PR #1314, opened 2022-12-24
feat(require-typed-module-mocks): add rule to use strict module mock factory types

### LINKED ISSUE #1313: [new rule] Require that TypeScript tests use a typed jest.mock factory
A new rule that requires type-checking. 

`jest.mock` and `jest.doMock` take an optional generic, representing the return type of the factory. For example,
```ts
// The optional type argument provides typings for the module factory
jest.mock<typeof import('../moduleName')>('../moduleName', () => {
  return jest.fn(() => 42);
});
```

Without specifying `typeof import('../moduleName')`, Jest's types doesn't know what the module looks like and will allow anything to be returned in the factory.

I'd like to have an ESLint rule to enforce that the generic parameter is used whenever a factory function is used. Jest auto-mocks (no factory) and virtual mocks (no corresponding actual module) don't need to have the generic parameter.

```ts
jest.mock('../moduleName', () => {
  return jest.fn(() => 42);
});
jest.mock('./module', () => ({
  ...jest.requireActual('./module'),
  foo: jest.fn()
}));
jest.mock('random-num', () => {
  return jest.fn(() => 42);
});
```

```ts
// Uses typeof import()
jest.mock<typeof import('../moduleName')>('../moduleName', () => {
  return jest.fn(() => 42);
});
jest.mock<typeof import('./module')>('./module', () => ({
  ...jest.requireActual('./module'),
  foo: jest.fn()
}));

// Uses custom type
jest.mock<() => number>('random-num', () => {
  return jest.fn(() => 42);
});
// No factory
jest.mock('random-num');
// Virtual mock
jest.mock(
  '../moduleName',
  () => {
    return jest.fn(() => 42)
  },
  {virtual: true},
);
```

**Auto-fixable:** Yes

NotWoods:
Closes #1313

While the rule itself doesn't need type-checking, the auto-fix does. It it possible to disable the auto-fix based on the parser & source file? If not, the rule could be switched to use editor suggestions.

### DIFF
--- a/docs/rules/no-untyped-mock-factory.md
+++ b/docs/rules/no-untyped-mock-factory.md
@@ -1,0 +1,72 @@
+# Disallow using `jest.mock()` factories without an explicit type parameter (`no-untyped-mock-factory`)
+
+🔧 This rule is automatically fixable by the
+[`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+By default, `jest.mock` and `jest.doMock` allow any type to be returned by a
+mock factory. A generic type parameter can be used to enforce that the factory
+returns an object with the same shape as the original module, or some other
+strict type. Requiring a type makes it easier to use TypeScript to catch changes
+needed in test mocks when the source module changes.
+
+> **Warning**
+>
+> This rule expects to be run on TypeScript files **only**. If you are using a
+> codebase that has a mix of JavaScript and TypeScript tests, you can use
+> [overrides](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#how-do-overrides-work)
+> to apply this rule to just your TypeScript test files.
+
+## Rule details
+
+This rule triggers a warning if `mock()` or `doMock()` is used without a generic
+type parameter or return type.
+
+The following patterns are considered errors:
+
+```ts
+jest.mock('../moduleName', () => {
+  return jest.fn(() => 42);
+});
+
+jest.mock('./module', () => ({
+  ...jest.requireActual('./module'),
+  foo: jest.fn(),
+}));
+
+jest.mock('random-num', () => {
+  return jest.fn(() => 42);
+});
+```
+
+The following patterns are not considered errors:
+
+```ts
+// Uses typeof import()
+jest.mock<typeof import('../moduleName')>('../moduleName', () => {
+  return jest.fn(() => 42);
+});
+
+jest.mock<typeof import('./module')>('./module', () => ({
+  ...jest.requireActual('./module'),
+  foo: jest.fn(),
+}));
+
+// Uses custom type
+jest.mock<() => number>('random-num', () => {
+  return jest.fn(() => 42);
+});
+
+// No factory
+jest.mock('random-num');
+
+// Virtual mock
+jest.mock(
+  '../moduleName',
+  () => {
+    return jest.fn(() => 42);
+  },
+  { virtual: true },
+);
+```
--- a/src/rules/__tests__/no-untyped-mock-factory.test.ts
+++ b/src/rules/__tests__/no-untyped-mock-factory.test.ts
@@ -1,0 +1,155 @@
+import { TSESLint } from '@typescript-eslint/utils';
+import dedent from 'dedent';
+import rule from '../no-untyped-mock-factory';
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: require.resolve('@typescript-eslint/parser'),
+});
+
+ruleTester.run('no-untyped-mock-factory', rule, {
+  valid: [
+    "jest.mock('random-number');",
+    dedent`
+      jest.mock<typeof import('../moduleName')>('../moduleName', () => {
+        return jest.fn(() => 42);
+      });
+    `,
+    dedent`
+      jest.mock<typeof import('./module')>('./module', () => ({
+        ...jest.requireActual('./module'),
+        foo: jest.fn()
+      }));
+    `,
+    dedent`
+      jest.mock<typeof import('foo')>('bar', () => ({
+        ...jest.requireActual('bar'),
+        foo: jest.fn()
+      }));
+    `,
+    dedent`
+      jest.doMock('./module', (): typeof import('./module') => ({
+        ...jest.requireActual('./module'),
+        foo: jest.fn()
+      }));
+    `,
+    dedent`
+      jest.mock('../moduleName', function (): typeof import('../moduleName') {
+        return jest.fn(() => 42);
+      });
+    `,
+    dedent`
+      jest.mock<() => number>('random-num', () => {
+        return jest.fn(() => 42);
+      });
+    `,
+    dedent`
+      jest['doMock']<() => number>('random-num', () => {
+        return jest.fn(() => 42);
+      });
+    `,
+    dedent`
+      jest.mock<any>('random-num', () => {
+        return jest.fn(() => 42);
+      });
+    `,
+    dedent`
+      jest.mock(
+        '../moduleName',
+        () => {
+          return jest.fn(() => 42)
+        },
+        {virtual: true},
+      );
+    `,
+    dedent`
+      jest.mock('../moduleName', function (): (() => number) {
+        return jest.fn(() => 42);
+      });
+    `,
+    // Should not match
+    dedent`
+      mockito<() => number>('foo', () => {
+        return jest.fn(() => 42);
+      });
+    `,
+  ],
+  invalid: [
+    {
+      code: dedent`
+        jest.mock('../moduleName', () => {
+          return jest.fn(() => 42);
+        });
+      `,
+      output: dedent`
+        jest.mock<typeof import('../moduleName')>('../moduleName', () => {
+          return jest.fn(() => 42);
+        });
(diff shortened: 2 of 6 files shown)


G-Rath (commented):
Thanks, I'll do a proper review when I'm back at my laptop, but giving it a quick look over I've already got two main comments:
1. You don't need type checking becuase you're not using the TypeScript service
2. I feel like ideally we should have "factory" in the rule name, and maybe "generic" too, since this rule is pretty specific to those things - what do you think about `no-untyped-mock-factory` or `require-generic-type-with-mock-factory`?

G-Rath:
Also you need to use `parseJestFn` to check if a function is a call to the `jest` global/import - have a look at how some of the other rules are doing this, or otherwise I can give you some snippets when I'm back at my laptop :)

G-Rath (changes requested):
Looks like a good start, but need to address my comments so far + ensure CI is passing; let me know if you need any help :)

src/rules/__tests__/no-untyped-mock-factory.test.ts
```diff
@@ -0,0 +1,137 @@
...
+const rootPath = getFixturesRootDir();
+
+const ruleTester = new ESLintUtils.RuleTester({
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    sourceType: 'module',
+    tsconfigRootDir: rootPath,
+    project: './tsconfig.json',
+  },
+});
```
G-Rath, line 19: we shouldn't need this:

```suggestion
const ruleTester = new ESLintUtils.RuleTester({
  parser: '@typescript-eslint/parser',
});
```

G-Rath, line 19: We shouldn't be configuring a TypeScript project since that's not needed by this rule (+ let's switch to using `TSESLint` because that's what we use in our other tests):

```suggestion
import { TSESLint } from '@typescript-eslint/utils';
import dedent from 'dedent';
import rule from '../no-untyped-mock-factory';

const ruleTester = new TSESLint.RuleTester({
  parser: require.resolve('@typescript-eslint/parser'),
});
```

src/rules/__tests__/no-untyped-mock-factory.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+ruleTester.run('no-untyped-mock-factory', rule, {
+  valid: [
```
G-Rath, line 22: Could you add a test for `jest.mock<typeof import('foo')>('bar', ...)`, to showcase that we don't check the actual import path?

src/rules/no-untyped-mock-factory.ts
```diff
@@ -0,0 +1,95 @@
...
+const hasReturnType = (
+  node:
+    | TSESTree.FunctionExpression
+    | TSESTree.ArrowFunctionExpression
+    | TSESTree.Node,
+): boolean => {
+  if (
+    node.type === AST_NODE_TYPES.FunctionExpression ||
+    node.type === AST_NODE_TYPES.ArrowFunctionExpression
+  ) {
```
G-Rath, line 13: Let's use our `isFunction` utility here (tbh I'd probably inline this whole function since we only do this check in one place, but up to you)

src/rules/no-untyped-mock-factory.ts
```diff
@@ -0,0 +1,95 @@
...
+        if (
+          callee.type === AST_NODE_TYPES.MemberExpression &&
+          node.arguments.length === 2
+        ) {
+          const { property } = callee;
+
+          if (
+            isTypeOfJestFnCall(node, context, ['jest']) &&
+            property.type === AST_NODE_TYPES.Identifier &&
+            ['mock', 'doMock'].includes(property.name)
```
G-Rath, line 66: While we don't technically require all of our rules to support square brackets, you should be able to do that here by using `isSupportedAccessor` and `getAccessorValue` utilities (don't forget to add a couple of tests!)

Something like this is what you want:

```suggestion
            isSupportedAccessor(property) &&
            ['mock', 'doMock'].includes(getAccessorValue(property))
```

src/rules/no-untyped-mock-factory.ts
```diff
@@ -0,0 +1,95 @@
...
+          ) {
+            const hasTypeParameter =
+              typeParameters !== undefined && typeParameters.params.length > 0;
+
+            if (!hasTypeParameter && !hasReturnType(node.arguments[1])) {
```
G-Rath, line 71: Let's do an early return here, since it'll remove a level of indention for the rest of the rule + you'll not have to invert the checks (which generally is easier to reason about)

src/rules/no-untyped-mock-factory.ts
```diff
@@ -0,0 +1,95 @@
...
+  },
+  create(context) {
+    return {
+      CallExpression(node: TSESTree.CallExpression): void {
+        const { callee, typeParameters } = node;
+
+        if (
+          callee.type === AST_NODE_TYPES.MemberExpression &&
+          node.arguments.length === 2
+        ) {
```
G-Rath, line 60: nit: I typically prefer early returns to avoid extra levels of indentation, but I wouldn't say it's required

src/rules/no-untyped-mock-factory.ts
```diff
@@ -0,0 +1,95 @@
...
+        'Disallow using `jest.mock()` factories without an explicit type parameter',
+      recommended: false,
+    },
+    messages: {
+      addTypeParameterToModuleMock: DEFAULT_MESSAGE,
+    },
+    fixable: 'code',
+    schema: [],
+    type: 'suggestion',
+  },
```
G-Rath, line 51: Let's have TypeScript infer the generic types when possible and be consistent with how we structure our other rules:

```suggestion
export default createRule({
  name: __filename,
  meta: {
    docs: {
      category: 'Best Practices',
      description:
        'Disallow using `jest.mock()` factories without an explicit type parameter',
      recommended: false,
    },
    messages: {
      addTypeParameterToModuleMock:
        'Add a type parameter to the mock factory such as `typeof import({{ moduleName }})`',
    },
    fixable: 'code',
    schema: [],
    type: 'suggestion',
  },
  defaultOptions: [],
```

docs/rules/no-untyped-mock-factory.md
```diff
@@ -0,0 +1,57 @@
...
+
+<!-- end auto-generated rule header -->
+
+By default, `jest.mock` and `jest.doMock` allow any type to be returned by a
+mock factory. A generic type parameter can be used to enforce that the factory
+returns an object with the same shape as the original module, or some other
+strict type. Requiring a type makes it easier to use TypeScript to catch changes
+needed in test mocks when the source module changes.
+
+## Rule details
```
G-Rath, line 11: Let's add a warning about the fact that this rule should only be run on TS files, and that overrides can be used to help with that:

```suggestion
needed in test mocks when the source module changes.

> **Warning**
> 
> This rule expects to be run on TypeScript files **only**. If you are using a codebase that has a mix of JavaScript and TypeScript tests, you can use [overrides](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#how-do-overrides-work) to apply this rule to just your TypeScript test files.

## Rule details
```

(feel free to make changes to what I've written)

docs/rules/no-untyped-mock-factory.md
```diff
@@ -0,0 +1,57 @@
...
+
+This rule triggers a warning if `mock()` or `doMock()` is used without a generic
+type parameter or return type.
+
+The following patterns are considered errors:
+
+```ts
+jest.mock('../moduleName', () => {
```
G-Rath, line 19: nit: these feel a bit cramped to me - it'd be nice to space them out with some newlines between each example (same with the second codeblock below)

G-Rath (changes requested):
Looks good - have a few tweaks I'd like to see, but they should be pretty straightforward.

NotWoods:
Thank you for pointing me at some useful utilities! Updated the PR to address comments.

NotWoods:
Added an extra test to check for spots where the rule shouldn't match, in order to reach full code coverage.

src/rules/no-untyped-mock-factory.ts
```diff
@@ -0,0 +1,87 @@
...
+import {
+  createRule,
+  getAccessorValue,
+  isFunction,
+  isSupportedAccessor,
+  isTypeOfJestFnCall,
+} from './utils';
+
+const findModuleName = (
+  node: TSESTree.Literal | TSESTree.Node,
```
G-Rath, line 11: nit: iirc `TSESTree.Literal` is a subset of `TSESTree.Node`, so you shouldn't need it here

G-Rath (approved):
Looks good, just need to re-generate the docs :)

Blfrg:
Thank you, @NotWoods, for this PR.

To clarify for anybody looking for this rule, it was ultimately named [no-untyped-mock-factory](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-untyped-mock-factory.md).
The release notes and changelog picked up the original working name.

G-Rath:
@Blfrg thanks for flagging - I've updated the release notes and have opened https://github.com/jest-community/eslint-plugin-jest/pull/1317 updating the changelog

PR #1320, opened 2023-01-03
refactor: use `Object.fromEntries` to build rule config maps

NotWoods:
This is a tiny change to use [`Object.fromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) (supported in Node 12+) instead of a `.reduce` loop. 

This improves performance a little by making it a O(n) instead of O(n^2) operation, since spreading the accumulator means creating a new object each iteration of the loop.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,36 +27,27 @@ const importDefault = (moduleName: string) =>
 const rulesDir = join(__dirname, 'rules');
 const excludedFiles = ['__tests__', 'detectJestVersion', 'utils'];
 
-const rules = readdirSync(rulesDir)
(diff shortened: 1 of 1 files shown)


G-Rath (approved):
Sure, though I don't think it's worth marking it as a performance improvement unless you can actually show a clear speedup, since this is only done the first time the module is required and is for a relatively low number of files (which is why I didn't do this change in the first place)

NotWoods:
I pretty much only chose "perf" because I couldn't think of a better label :) Happy to switch to "refactor" or anything else

PR #1341, opened 2023-02-03
chore(deps): update @typescript-eslint/utils from 5.10.0 to 5.50.0

dominicfraser:
https://github.com/typescript-eslint/typescript-eslint/releases

The repo is using `@typescript-eslint/utils` `5.10.0` which is now a year old.

This package contains a [minimum and maximum TypeScript version](https://github.com/typescript-eslint/typescript-eslint/blob/a94872960423893200060fc9e5c178da33ec6de8/packages/typescript-estree/src/parseSettings/warnAboutTSVersion.ts#L9) and logs a console warning if using TS outside of this range.

In version `5.10.0` this is `>=3.3.1 <4.8.0`.

TypeScript is now on [version `4.9.5`](https://github.com/microsoft/TypeScript/releases) with the `5.0.0` beta released.

For anyone outside of this range when running linting the following error is logged: 

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.8.0

YOUR TYPESCRIPT VERSION: 4.8.4

Please only submit bug reports when using the officially supported version.

=============
```

This PR updates to the latest version, which has a range of `>=3.3.1 <5.0.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     ]
   },
   "dependencies": {
-    "@typescript-eslint/utils": "^5.10.0"
+    "@typescript-eslint/utils": "^5.50.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2799,6 +2799,16 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@typescript-eslint/scope-manager@npm:5.50.0":
+  version: 5.50.0
+  resolution: "@typescript-eslint/scope-manager@npm:5.50.0"
+  dependencies:
+    "@typescript-eslint/types": 5.50.0
+    "@typescript-eslint/visitor-keys": 5.50.0
+  checksum: bd49447a834c82cb130e6900644042c3a84195bf7a63483385e90b6454c65856d6f276c997cad6bf9c36c9d0cb168fdde625ce4c78c3b8bcce42da782270794b
+  languageName: node
+  linkType: hard
+
 "@typescript-eslint/type-utils@npm:5.49.0":
   version: 5.49.0
   resolution: "@typescript-eslint/type-utils@npm:5.49.0"
@@ -2823,6 +2833,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@typescript-eslint/types@npm:5.50.0":
+  version: 5.50.0
+  resolution: "@typescript-eslint/types@npm:5.50.0"
+  checksum: 1189c63d35abeec685dd519fd923926b884e63d5e10e4a9fe995aebfde59b8a2e10773090ec3ba32a0ec408746b18f6a454d9bedb0b6c7ce8b6066547144fb4d
+  languageName: node
+  linkType: hard
+
 "@typescript-eslint/typescript-estree@npm:5.49.0":
   version: 5.49.0
   resolution: "@typescript-eslint/typescript-estree@npm:5.49.0"
@@ -2841,7 +2858,25 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@typescript-eslint/utils@npm:5.49.0, @typescript-eslint/utils@npm:^5.10.0, @typescript-eslint/utils@npm:^5.38.1":
+"@typescript-eslint/typescript-estree@npm:5.50.0":
+  version: 5.50.0
+  resolution: "@typescript-eslint/typescript-estree@npm:5.50.0"
+  dependencies:
+    "@typescript-eslint/types": 5.50.0
+    "@typescript-eslint/visitor-keys": 5.50.0
+    debug: ^4.3.4
+    globby: ^11.1.0
+    is-glob: ^4.0.3
+    semver: ^7.3.7
+    tsutils: ^3.21.0
+  peerDependenciesMeta:
+    typescript:
+      optional: true
+  checksum: cb1ac8d39647da6d52750c713d9635750ed41245ec82f937a159a71ad3bf490ebabfad3b43eeca07bca39d60df30d3a2f31f8bed0061381731d92a62e284b867
+  languageName: node
+  linkType: hard
+
+"@typescript-eslint/utils@npm:5.49.0, @typescript-eslint/utils@npm:^5.38.1":
   version: 5.49.0
   resolution: "@typescript-eslint/utils@npm:5.49.0"
   dependencies:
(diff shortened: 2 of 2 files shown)


G-Rath:
No, we _support_ as far back as v5.10.0 but we use whatever version is installed by the downstream package manager which is guaranteed by our dependency constraint to be between v5.10.0 and v6, and that is generally the most recent version possible.

dominicfraser:
I would also have thought the caret version would allow updates to 5.50.0, but when installing the latest version of `eslint-plugin-jest`, and running and `npm dedupe`, this is the output of `npm ls @typescript-eslint/typescript-estree@5.15.0`:

```
│ └─┬ eslint-plugin-jest@27.2.1
│   └─┬ @typescript-eslint/utils@5.15.0    
│     └── @typescript-eslint/typescript-estree@5.15.0
└─┬ eslint-plugin-testing-library@5.10.0
  └─┬ @typescript-eslint/utils@5.50.0
    └── @typescript-eslint/typescript-estree@5.50.0
```

as I see no problems with the higher version it made sense to me to bring this package up to date, as it being a year old it seemed reasonable.

>  and in fact doing so might make it harder for people to continue using this plugin if they are not able to upgrade to a higher version of TypeScript.

Can you explain how this would be the case?

In `5.10.0` this is `>=3.3.1 <4.8.0`.

In `5.50.0` this is `>=3.3.1 <5.0.0`.

The lower bound does not change; it does not require people on an older version to update, it simply increases the upper limit of versions.

Is there another reason why this would cause problems?

G-Rath:
That is an implementation detail of the package manager, which can change between tool, versions, etc - fwiw if I do `npm i eslint-plugin-jest` on a fresh project I get v5.50.0; deduplication in particular has always been something that package managers have differed on. (I would expect you can run [`npm_remove_package_from_lock @typescript-eslint/utils`](https://github.com/G-Rath/scripts/blob/main/src/npm_remove_package_from_lock.js) and then `npm install` will give you the deduplication you want)

Likewise, it _actually_ causing problems with things like what TypeScript version is supported is not something we need to care about beyond the versions we support - so the fact that right now the version ranges happen to start with the same version doesn't mean that they will in future, but also is not something we need to think much about (i.e. while it would be going against semver, it is possible that v5.51.0 could drop support for those earlier versions of TypeScript).

Tightening the constraint in this case just means that there are less versions that can be used, restricting what people can use and making it harder to deduplicate.

To put it another way, is there a problem that this will fix?

dominicfraser:
Interesting, thanks.

> so the fact that right now the version ranges happen to start with the same version doesn't mean that they will in future

Agree with this - and so at the point of update I think it is something to check. The check in this case shows that it doesn't add a restriction, so I'd say it passes that check?

While it may mean that a consumer who is also using a very old version then ends up with minor duplication in their lockfile, as this is duplication without negative effect, i.e. it does not introduce breaking changes, it widens a range rather than tightens it, it does not seem a restrictive change there?

At what point does eslint-plugin-jest normally consider it good practice to bump a package? I'd normally assume that being a year out of date is reason enough for a bump, but it seems that is not the approach here? (I couldn't find a contributing guide so just raised this on our own practices).

Causing the need for manual intervention in lockfiles for consumers is the reason to update to me - while it may be a minor frustration, it is also a very minor update to fix it, from the perspective of our team's dependency practices.

G-Rath:
> i.e. it does not introduce breaking changes, it widens a range rather than tightens it, it does not seem a restrictive change there?

No, it only tightens the range because we've now gone from supporting `>= 5.10.0, < 6`, to `>= 5.50.0, < 6` - the second range is a subset of the first range, meaning all the versions that are compatible with that range are already compatible with the first range.

> At what point does eslint-plugin-jest normally consider it good practice to bump a package?

When we actually need to - our _constraint_ is a year out of date, but you can use the most recent versions ([which we ourselves are doing](https://github.com/jest-community/eslint-plugin-jest/blob/main/yarn.lock#L2844-L2860)) just fine so there's no need to update our constraint.

> Causing the need for manual intervention in lockfiles for consumers is the reason to update to me - while it may be a minor frustration, it is also a very minor update to fix it, from the perspective of our team's dependency practices.

I love an efficiently deduplicated tree as well, but this problem is not created by us - its a result of an inefficiency in the _package manager_; our constraint allows the deduplication you're after, it's the package manager whose not doing it - while yes we could attempt to force that deduplication with this PR, it could have the exact opposite effect on someone else's lockfile.

The answer here is what we're already doing: a constraint that is for the minimum version we require, but ideally no higher to allow the maximum deduplication potential.

dominicfraser:
>  tightens the range

The range I was referring to was the TS range - which extends rather than tightens.

I do not agree with you, but this also is not my package. I think the trade off after a year is towards the newer version of a package, rather than the older.

Thank you for the link to https://github.com/G-Rath/scripts/blob/main/src/npm_remove_package_from_lock.js, however so far I cannot get this to alter our lockfile to allow eslint-plugin-jest to be anything other than the snippet shared above.

G-Rath:
> The range I was referring to was the TS range - which extends rather than tightens.

That range is already that wide with our current constraint, because as a set that includes all versions it thus covers all those TypeScript ranges as well.

> however so far I cannot get this to alter our lockfile to allow eslint-plugin-jest to be anything other than the snippet shared above.

If you share a reproduction repo, I'm happy to give it a go locally - I do have an expanded version of that script locally which I think might make the difference, but I've not pushed it up yet because I've not tested it extensively.

dominicfraser:
Typically I cannot recreate our lockfile outside of the project, and as it has internally packages in it cannot share the exact file 😅 

It is indeed npm 8, though unfortunately the `update` command does not help - thank you for suggesting it.

G-Rath:
hmm I find that a bit suspect - all other things being up to date and equal, npm v8 should not be picking v5.15.0 for our plugin because that's just not how package constraints work and npm very much knows that.

The main other thing I'd try more as a proof of concept than anything else is removing our plugin + the typescript ones (and anything else you might have that depend on them), and then reinstalling them in a single `npm install` command (make sure to completely wipe out your `node_modules` first!)

I'll close this now though since we've established the issue lies with the package manager rather than with our constraint.

dominicfraser:
> our plugin + the typescript ones (and anything else you might have that depend on them), and then reinstalling them in a single npm install command (make sure to completely wipe out your node_modules first!)

This was very helpful, while I had already done that the process of thinking about it made me try in a nested package.json (don't ask) to where the lint commands are - that had a dependency on storybook, that also had an old version. Uninstalling everything in both directories, then reinstalling again, appears to have worked! Thank you.

G-Rath:
Awesome, happy to have helped!

PR #1402, opened 2023-07-13
chore: upgrade prettier

SimenB:
They still support Node 14, so this is safe: https://prettier.io/blog/2023/07/05/3.0.0.html#api-1

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2449,6 +2449,20 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@pkgr/utils@npm:^2.3.1":
+  version: 2.4.2
+  resolution: "@pkgr/utils@npm:2.4.2"
+  dependencies:
+    cross-spawn: ^7.0.3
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -4988,18 +5056,22 @@ __metadata:
...
+  version: 5.0.0
+  resolution: "eslint-plugin-prettier@npm:5.0.0"
   dependencies:
     prettier-linter-helpers: ^1.0.0
+    synckit: ^0.8.5
   peerDependencies:
-    eslint: ">=7.28.0"
-    prettier: ">=2.0.0"
+    "@types/eslint": ">=8.0.0"
+    eslint: ">=8.0.0"
```
SimenB, line 5067: hmm, this might actually stop us as we still support ESLint 7

SimenB: CI is happy, so I guess it doesn't matter. This isn't exposed outside of this repo anyways

PR #1403, opened 2023-07-13
chore: update dedent

SimenB:
For some reason, nested numbers aren't allowed in the types...


G-Rath:
That's because the types are wrong - I've opened https://github.com/dmnd/dedent/pull/44 to address this

PR #1410, opened 2023-07-31
chore: update `dedent` dep

SimenB:
See https://github.com/dmnd/dedent/pull/65#issuecomment-1657844742

However, this change actually exposed a false negative in the test. Using manual string gives the same error.

```
    {
      code:
        'const x = "snapshot";' +
        '\n' +
        'expect(1).toMatchSnapshot(`my ${x}`);',
      options: ['always'],
    },
```

(with eslint removing escapes).

I think this is by design.

https://github.com/jest-community/eslint-plugin-jest/blob/a058f22f94774eeea7980ea2d1f24c6808bf3e2c/src/rules/utils/accessors.ts#L57-L58

We explicitly bail if there are arguments in the template string.

Closes #1411


G-Rath (approved):
yeah tbh the tests with escaped interpolation have confused me when I previously looked at them... 😅

PR #1425, opened 2023-09-02
feat: add no-confusing-set-time rule

### LINKED ISSUE #1366: Add rule to check placement of jest.setTimeout
The `jest.setTimeout()` sets default timeout for whole test file regardless of its actual placement in the file. The Jest API documentation [mentions it](https://jestjs.io/docs/jest-object#jestsettimeouttimeout), but it's not clear how this works under the hood. In fact, only last call applies and sets the timeout for all tests before they are started. As it might be tempting to place that command anywhere in the file, it would be useful to create an ESLint rule that:

* checks if jest.setTimeout() is placed in global scope only (this itself would have immense value)
* checks if jest.setTimeout() is only once in the file (nice to have)
* checks if jest.setTimeout() is placed before any Jest test methods (before*, after*, test*, it*, describe*) (nice to have)

As an example, in this test file, the timeout for all tests is **_set to 800ms_**:

```js
jest.setTimeout(1000);

describe('A', () => {
  beforeEach(async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });}); // applied timeout: 800ms

  it('A.1', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });}); // applied timeout: 800ms

  jest.setTimeout(2000);

  it('A.2', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });}); // applied timeout: 800ms
});

jest.setTimeout(1500);

describe('B', () => {
  it('B.1', async () => { await new Promise(resolve => {
    jest.setTimeout(1000);
    setTimeout(resolve, 10000).unref();
  });}); // applied timeout: 800ms

  it('B.2', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });}); // applied timeout: 800ms

  jest.setTimeout(500);
});

jest.setTimeout(800); // only value that applies
```

eryue0220:
Fix: #1366

### DIFF
--- a/docs/rules/no-conditional-expect.md
+++ b/docs/rules/no-conditional-expect.md
@@ -118,7 +118,7 @@ thrown by the wrapped function:
 ```typescript
 class NoErrorThrownError extends Error {}
 
-const getError = async <TError>(call: () => unknown): Promise<TError> => {
+const getError = async <TError,>(call: () => unknown): Promise<TError> => {
   try {
     await call();
 
--- a/docs/rules/no-conditional-expect.md
+++ b/docs/rules/no-conditional-expect.md
@@ -118,7 +118,7 @@ thrown by the wrapped function:
 ```typescript
 class NoErrorThrownError extends Error {}
 
-const getError = async <TError,>(call: () => unknown): Promise<TError> => {
+const getError = async <TError>(call: () => unknown): Promise<TError> => {
   try {
     await call();
 
--- a/docs/rules/no-confusing-set-timeout.md
+++ b/docs/rules/no-confusing-set-timeout.md
@@ -1,4 +1,4 @@
-# Disallow using confusing setTimeout in test (`no-confusing-set-timeout`)
+# Disallow confusing usages of jest.setTimeout (`no-confusing-set-timeout`)
 
 <!-- end auto-generated rule header -->
 
--- a/docs/rules/no-confusing-set-timeout.md
+++ b/docs/rules/no-confusing-set-timeout.md
@@ -1,0 +1,61 @@
+# Disallow using confusing setTimeout in test (`no-confusing-set-timeout`)
+
+<!-- end auto-generated rule header -->
+
+This rule will raise a warning about confusing `jest-setTimeout` in test.
+
+## Rule details
+
+This rule describes some tricky ways about `jest.setTimeout` that should not
+recommend in Jest:
+
+- should set `jest.setTimeout` in any testsuite methods before(such as
+  `describe`, `test` or `it`);
+- should set `jest.setTimeout` in global scope.
+- should only call `jest.setTimeout` once in a single test file;
+
+Examples of **incorrect** code for this rule:
+
+```js
+describe('A', () => {
+  jest.setTimeout(1000);
+  it('test-description', () => {
+    // test logic;
+  });
+});
+
+describe('B', () => {
+  it('test-description', () => {
+    jest.setTimeout(1000);
+    // test logic;
+  });
+});
+
+test('C', () => {
+  jest.setTimeout(1000);
+});
+
+describe('D', () => {
+  beforeEach(() => {
+    jest.setTimeout(1000);
+  });
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+jest.setTimeout(500);
+test('A', () => {
+  // do some stuff
+});
+```
+
+```js
+jest.setTimeout(1000);
+describe('B', () => {
+  it('test-description', () => {
+    // test logic;
+  });
+});
+```
--- a/docs/rules/no-confusing-set-timeout.md
+++ b/docs/rules/no-confusing-set-timeout.md
@@ -2,19 +2,18 @@
 
 <!-- end auto-generated rule header -->
 
-`jest.setTimeout` can be called multiple times anywhere within a single test
-file. However, only the last call will have an effect, and it will actually be
-invoked before any other jest functions.
+While `jest.setTimeout` can be called multiple times anywhere within a single
+test file only the last call before any test functions run will have an effect.
 
 ## Rule details
 
-This rule describes some tricky ways about `jest.setTimeout` that should not
-recommend in Jest:
+his rule checks for several confusing usages of `jest.setTimeout` that looks
+like it applies to specific tests within the same file, such as:
 
-- should set `jest.setTimeout` in any testsuite methods before(such as
-  `describe`, `test` or `it`);
-- should set `jest.setTimeout` in global scope.
-- should only call `jest.setTimeout` once in a single test file;
+- being called anywhere other than in global scope
+- being called multiple times
+- being called after other Jest functions like hooks, `describe`, `test`, or
+  `it`
 
 Examples of **incorrect** code for this rule:
 
--- a/docs/rules/no-confusing-set-timeout.md
+++ b/docs/rules/no-confusing-set-timeout.md
@@ -2,7 +2,9 @@
 
 <!-- end auto-generated rule header -->
 
-This rule will raise a warning about confusing `jest-setTimeout` in test.
+`jest.setTimeout` can be called multiple times anywhere within a single test
+file. However, only the last call will have an effect, and it will actually be
+invoked before any other jest functions.
 
 ## Rule details
 
@@ -17,25 +19,25 @@ recommend in Jest:
 Examples of **incorrect** code for this rule:
 
 ```js
-describe('A', () => {
+describe('test foo', () => {
   jest.setTimeout(1000);
   it('test-description', () => {
     // test logic;
   });
 });
 
-describe('B', () => {
+describe('test bar', () => {
   it('test-description', () => {
     jest.setTimeout(1000);
     // test logic;
   });
 });
 
-test('C', () => {
+test('foo-bar', () => {
   jest.setTimeout(1000);
 });
 
-describe('D', () => {
+describe('unit test', () => {
   beforeEach(() => {
     jest.setTimeout(1000);
   });
@@ -46,14 +48,14 @@ Examples of **correct** code for this rule:
 
 ```js
 jest.setTimeout(500);
-test('A', () => {
+test('test test', () => {
   // do some stuff
 });
 ```
 
 ```js
 jest.setTimeout(1000);
-describe('B', () => {
+describe('test bar bar', () => {
   it('test-description', () => {
     // test logic;
   });
--- a/docs/rules/no-confusing-set-timeout.md
+++ b/docs/rules/no-confusing-set-timeout.md
@@ -7,7 +7,7 @@ test file only the last call before any test functions run will have an effect.
 
 ## Rule details
 
-his rule checks for several confusing usages of `jest.setTimeout` that looks
+this rule checks for several confusing usages of `jest.setTimeout` that looks
 like it applies to specific tests within the same file, such as:
 
 - being called anywhere other than in global scope
--- a/src/rules/__tests__/no-confusing-set-timeout.test.ts
+++ b/src/rules/__tests__/no-confusing-set-timeout.test.ts
@@ -1,0 +1,172 @@
+import { TSESLint } from '@typescript-eslint/utils';
+import dedent from 'dedent';
+import rule from '../no-confusing-set-timeout';
+import { espreeParser } from './test-utils';
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: espreeParser,
+  parserOptions: {
+    ecmaVersion: 2020,
+  },
+});
+
+ruleTester.run('no-confusing-set-timeout', rule, {
+  valid: [
+    dedent`
+      jest.setTimeout(1000);
+      describe('A', () => {
+        beforeEach(async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+        it('A.1', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+        it('A.2', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+      });
+    `,
+    {
+      code: dedent`
+        import { jest } from '@jest/globals';
+        jest.setTimeout(800);
+        describe('A', () => {
+          beforeEach(async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+          it('A.1', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+          it('A.2', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+        });
+      `,
+      parserOptions: { sourceType: 'module' },
+    },
+    dedent`
+      function handler() {}
+      jest.setTimeout(800);
+      describe('A', () => {
+        beforeEach(async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+        it('A.1', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+        it('A.2', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+      });
+    `,
+    dedent`
+      const jest = require('@jest/globals');
+      jest.setTimeout(800);
+      describe('A', () => {
+        beforeEach(async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+        it('A.1', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+        it('A.2', async () => { await new Promise(resolve => { setTimeout(resolve, 10000).unref(); });});
+      });
+    `,
+  ],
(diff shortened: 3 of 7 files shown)


docs/rules/no-confusing-set-timeout.md
```diff
@@ -0,0 +1,61 @@
+# Disallow using confusing setTimeout in test (`no-confusing-set-timeout`)
+
+<!-- end auto-generated rule header -->
+
+This rule will raise a warning about confusing `jest-setTimeout` in test.
```
G-Rath, line 5: Generally this section tends to describe the situation that the rule addresses rather than the specifics of the rule itself - so it shouldn't start with "this rule" since that's what the "rule details" section generally starts with.

Instead this should talk more about the situation as described in #1366 - something like "While `jest.setTimeout` can be called anywhere in a test file implying its stateful, it actually applies to the entire file and is called before any test starts..."

docs/rules/no-confusing-set-timeout.md
```diff
@@ -0,0 +1,61 @@
...
+
+- should set `jest.setTimeout` in any testsuite methods before(such as
+  `describe`, `test` or `it`);
+- should set `jest.setTimeout` in global scope.
+- should only call `jest.setTimeout` once in a single test file;
+
+Examples of **incorrect** code for this rule:
+
+```js
+describe('A', () => {
```
G-Rath, line 20: Can you expand these examples so they're not just using single letters for the titles

src/rules/__tests__/no-confusing-set-timeout.test.ts
```diff
@@ -0,0 +1,172 @@
...
+import { espreeParser } from './test-utils';
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: espreeParser,
+  parserOptions: {
+    ecmaVersion: 2020,
+  },
+});
+
+ruleTester.run('no-confusing-set-timeout', rule, {
```
G-Rath, line 13: These look like a good start, but it would be good to also have some more granular ones to make it easier to narrow down problems, i.e. single line calls to `setTimeout`, `window.setTimeout`, etc, a case with literally just two `jest.setTimeout` calls, and so on.

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,108 @@
...
+    'test',
+    'it',
+    'beforeEach',
+    'afterEach',
+    'before',
+    'after',
+  ].includes(name);
+}
+
+export default createRule({
```
G-Rath, line 44: Please rewrite this to use [`parseJestFnCall`](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/utils/parseJestFnCall.ts#L199) - it will (more correctly) handle the logic you have above for detecting if an expression is a call to `jest.setTimeout` and other such Jest functions; have a look at other rules if you want pointers on how you can use it.

As a general rule of thumb, you should _never_ need to write custom logic in a rule to determine if an expression is related to Jest or if its global vs imported; that should all be handled by the `parseJestFn` util :)

([`require-hook` actually does a bit of extra work](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/require-hook.ts#L15-L24) which you might need here in which case feel free to do so - I'll review that some time this week to refresh my memory on why that might be, or if it's just old code that can be removed)

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,108 @@
...
+    docs: {
+      category: 'Best Practices',
+      description: 'Disallow using confusing setTimeout in test',
+      recommended: false,
+    },
+    messages: {
+      globalSetTimeout: '`jest.setTimeout` should be call in `global` scope.',
+      onlyOneSetTimeout: 'Only the last one `jest.setTimeout` call work.',
+      topSetTimeout:
+        'The `jest.setTimout` should be placed before all of testsuite methods.',
```
G-Rath, line 56:
```suggestion
      globalSetTimeout: '`jest.setTimeout` should only be called in the global scope',
      multipleSetTimeouts: 'Do not call `jest.setTimeout` multiple times, as only the last call will have an effect',
      orderSetTimeout: '`jest.setTimeout` should be placed before any other jest methods',
```

G-Rath (changes requested):
This is a good start but requires more work - I've done an initial review to give you some pointers, but really your whole implementation needs to be redone to use `parseJestFnCall` so I've not reviewed that for now.

eryue0220:
Thanks for your excellent guidance and review. I've upgrade according your suggestion.

src/rules/no-confusing-set-timeout.ts
eryue0220: > Please rewrite this to use [`parseJestFnCall`](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/utils/parseJestFnCall.ts#L199) - it will (more correctly) handle the logic you have above for detecting if an expression is a call to `jest.setTimeout` and other such Jest functions; have a look at other rules if you want pointers on how you can use it.
> 
> As a general rule of thumb, you should _never_ need to write custom logic in a rule to determine if an expression is related to Jest or if its global vs imported; that should all be handled by the `parseJestFn` util :)
> 
> ([`require-hook` actually does a bit of extra work](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/require-hook.ts#L15-L24) which you might need here in which case feel free to do so - I'll review that some time this week to refresh my memory on why that might be, or if it's just old code that can be removed)


Yes, you're right. Some of the test cases I've written may be misleading. I need to find a way to determine whether jest.setTimeout is in the global scope. However, in the provided example, jest is imported first. I've optimized the test cases to accurately reflect the scenario. But the `parseJestFnCall` you've said really help me a lot.

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,81 @@
...
+    let nonJestTimeout = 0;
+
+    return {
+      CallExpression(node) {
+        const scope = context.getScope();
+        const jestFnCall = parseJestFnCall(node, context);
+
+        if (!jestFnCall) return;
+
+        if (
```
G-Rath, line 44: If `jestFnCall` returns non-null it means that we believe the call is something from Jest, even if the type is `unknown` - so we should be fine to just check if the call is to `setTimeout` or not; and given that I'd move all that logic into `isJsetSetTimeout` (i.e. have it take the `ParsedJestFnCall` and handle checking that the `type` is `jest`)

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,81 @@
...
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const errors: ErrorType[] = [];
+    let callJestTimeout = 0;
+    let nonJestTimeout = 0;
+
+    return {
+      CallExpression(node) {
```
G-Rath, line 38: My gut is saying we should be able to make this simpler by focusing on the conditions for each message we're checking for kind of like bailing early, rather than tracking counts - i.e.
  * if we're not in the global scope, then any use of `setTimeout` should trigger `globalSetTimeout`
  * if we've already see a usage of `jest.setTimeout`, then any other usage we come across should trigger `multipleSetTimeout`
  * if we see a `setTimeout` after seeing any other jest function, then that should trigger `orderSetTimeout`

Could you explore this and see if it is possible?

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,81 @@
...
+        const scope = context.getScope();
+        const jestFnCall = parseJestFnCall(node, context);
+
+        if (!jestFnCall) return;
```
G-Rath, line 42: nit: prefer having braces

```suggestion
        if (!jestFnCall) {
          return;
        }
```

(I should setup the lint rule for this tbh 😅)

G-Rath (changes requested):
Have done an initial re-review of the code - its coming together nicely; I'll have a look at the rest during the week, but in the meantime if you have the time to address what I've mentioned currently that'd be great

src/rules/no-confusing-set-timeout.ts
eryue0220: I will solve in another PR.

G-Rath: I'd prefer if you address this in this PR - if you're talking about adding the lint rule, I can do that now

eryue0220: > I'd prefer if you address this in this PR - if you're talking about adding the lint rule, I've opened #1429 for that

Ok, I fixed it.

docs/rules/no-conditional-expect.md
```diff
@@ -118,7 +118,7 @@ thrown by the wrapped function:
 ```typescript
 class NoErrorThrownError extends Error {}
 
-const getError = async <TError>(call: () => unknown): Promise<TError> => {
+const getError = async <TError,>(call: () => unknown): Promise<TError> => {
```
SimenB, line 121: https://github.com/jest-community/eslint-plugin-jest/pull/1427

weird...

SimenB: ah, ci hadn't run

docs/rules/no-confusing-set-timeout.md
```diff
@@ -0,0 +1,63 @@
...
+
+<!-- end auto-generated rule header -->
+
+`jest.setTimeout` can be called multiple times anywhere within a single test
+file. However, only the last call will have an effect, and it will actually be
+invoked before any other jest functions.
+
+## Rule details
+
+This rule describes some tricky ways about `jest.setTimeout` that should not
```
SimenB, line 11:
```suggestion
This rule describes some tricky ways about `jest.setTimeout` that is not
```

docs/rules/no-confusing-set-timeout.md
```diff
@@ -0,0 +1,63 @@
...
+recommend in Jest:
+
+- should set `jest.setTimeout` in any testsuite methods before(such as
```
SimenB, line 14:
```suggestion
- should set `jest.setTimeout` in any test suite methods before (such as
```

docs/rules/no-confusing-set-timeout.md
```diff
@@ -0,0 +1,63 @@
+# Disallow using confusing setTimeout in test (`no-confusing-set-timeout`)
+
+<!-- end auto-generated rule header -->
+
+`jest.setTimeout` can be called multiple times anywhere within a single test
+file. However, only the last call will have an effect, and it will actually be
+invoked before any other jest functions.
```
G-Rath, line 7:
```suggestion
While `jest.setTimeout` can be called multiple times anywhere within a single test
file only the last call before any test functions run will have an effect.
```

docs/rules/no-confusing-set-timeout.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Rule details
+
+This rule describes some tricky ways about `jest.setTimeout` that should not
+recommend in Jest:
```
G-Rath, line 12: How about this:

```suggestion
This rule checks for several confusing usages of `jest.setTimeout` that looks like it applies to specific tests within the same file, such as:

- being called anywhere other than in global scope
- being called multiple times
- being called after other Jest functions like hooks, `describe`, `test`, or `it`
```

docs/rules/no-confusing-set-timeout.md
G-Rath: fwiw this should actually be "are" since its multiple; but I've suggested a bigger rewrite to this below.

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,66 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+import { createRule, getNodeName, parseJestFnCall } from './utils';
+
+function isJestSetTimeout(node: TSESTree.Node) {
+  return getNodeName(node) === 'jest.setTimeout';
+}
```
G-Rath, line 6: This check is not correct because it does not account for aliases i.e. this will not get caught:

```
import { jest as Jest } from '@jest/globals';
{
  Jest.setTimeout(800);
}
```

While a bit contrived, this is why `parseJestFnCall` exists and is what you should be checking instead:

```
function isJestSetTimeout(jestFnCall: ParsedJestFnCall) {
  return (
    jestFnCall.type === 'jest' &&
    isIdentifier(jestFnCall.members[0], 'setTimeout')
  );
}
```

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,66 @@
...
+    return {
+      CallExpression(node) {
+        const scope = context.getScope();
+        const jestFnCall = parseJestFnCall(node, context);
+
+        if (!jestFnCall) {
+          return;
+        }
+
+        const result = isJestSetTimeout(node);
```
G-Rath, line 40: let's call this something like `thisIsJestSetTimeoutCall` - its a bit verbose, but there's no downside and I think its important to be clear given all the "is jest settimeout" names we've got going on

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        if (!result) {
+          if (jestFnCall.type !== 'unknown') {
+            shouldEmitOrderSetTimeout = true;
+          }
```
G-Rath, line 45: By definition there is no known Jest function call of type `unknown` - it only exists to gracefully handle if Jest ever did get a new type in older versions of the plugin (since we'd have to release a new version adding support); so you shouldn't worry about it here:

```suggestion
          shouldEmitOrderSetTimeout = true;
```

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,66 @@
...
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Disallow using confusing setTimeout in test',
+      recommended: false,
+    },
+    messages: {
+      globalSetTimeout: '`jest.setTimeout` should be call in `global` scope.',
+      multipleSetTimeouts:
+        'Do not call `jest.setTimeout` multiple times, as only the last call will have an effect.',
```
G-Rath, line 19: Linting messages shouldn't use full stops because they might appear in sentences constructed by IDEs and such:

```suggestion
      globalSetTimeout: '`jest.setTimeout` should be call in `global` scope',
      multipleSetTimeouts:
        'Do not call `jest.setTimeout` multiple times, as only the last call will have an effect',
```

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,66 @@
...
+function isJestSetTimeout(node: TSESTree.Node) {
+  return getNodeName(node) === 'jest.setTimeout';
+}
+
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Disallow using confusing setTimeout in test',
```
G-Rath, line 13:
```suggestion
      description: 'Disallow confusing usages of jest.setTimeout',
```

G-Rath (changes requested):

src/rules/no-confusing-set-timeout.ts
eryue0220: Yeah, naming makes my head burn out.

eryue0220:
Done~ Thanks for your guidances and patience

src/rules/no-confusing-set-timeout.ts
```diff
@@ -0,0 +1,72 @@
...
+    return {
+      CallExpression(node) {
+        const scope = context.getScope();
+        const jestFnCall = parseJestFnCall(node, context);
+
+        if (!jestFnCall) {
+          return;
+        }
+
+        const thisIsJestSetTimeoutCall = isJestSetTimeout(jestFnCall);
```
G-Rath, line 48: We actually don't need this variable or your `else` block at the end because we only do more work if this is a `jest.setTimeout` - so you can just do:

```
if (!isJestSetTimeout(jestFnCall)) {
  shouldEmitOrderSetTimeout = true;

  return;
}

// ...

seenJestTimeout = true;
```

docs/rules/no-confusing-set-timeout.md
```diff
@@ -0,0 +1,62 @@
+# Disallow using confusing setTimeout in test (`no-confusing-set-timeout`)
+
+<!-- end auto-generated rule header -->
+
+While `jest.setTimeout` can be called multiple times anywhere within a single
+test file only the last call before any test functions run will have an effect.
+
+## Rule details
+
+his rule checks for several confusing usages of `jest.setTimeout` that looks
```
G-Rath, line 10:
```suggestion
This rule checks for several confusing usages of `jest.setTimeout` that looks
```

G-Rath (changes requested):
Just got a small cleanup and a typo, and then I think this is good to go!

SimenB:
Prettier is still unhappy about something 😅

PR #1440, opened 2023-09-29
`no-focused-tests` now highlights only the `fit` and `fdescribe` part of the node

### LINKED ISSUE #1439: `no-focused-tests` should highlight only `fit` and `fdescribe` instead of the whole thing
When I use this rule, which is recommended also, while I fix a test, I usually do `fit` while I fix it so the whole file doesnt have to run.

Ideally, this rule should stop me from committing with the `fit` still. But when it makes the whole test red, its really hard to read and fix a test.

This rule should red-highlight the `fit` part of `fit("checks something", () => {`, not the whole thing.

pvinis:
Fixes #1439.

<table>
<tr>
 <td>
Before
 <td>
After
<tr>
 <td>
Screenshot 2023-09-29 at 20 05 25

 <td>
Screenshot 2023-09-29 at 20 05 45

</table>

We have control over what is highlighter by eslint. The `report` gives us the power to do that. You can see before that the editor just guesses, and sometimes it works, some it doesnt. You also see that after, things all are highlighted correctly.

### DIFF
--- a/src/rules/no-focused-tests.ts
+++ b/src/rules/no-focused-tests.ts
@@ -31,6 +31,7 @@ export default createRule({
           context.report({
             messageId: 'focusedTest',
             node,
+            loc: jestFnCall.head.node.loc,
             suggest: [
               {
                 messageId: 'suggestRemoveFocus',
@@ -63,6 +64,7 @@ export default createRule({
         context.report({
           messageId: 'focusedTest',
           node: onlyNode,
+          loc: onlyNode.loc,
           suggest: [
             {
               messageId: 'suggestRemoveFocus',


src/rules/no-focused-tests.ts
```diff
@@ -31,6 +31,7 @@ export default createRule({
           context.report({
             messageId: 'focusedTest',
             node,
+            loc: jestFnCall.head.node.loc,
```
pvinis, line 34: this is the main change here, and the following

package.json
```diff
@@ -122,6 +122,7 @@
     "eslint-plugin-eslint-comments": "^3.1.2",
     "eslint-plugin-eslint-plugin": "^5.0.6",
     "eslint-plugin-import": "^2.25.1",
+    "eslint-plugin-jest": "link:./lib",
```
pvinis, line 125: this was so i can test the changes on the editor directly. super useful, I can make it a separate PR if you prefer.

G-Rath:
I'm currently doing a review of all our rules - I would recommend not continuing with this PR for now.

pvinis:
oh god, commitlint. my worst enemy. renaming..

pvinis:
this PR is ready if you want it. Up to you, but I'd love to be able to ready my tests when i do a `fit` rather than see an ocean of red 😅

pvinis:
Actually, I see the CI fails for the playground thing I added. I think I can keep this PR super simple, and then I'll make another PR with the playground, and if you see this as useful, you can help me fix it.

G-Rath:
I'm not interested in the playground stuff right now - I should have a PR landed for updating some of the ranges shortly, though in the meantime you can switch to using `.skip`.

Note that as I said in the issue, this is still primarily a IDE problem so I'm not making any promises that all rules will have their reporting locations changed to resolve "seas of red" that you get in VSCode - I'll be making changes based on what I feel is fair and to ensure consistency within the rules (but yes `no-focused-tests` will change)

pvinis:
Makes sense. I cleaned this up, now it's simple and focused as it can be. I will not use the `.only` versions, since `fit` is my preference, I will wait for this to land and whatever other fixes you end up making.

Also the lighlighting is *not* an IDE problem. ESLint very clearly should do this https://eslint.org/docs/latest/extend/custom-rules#reporting-problems, check the `loc` bit. IDEs just listen to what the eslint server tells them. If eslint (through the rules of this repo) say the problem is the node, then it will highlight whatever the node is, which in the case of `fit` is the whole thing. If we make the rule tell the server to tell the IDE the right location of the error, then we are golden 👍.

Thanks for your comments and your work!

PR #1453, opened 2023-10-20
fix: replace use of deprecated methods

G-Rath:
Per https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/

Marking as a draft for now since there isn't an immediate need to land this and I want to dig a bit deeper into ESLint v9 etc in case there are other deprecations that might be upcoming.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import { readdirSync } from 'fs';
 import { join, parse } from 'path';
-import type { TSESLint } from '@typescript-eslint/utils';
+import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
 import {
   name as packageName,
   version as packageVersion,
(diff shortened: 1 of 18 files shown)


src/index.ts
```diff
@@ -19,6 +19,32 @@ declare module '@typescript-eslint/utils/dist/ts-eslint/Rule' {
   }
 }
 
+declare module '@typescript-eslint/utils/dist/ts-eslint/SourceCode' {
```
SimenB, line 22: can we avoid this by upgrading to v6?

src/rules/utils/misc.ts
```diff
@@ -229,3 +229,63 @@ export const getFirstMatcherArg = (
 
   return followTypeAssertionChain(firstArg);
 };
+
+/* istanbul ignore next */
```
SimenB, line 233: why do we need to ignore coverage?

SimenB (approved):
woo!

should we set up CI to test against v9?

src/rules/utils/misc.ts
G-Rath: because only one branch in each function should ever get used depending on what ESLint version you're using, and coverage is only collected against a single ESLint version at a time (technically we could merge reports for CI but that'd still be a pain locally)

G-Rath:
> should we set up CI to test against v9?

Yup, though I'll do that in a follow-up PR

src/index.ts
G-Rath: Yeah I implemented the types in https://github.com/typescript-eslint/typescript-eslint/pull/7812

SimenB: sweet! we should probably do that in a major

Issue #1408, opened 2023-07-26
Support ESLint "flat" configs


michaelfaith:
I searched through the issues for something similar, and didn't find anything related to supporting the new [Flat Config](https://eslint.org/docs/latest/use/configure/configuration-files-new).  So I thought it would make sense to create this for tracking.

I'm currently building out a config using `eslint.config.js` and the new flat config structure.  When using this plugin, I receive the following error:
```bash
TypeError: Key "plugins": Key "0": Expected an object.
    at Object.validate (D:\src\ux\.yarn\cache\eslint-npm-8.45.0-bb6c5f1226-3e6dcce5cc.zip\node_modules\eslint\lib\config\flat-config-schema.js:315:23)
    at ObjectSchema.validate (D:\src\ux\.yarn\cache\@humanwhocodes-object-schema-npm-1.2.1-43b3ca2646-a824a1ec31.zip\node_modules\@humanwhocodes\object-schema\src\object-schema.js:218:35)
    at D:\src\ux\.yarn\cache\@humanwhocodes-object-schema-npm-1.2.1-43b3ca2646-a824a1ec31.zip\node_modules\@humanwhocodes\object-schema\src\object-schema.js:171:18
    at Array.reduce (<anonymous>)
    at ObjectSchema.merge (D:\src\ux\.yarn\cache\@humanwhocodes-object-schema-npm-1.2.1-43b3ca2646-a824a1ec31.zip\node_modules\@humanwhocodes\object-schema\src\object-schema.js:169:24)
    at D:\src\ux\.yarn\cache\@humanwhocodes-config-array-npm-0.11.10-33012f902e-1b1302e240.zip\node_modules\@humanwhocodes\config-array\api.js:916:42
    at Array.reduce (<anonymous>)
    at FlatConfigArray.getConfig (D:\src\ux\.yarn\cache\@humanwhocodes-config-array-npm-0.11.10-33012f902e-1b1302e240.zip\node_modules\@humanwhocodes\config-array\api.js:915:39)
    at FlatConfigArray.isFileIgnored (D:\src\ux\.yarn\cache\@humanwhocodes-config-array-npm-0.11.10-33012f902e-1b1302e240.zip\node_modules\@humanwhocodes\config-array\api.js:943:15)
[...]

PR #1454, opened 2023-10-21
feat: include plugin `meta` information for ESLint v9

G-Rath:
Relates to #1408

See https://eslint.org/docs/latest/extend/plugin-migration-flat-config#adding-plugin-meta-information

It seems like this is backwards compatible - I've at least done a quick test with an older version of ESLint and it didn't care 🤷

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,10 @@
 import { readdirSync } from 'fs';
 import { join, parse } from 'path';
 import type { TSESLint } from '@typescript-eslint/utils';
+import {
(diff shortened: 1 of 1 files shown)


SimenB (commented):
Yeah, should be fine 👍

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/61e66017f0c582c8a2537ca23b2e858e1fad8908/index.js#L30-L33

SimenB (approved):

PR #1455, opened 2023-10-21
chore: update `moduleResolution` and `module` to `node16`

G-Rath:
I'm honestly still trying to wrap my head around these new settings, but this is required for `@typescript-eslint/utils` v6 and doesn't seem to result in any different output (which makes sense since we're compiling with Babel) so I think this is safe to land.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,8 @@
 {
   "extends": "@tsconfig/node14/tsconfig.json",
   "compilerOptions": {
-    "moduleResolution": "node",
-    "module": "commonjs",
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "extends": "@tsconfig/node14/tsconfig.json",
   "compilerOptions": {
-    "moduleResolution": "node",
-    "module": "commonjs",
+    "moduleResolution": "node16",
```
SimenB, line 4: I think we can remove them, as `@tsconfig/node14/` sets it?

PR #1460, opened 2023-10-26
feat(valid-title): allow disabling titleMustBeString on test titles

### LINKED ISSUE #1234: [New option] [valid-title] - Allow types that yield string.
Right now there is ability to `{ ignoreTypeOfDescribeName: true }` that ignores any type whether its a `number` or a `function`.

How about we have a rule that restricts everything that does not yield type `typeof {title} === 'string'`. Or allows only the things that yield a string title.

Why?

Because a lot of codebases use `action_types`, `enums`, `consts`, etc.

Right now:

```js
// "jest/valid-title": ["error", { ignoreTypeOfDescribeName: true }]

it.todo(123) // is good

// "jest/valid-title": ["error", { ignoreTypeOfDescribeName: false }]

const SOME_ACTION_TYPE = "SOME_ACTION_TYPE"
it.todo(SOME_ACTION_TYPE) // error
```

Nokel81:
Resolves #1234

### DIFF
--- a/docs/rules/valid-title.md
+++ b/docs/rules/valid-title.md
@@ -51,10 +51,9 @@ xtest('foo', () => {});
 
 **titleMustBeString**
 
-Titles for test blocks should always be a string.
-
-This is also applied to `describe` blocks by default, but can be turned off via
-the `ignoreTypeOfDescribeName` or `ignoreTypeOfTestName` options:
+Titles for `describe`, `test`, and `it` blocks should always be a string; you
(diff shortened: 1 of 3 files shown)


Nokel81:
The danger failure is from 403's, so I think unrelated to my code changes.

G-Rath:
I think it might be a contributor thing? @SimenB would be the one to check - but had it been successful it would be reported that your PR does not have a description.

G-Rath (approved):

SimenB:
I think d8b10b47e7040199ac105b1170fd784f78d6a7ee fixes the CI issue

PR #1486, opened 2024-01-08
chore: do not use resolutions for typescript dependency

SimenB:


SimenB:
image

I'm guessing this is something deep inside ts-eslint…

SimenB:
#1525

PR #1487, opened 2024-01-10
chore: remove syntax error from test

SimenB:
https://github.com/typescript-eslint/typescript-eslint/issues/8232


G-Rath (approved):
Tbh we should look at resyncing our tests from the upstream at some point

PR #1490, opened 2024-01-16
feat: prefer importing jest globals [new rule]

### LINKED ISSUE #1101: [new-rule] prefer importing from @jest/globals

MadeinFrance:
- Solves #1101

Issue: https://github.com/jest-community/eslint-plugin-jest/issues/1101

# Prefer importing Jest globals (`prefer-importing-jest-globals`)

🔧 This rule is automatically fixable by the
[`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).

This rule aims to enforce explicit imports from `@jest/globals`.

1. This is useful for ensuring that the Jest APIs are imported the same way in
   the codebase.
2. When you can't modify Jest's
   [`injectGlobals`](https://jestjs.io/docs/configuration#injectglobals-boolean)
   configuration property, this rule can help to ensure that the Jest globals
   are imported explicitly and facilitate a migration to `@jest/globals`.

## Rule details

Examples of **incorrect** code for this rule

```js
/* eslint jest/prefer-importing-jest-globals: "error" */

describe('foo', () => {
  it('accepts this input', () => {
    // ...
  });
});
```

Examples of **correct** code for this rule

```js
/* eslint jest/prefer-importing-jest-globals: "error" */

import { describe, it } from '@jest/globals';

describe('foo', () => {
  it('accepts this input', () => {
    // ...
  });
});
```

## Further Reading

- [Documentation](https://jestjs.io/docs/api)

image


src/rules/prefer-jest-globals.ts
```diff
@@ -0,0 +1,60 @@
...
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const jestFunctions = Object.keys(globalsJson);
+    const importedJestFunctions: any[] = [];
+    const usedJestFunctions = new Set();
+
+    return {
```
G-Rath, line 26: Can you look at using `parseJestFnCall` for this? it already handles determining if a Jest function is imported or not as part of its core logic, so it should greatly simplify the logic

src/rules/__tests__/prefer-jest-globals.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: espreeParser,
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('prefer-jest-globals.test', rule, {
```
G-Rath, line 14: we need more tests than this - even though this should just be a thin wrapper for `parseJestFnCall` which already has a lot of tests, we should stick a few here anyway, including checking a few different calls with chains and aliases (both via `import` and doing e.g. `const context = describe`, which it's fair _not_ to report on `context`)

yarn.lock
G-Rath, line 1: I don't see a reason for this needing to have changed - can you revert this please?

src/rules/prefer-jest-globals.ts
```diff
@@ -0,0 +1,60 @@
...
+    docs: {
+      category: 'Best Practices',
+      description: 'Prefer importing Jest globals',
+      recommended: false,
+    },
+    messages: {
+      preferJestGlobal:
+        "Jest function \"{{ jestFunction }} is used but not imported from '@jest/globals'",
+    },
+    fixable: 'code',
```
G-Rath, line 16: This rule is not fixable

src/rules/prefer-jest-globals.ts
G-Rath, line 1: Let's rename this to `prefer-importing-jest`

G-Rath (changes requested):

yarn.lock
MadeinFrance: done

src/rules/prefer-jest-globals.ts
MadeinFrance: Type is now a `problem`

src/rules/prefer-jest-globals.ts
MadeinFrance: I think the notion of `globals` should be present like in Jest documentation. What do you think of `prefer-importing-jest-globals`?

G-Rath: My concern is that it's confusing with the existing global functions which are what this rule is wanting you _not_ to use.

Consider if we are to have an opposite rule, what would we call that if we're referring to this using global?

MadeinFrance: renamed

src/rules/prefer-jest-globals.ts
MadeinFrance: yep, refactored using `parseJestFnCall`

src/rules/__tests__/prefer-jest-globals.test.ts
MadeinFrance: Added new alias units tests

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+        if (
+          jestFnCall.head.type === 'import' &&
+          jestGlobalFunctions.includes(jestFnCall.name)
+        ) {
+          importedJestFunctions.push(jestFnCall.name);
+        }
+
+        /* istanbul ignore else */
+        if (jestGlobalFunctions.includes(jestFnCall.name)) {
```
G-Rath, line 40: You don't need to be doing this as it's handled by `parseJestFnCall`  (that's why the "else" isn't covered as the condition is always met 🙂)

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,57 @@
...
+          usedJestFunctions.add(jestFnCall.name);
+        }
+      },
+      'Program:exit'() {
+        usedJestFunctions.forEach(jestFunction => {
+          if (!importedJestFunctions.includes(jestFunction)) {
+            context.report({
```
G-Rath, line 47: We should report when we find a non-imported jest function call both for efficiency and it means we can flag the specific function rather than just the whole file.

Up to you if you want to keep the tracking of already reported functions - I don't think it's that big of a deal if we report on _every_ function call since it'll all get fixed by the same line, but also its not much extra code having the `Set` 🤷

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: espreeParser,
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('prefer-importing-jest-globals', rule, {
```
G-Rath, line 14: We still need more tests - please think more about the _behaviour_ which isn't necessarily enforced by coverage; e.g. what does this rule do if there are multiple usages of the same function call? what if we're already importing one jest function but not another? what if we reassign a jest function to a variable and then used? what if we import a jest-like function from elsewhere?

In this case the answers to those are pretty obvious from looking over the code, but we want tests to reflect those so that it's obvious if they change

G-Rath (changes requested):

MadeinFrance:
> Is it possible to add an autofix for this? I.e. detecting what globals are used, and add either an `import` or `require` of them?
> 
> I think just flagging global usage can be done by not having `jest/globals` in the `env` config of ESLint, so if we add this rule, I think it should provide value above and beyond just marking them.

Hi @SimenB, I've added an auto fixer in the logic 👨🏻‍🔧

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: Condition removed and Istanbul comment removed 👍

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: I need to report at the end `Program:exit()` to know all the Jest functions that needs to be imported & build the complete fix. If I report during the `CallExpression` I might incomplete `import` statement.

See test failing:
```
    - import { describe, test, expect } from '@jest/globals';
    + import { describe } from '@jest/globals';
```

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,66 @@
...
+            jestFunctionsToImport.join(', ');
+
+          context.report({
+            node,
+            messageId: 'preferImportingJestGlobal',
+            data: { jestFunctions: jestFunctionsToImportFormatted },
+            fix(fixer) {
+              return fixer.insertTextBefore(
+                node,
+                `import { ${jestFunctionsToImportFormatted} } from '@jest/globals';\n`,
```
SimenB, line 58: this should be either `import` or `require` depending on what the rest of the file uses. If neither form is used, these should be suggestions for either form

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,66 @@
...
+    return {
+      CallExpression(node) {
+        const jestFnCall = parseJestFnCall(node, context);
+
+        if (!jestFnCall) {
+          return;
+        }
+        if (
+          jestFnCall.head.type === 'import' &&
+          jestGlobalFunctions.includes(jestFnCall.name)
```
G-Rath, line 34: You don't need this, it's handled by `parseJestFnCall`

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        usedJestFunctions.add(jestFnCall.name);
+      },
+      'Program:exit'() {
+        const jestFunctionsToImport = Array.from(usedJestFunctions).filter(
+          jestFunction => !importedJestFunctions.includes(jestFunction),
+        );
+
+        if (jestFunctionsToImport.length > 0) {
+          const node = context.getSourceCode().ast;
```
G-Rath, line 47: We should be reporting on the Jest fn call node, not the whole AST

src/rules/prefer-importing-jest-globals.ts
G-Rath: we also need to ensure this comes after `'use strict'` if present, otherwise it'll break the whole file (we don't need to worry about hashbangs because they're parsed as comments, so this'll already come after that)

src/rules/__tests__/prefer-importing-jest-globals.test.ts
G-Rath: This is getting better - not exhaustive but we're at least missing cases for fixing when:
  - there's a hashbang
  - there's a comment at the start of the file (separate from a hashbang) 
  - there's `'use strict'`
  - there's already an import line for `@jest/globals`;

We also should have some more general tests of different complexities and using chains like `describe.each`

G-Rath (changes requested):

src/rules/__tests__/prefer-importing-jest-globals.test.ts
MadeinFrance: Added extensive unit tests

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: `import` / `require` / `use strict` statements are now properly handled.

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: Thank you, it's now removed 👍

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: Removed, I'm reporting now on the 1st `jestFnCall` node.

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,186 @@
...
+
+        if (!jestFnCall) {
+          return;
+        }
+
+        if (jestFnCall.head.type === 'import') {
+          importedJestFunctions.push(jestFnCall.name);
+        }
+
+        usedJestFunctions.push(jestFnCall);
```
SimenB, line 47: if there are no existing `require` statements, it should be suggestions of either format so the user selects explicitly

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,186 @@
...
+          messageId: 'preferImportingJestGlobal',
+          data: { jestFunctions: jestFunctionsToImportFormatted },
+          fix(fixer) {
+            const sourceCode = context.getSourceCode();
+            const usesImport = sourceCode.ast.body.some(
+              node => node.type === 'ImportDeclaration',
+            );
+            const [firstNode] = sourceCode.ast.body;
+
+            let firstNodeValue;
```
G-Rath, line 75: There's a lot of change that should happen to the code beyond this - it would be good if you could look at:
  * removing repeated/redundant static checks (i.e. doing `node.type === 'something'` consecutively)
  * addressing all coverage ignores - if you can't cover something, that's a sign generally either your code can be simplified or you've missed some syntax; it's not impossible to have code that can't be covered (generally because it's to make TypeScript happy) but consider that we've got 13 coverage ignores in `main` right now, 5 of which are for temporality for addressing deprecated functions in ESLint v8/9 so for now assume if its not covered _you_ have missed something 🙂 
  * replacing casts with proper checks, and especially don't cast to `any` - our [utilities](https://github.com/jest-community/eslint-plugin-jest/tree/main/src/rules/utils) can help with some of that

G-Rath (changes requested):

src/rules/prefer-importing-jest-globals.ts
G-Rath: I wonder if ESLint exposes the `sourceType` option to us through `context`? that'd be a reasonable way of deciding automatically which to use right?

MadeinFrance: I can use the `context.parserOptions.sourceType` to decide automatically the right fix 👍

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,175 @@
+import { TSESLint } from '@typescript-eslint/utils';
+import dedent from 'dedent';
+import rule from '../prefer-importing-jest-globals';
+import { espreeParser } from './test-utils';
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: espreeParser,
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module',
```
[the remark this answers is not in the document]
MadeinFrance: I removed it in order to keep the `default`

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      `,
+      output: dedent`
+        #!/usr/bin/env node
+        const { describe, test, expect } = require('@jest/globals');
+        describe("suite", () => {
+          test("foo");
+          expect(true).toBeDefined();
+        })
+    `,
+      parserOptions: { sourceType: 'module' },
```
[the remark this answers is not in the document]
MadeinFrance: Fixed. I only keep `parserOptions` when necessary.

src/rules/prefer-importing-jest-globals.ts
SimenB: oooh, nice! didn't think about that one 👍

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: espreeParser,
+  parserOptions: {
+    ecmaVersion: 2015,
+  },
+});
+
+ruleTester.run('prefer-importing-jest-globals', rule, {
+  valid: [
```
SimenB, line 14: maybe add a case where method _is_ imported, but not from `@jest/globals`? e.g. `const test = require('my-test');`. But I guess that's probably handled by the `parseJestCall` utility?

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      code: dedent`
+        // with comment above
+        describe("suite", () => {
+          test("foo");
+          expect(true).toBeDefined();
+        })
+      `,
+      output: dedent`
+        // with comment above
+        const { describe, test, expect } = require('@jest/globals');
```
SimenB, line 80: hmm, ideally this should be inserted _above_ the comment, no? I guess it's hard to make this generically correct

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,186 @@
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import type { Literal } from 'estree';
+import { type ParsedJestFnCall, createRule, parseJestFnCall } from './utils';
+
+const createFixerImports = (isModule: boolean, functionsToImport: string[]) => {
+  const allImportsFormatted = functionsToImport.filter(Boolean).join(', ');
```
SimenB, line 6: why is the `filter(Boolean)` needed?

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,186 @@
...
+                      '@jest/globals',
+                ),
+            );
+
+            if (requireNode && requireNode.type === 'VariableDeclaration') {
+              const existingImports =
+                requireNode.declarations[0]?.id.type === 'ObjectPattern'
+                  ? requireNode.declarations[0]?.id.properties?.map(
+                      property => {
+                        /* istanbul ignore else */
```
SimenB, line 146: lots of coverage ignores here - can we add tests which cover these cases? If not - why do we have the uncovered branches to begin with?

SimenB (commented):
thanks for your patience with this PR, @MadeinFrance! 🙏

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: - I'm now using types from `AST_NODE_TYPES`. 
- When searching for existing import specifier what can add in the unit test to cover the `else` and remove the coverage ignore statement in this example?

Any suggestion with an utility function that should return `[ 'test' ]` for `existingImports` 

## Test data:
```
code: dedent`
  import React from 'react';
  import { yourFunction } from './yourFile';
  import something from "something";
  import { test } from '@jest/globals';
  import { xit } from '@jest/globals';
  describe("suite", () => {
    test("foo");
    expect(true).toBeDefined();
  })
`,
output: dedent`
  import React from 'react';
  import { yourFunction } from './yourFile';
  import something from "something";
  import { test, describe, expect } from '@jest/globals';
  import { xit } from '@jest/globals';
  describe("suite", () => {
    test("foo");
    expect(true).toBeDefined();
  })
`,
```

## Program:
```js
const importNode = sourceCode.ast.body.find(
  node =>
    node.type === AST_NODE_TYPES.ImportDeclaration &&
    node.source.value === '@jest/globals',
);

if (
  importNode &&
  importNode.type === AST_NODE_TYPES.ImportDeclaration
) {
  const existingImports = importNode.specifiers.reduce(
    (imports, specifier) => {
      /* istanbul ignore else */
      if (
        specifier.type === AST_NODE_TYPES.ImportSpecifier &&
        specifier.imported?.name
      ) {
        imports.push(specifier.imported.name);
      }

      return imports;
    },
    [] as string[],
  );

  const allImports = [
    ...new Set([...existingImports, ...jestFunctionsToImport]),
  ];
  
  // fixer calls...
  ```

File: src/rules/__tests__/prefer-importing-jest-globals.test.ts (hunk 14 above)
Comment (MadeinFrance) [reply src/rules/__tests__/prefer-importing-jest-globals.test.ts:14]: Added

File: src/rules/__tests__/prefer-importing-jest-globals.test.ts (hunk 15 above)
Comment (MadeinFrance) [reply src/rules/__tests__/prefer-importing-jest-globals.test.ts:86]: It was [requested](https://github.com/jest-community/eslint-plugin-jest/pull/1490#discussion_r1463693631) to have a comment at the top of the file.

To have the `require` on top of the file I need to identify if the 1st node is a comment or shebang. Unfortunately, I have no solution to detect this kind of identifier.

I tried this solution but `firstToken.type` is evaluated as `Identifier` instead of the conditions listed below.

```js
const firstToken = sourceCode.getFirstToken(sourceCode.ast, { includeComments: true });

if (
  firstToken &&
  (firstToken.type === 'LineComment' || firstToken.type === 'BlockComment')
) {
  return fixer.insertTextBeforeRange(
    [0, 0],
    `${createFixerImports(isModule, jestFunctionsToImport)}\n`,
  );
}
```

Maybe you have a solution to identify the type of `firstToken`?

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: See the [current thread](https://github.com/jest-community/eslint-plugin-jest/pull/1490#discussion_r1478408327) about code coverage. Maybe you know some tricks?

.DS_Store
SimenB, line 1: Delete this

SimenB (approved):
LGTM 👍

Would love for @G-Rath to take another look as well

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,182 @@
...
+          node: reportingNode,
+          messageId: 'preferImportingJestGlobal',
+          data: { jestFunctions: jestFunctionsToImportFormatted },
+          fix(fixer) {
+            const sourceCode = context.getSourceCode();
+            const [firstNode] = sourceCode.ast.body;
+
+            const useStrictDirectiveExists =
+              firstNode.type === AST_NODE_TYPES.ExpressionStatement &&
+              (firstNode.expression as Literal).value === 'use strict';
```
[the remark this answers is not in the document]
MadeinFrance: I removed the casting and transformed it to `if` with a comment.

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,182 @@
...
+        const jestFunctionsToImportFormatted = jestFunctionsToImport.join(', ');
+
+        const isModule = context.parserOptions.sourceType === 'module';
+
+        context.report({
+          node: reportingNode,
+          messageId: 'preferImportingJestGlobal',
+          data: { jestFunctions: jestFunctionsToImportFormatted },
+          fix(fixer) {
+            const sourceCode = context.getSourceCode();
```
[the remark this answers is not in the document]
MadeinFrance: Now using the local function from `utils`: `getSourceCode(context)`

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,182 @@
...
+              node =>
+                node.type === AST_NODE_TYPES.ImportDeclaration &&
+                node.source.value === '@jest/globals',
+            );
+
+            if (
+              importNode &&
+              importNode.type === AST_NODE_TYPES.ImportDeclaration
+            ) {
+              const existingImports = importNode.specifiers.reduce(
```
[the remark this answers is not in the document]
MadeinFrance: Type added.

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,189 @@
...
+                node.source.value === '@jest/globals',
+            );
+
+            if (
+              importNode &&
+              importNode.type === AST_NODE_TYPES.ImportDeclaration
+            ) {
+              const existingImports = importNode.specifiers.reduce<string[]>(
+                (imports, specifier) => {
+                  /* istanbul ignore else */
```
G-Rath, line 100: this can be covered by e.g. a default import:

```
import describe from '@jest/globals';
describe("suite", () => {
  test("foo");
  expect(true).toBeDefined();
})
```

General tip for helping track down coverage: look at what types something can be.

In this case, `specifier` is of type `ImportClause` which is:

```
export declare type ImportClause = ImportDefaultSpecifier | ImportNamespaceSpecifier | ImportSpecifier;
```

So you want to play around with those three types of things to get coverage - and it seems in this case just having a default import `@jest/globals` is enough

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,189 @@
...
+                  if (
+                    specifier.type === AST_NODE_TYPES.ImportSpecifier &&
+                    specifier.imported?.name
+                  ) {
+                    imports.push(specifier.imported.name);
+                  }
+
+                  return imports;
+                },
+                [] as string[],
```
G-Rath, line 110: This is not needed

```suggestion
                [],
```

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,189 @@
...
+                      'Literal' &&
+                    (declaration.init as any).arguments?.[0]?.value ===
+                      '@jest/globals',
+                ),
+            );
+
+            if (
+              requireNode &&
+              requireNode.type === AST_NODE_TYPES.VariableDeclaration
+            ) {
```
G-Rath, line 141:
```suggestion
            if (requireNode?.type === AST_NODE_TYPES.VariableDeclaration) {
```

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,189 @@
...
+                node.type === AST_NODE_TYPES.VariableDeclaration &&
+                node.declarations.some(
+                  declaration =>
+                    declaration.init &&
+                    (declaration.init as any).callee &&
+                    (declaration.init as any).callee.name === 'require' &&
+                    (declaration.init as any).arguments?.[0]?.type ===
+                      'Literal' &&
+                    (declaration.init as any).arguments?.[0]?.value ===
+                      '@jest/globals',
```
G-Rath, line 134: Don't cast to `any` - even if you have to `istanbul ignore`, properly narrow these down through checks; this is a `CallExpression`, which you want to know if it has a single argument and if that argument is a literal with the value `@jest/globals`.

We have utilities to help you with this - checkout `hasOnlyOneArgument` and `isStringNode`; that should also lead you to realise there are holes you need to explore, like what if it's a template literal (which is a different type but covered by using `isStringNode`), template literal with interpolation, a variable, and so on.

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,225 @@
...
+import { espreeParser } from './test-utils';
+
+const ruleTester = new TSESLint.RuleTester({
+  parser: espreeParser,
+  parserOptions: {
+    ecmaVersion: 2015,
+  },
+});
+
+ruleTester.run('prefer-importing-jest-globals', rule, {
```
G-Rath, line 13: Please include the `line` in all your invalid cases, and you're missing more tests:
  * what if we're importing from default
  * what if we're using a name from not `@jest/globals` (i.e. `const describe = require('./mine')`)
  * what if we have our require at the end of the file (i.e. `const it = require('@jest/globals')` - while silly, that is valid if everything is wrapped in a `describe`)
  * what if template literals are used, and what if those have interopation?

A number of those will come out from my comments down below :)

(also please chuck in some more `import` based tests to mirror the `require` ones, since those involve a very different AST)

src/rules/prefer-importing-jest-globals.ts
G-Rath: Also this is the much longer codepath - can we flip and do `!==` here instead?

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,189 @@
...
+                      ) {
+                        return property.key.name;
+                      }
+                      /* istanbul ignore else */
+                      if (
+                        property.type === AST_NODE_TYPES.Property &&
+                        property.key.type === AST_NODE_TYPES.Literal
+                      ) {
+                        return property.key.value;
+                      }
```
G-Rath, line 158:
```suggestion
                      if (
                        property.type === AST_NODE_TYPES.Property &&
                        isSupportedAccessor(property.key)
                      ) {
                        return getAccessorValue(property.key);
                      }
```

G-Rath: Also you can cover this with [`[() => {}]`](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/utils/__tests__/parseJestFnCall.test.ts#L609-L615)

G-Rath (changes requested):

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: removed

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: changed

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: Cool, I could remove the cast to `any`. But I have type issue with the helper function. Could you come up with a code suggestion please?

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: Never seen this syntax but thanks, it covers the program 🙏

src/rules/__tests__/prefer-importing-jest-globals.test.ts
MadeinFrance: I did reflect more tests for `import` and added the `line` for the invalid cases.

The default import is not working as expected, see the failing test case:

```js
{
  code: dedent`
    const describe = require('@jest/globals');
    describe("suite", () => {
      test("foo");
      expect(true).toBeDefined();
    })
  `,
  output: dedent`
    const { describe, test, expect } = require('@jest/globals');
    describe("suite", () => {
      test("foo");
      expect(true).toBeDefined();
    })
`,
  errors: [
    {
      endColumn: 7,
      column: 3,
      line: 3,
      messageId: 'preferImportingJestGlobal',
    } as const,
  ],
},
```

<hr>

```
Received:
      "const { test, expect } = require('@jest/globals');
```

What would be your recommendation to solve this? 

I thought of trying to use `VariableDeclarator`, `Identifier` but it's hard to grasp.

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,451 @@
...
+    // },
+    // {
+    //   code: dedent`
+    //     const { test } = require('my-test-library');
+    //     test("foo");
+    //   `,
+    // },
+  ],
+  invalid: [
+    ...importImports,
```
G-Rath, line 372: I don't see any advantage to this and it gives you a worse typing & IDE experience since you'll get errors here rather than autocompletion when writing more cases (e.g. that's why you need to use `as const`) - can we just inline these?

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,451 @@
...
+        column: 1,
+        line: 1,
+        messageId: 'preferImportingJestGlobal',
+      } as const,
+    ],
+  },
+];
+
+ruleTester.run('prefer-importing-jest-globals', rule, {
+  valid: [
```
G-Rath, line 331: What's happened with all our valid cases? 😅

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -0,0 +1,214 @@
...
+                      AST_NODE_TYPES.Identifier &&
+                    declaration.init.callee.name === 'require' &&
+                    declaration.init.arguments[0]?.type === 'Literal' &&
+                    declaration.init.arguments[0]?.value === '@jest/globals' &&
+                    (declaration.id.type === AST_NODE_TYPES.Identifier ||
+                      declaration.id.type === AST_NODE_TYPES.ObjectPattern),
+                ),
+            );
+
+            if (requireNode?.type === AST_NODE_TYPES.VariableDeclaration) {
```
G-Rath, line 167: I'd prefer to do an early return for the other branch since its a lot shorter

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -0,0 +1,451 @@
...
+      `,
+      output: dedent`
+        'use strict';
+        const { describe, test, expect } = require('@jest/globals');
+        describe("suite", () => {
+          test("foo");
+          expect(true).toBeDefined();
+        })
+    `,
+      errors: [
```
G-Rath, line 441: nit: ideally these should align, but I'm happy to take care of that in post:

```suggestion
        })
      `,
      errors: [
```

G-Rath (changes requested):
awesome stuff getting complete coverage - I think we're almost there, just need to restore our valid test cases; I'm happy to take care of any further cleanup in post :)

src/rules/__tests__/prefer-importing-jest-globals.test.ts
MadeinFrance: Removed, it was easier to deal with specific unit tests.

src/rules/prefer-importing-jest-globals.ts
MadeinFrance: modified

src/rules/__tests__/prefer-importing-jest-globals.test.ts
MadeinFrance: Prettier took care of it:
image

src/rules/__tests__/prefer-importing-jest-globals.test.ts
MadeinFrance: Oops, I've added them again.

MadeinFrance:
@G-Rath can we move forward to Main with this PR? Most of the clean up is done.

G-Rath:
@MadeinFrance I'm waiting to see what's happening with our upcoming major whichll probably get released this week.

This PR looks good to me so you can consider it off your plate - I'll handle landing it and doing any revising that might be needed :)

G-Rath:
Thanks again @MadeinFrance! I have done some minor refactoring that I recommend checking out if you're interested but don't feel bad about any of it - you did the grunt work which I've just built on and I've been working with these rules for years so never expected a first-time contributor to have gotten my changes on their first rule.

The fixer is still a bit rough but I'm happy to improve that in follow-up PRs since you've already gone through a lot here and I think the roughness is just with very unlikely edge cases e.g.
  - `import jest from '@jest/globals'` should give `import jest, { ... }` rather than `import { jest, ... }`
  - we should retain template literals if they're used in the `require`
  - we shouldn't end up with the next line being dedented after adding the import
  - ideally we could probably support `settings.jest.globalAliases` too (this is super low priority though)

MadeinFrance:
@G-Rath I'm thrilled to see this landing on the release line. Thank you for your review and patience 🙏
Seeing how this PR has grown since the initial commit is amazing.

I will use this in our corporate repos and improve the edge cases if required.

PR #1504, opened 2024-02-15
refactor: use types from `@typescript-eslint/utils` for snapshot processor

G-Rath:
Mostly useful because it means we're more compatible with the `Plugin` type which I'm using in another branch


SimenB (approved):
a test is failing, but the change itself is 👍

PR #1506, opened 2024-02-16
refactor: make config rules static rather than based off rule metadata

G-Rath:
Rule metadata is really typed for how `@typescript-eslint` use it and their recent majors have changed exports so we can't actually argument their interfaces anymore so this switches us to just using a static object.

I've _not_ done anything fancy like having dedicated files in `configs/` because I don't think there's a big advantage at this point and frankly I just want to push through to the ESLint v9 / `@typescript-eslint` v<whatever> (I think 7..?) so will revisit restructuring later once the dust has settled.

I've also not removed the actual properties from the rules as I think that technically could be a breaking change? but I have at least tested without them locally to ensure all the scripts and workflow still work.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,14 +64,34 @@ const rules = Object.fromEntries(
     .map(rule => [rule, importDefault(join(rulesDir, rule)) as RuleModule]),
 );
 
-const recommendedRules = Object.fromEntries(
-  Object.entries(rules)
-    .filter(([, rule]) => rule.meta.docs.recommended)
-    .map(([name, rule]) => [
-      `jest/${name}`,
-      rule.meta.docs.recommended as TSESLint.Linter.RuleLevel,
-    ]),
-);
+const recommendedRules = {
+  'jest/expect-expect': 'warn',
+  'jest/no-alias-methods': 'error',
+  'jest/no-commented-out-tests': 'warn',
+  'jest/no-conditional-expect': 'error',
+  'jest/no-deprecated-functions': 'error',
+  'jest/no-disabled-tests': 'warn',
+  'jest/no-done-callback': 'error',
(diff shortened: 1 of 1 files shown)


src/__tests__/rules.test.ts
```diff
@@ -48,19 +48,38 @@ describe('rules', () => {
   });
 
   it('should export configs that refer to actual rules', () => {
+    const expectJestPlugin = expect.objectContaining({
+      meta: {
+        name: 'eslint-plugin-jest',
+        version: '27.8.0',
```
SimenB, line 54: this should not be hard coded as it'll break for every new release

src/index.ts
```diff
@@ -64,46 +64,94 @@ const rules = Object.fromEntries(
...
+  'jest/valid-expect-in-promise': 'error',
+  'jest/valid-title': 'error',
+} satisfies Record<string, TSESLint.Linter.RuleLevel>;
+
+const styleRules = {
+  'jest/no-alias-methods': 'warn',
+  'jest/prefer-to-be': 'error',
+  'jest/prefer-to-contain': 'error',
+  'jest/prefer-to-have-length': 'error',
+} satisfies Record<string, TSESLint.Linter.RuleLevel>;
```
SimenB, line 94: how can we ensure this is kept in sync with docs and the meta prop?

G-Rath: The docs are generated based on this, and the meta props are going away - I've not removed them yet because the types currently still require them and that it might technically be breaking, though I could axe `category` in this PR if you like

src/__tests__/rules.test.ts
G-Rath: oh yeah good point 🙈

PR #1401, opened 2023-07-13
feat: upgrade `@typescript-eslint/utils` to v6

SimenB:
There are so many breaking changes in v6 I don't think we'll be able to properly support v5 and v6 at the same time...

I started here and just ended up frustrated 😛 @G-Rath I dunno if you wanna give it a whack?

PR #1508, opened 2024-02-16
feat: upgrade `@typescript-eslint/utils` to v6

G-Rath:
Really the changes needed were just type and testing based, though I'm not sure if we can support v5 without a breaking change anyway because we'd have to switch to requiring it as a peer dependency which is arguably still breaking?

v6 dropped support for Node v14 which we still support.

Closes #1401

(this is meant to be the min. required for the upgrade - I'll do other stuff like updating the linting plugins in another PR)

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     ]
   },
   "dependencies": {
-    "@typescript-eslint/utils": "^5.10.0"
+    "@typescript-eslint/utils": "^6.0.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
--- a/src/rules/__tests__/unbound-method.test.ts
+++ b/src/rules/__tests__/unbound-method.test.ts
@@ -1,5 +1,5 @@
 import path from 'path';
-import { ESLintUtils, type TSESLint } from '@typescript-eslint/utils';
+import { TSESLint } from '@typescript-eslint/utils';
 import dedent from 'dedent';
 import type { MessageIds, Options } from '../unbound-method';
 
@@ -9,14 +9,24 @@ function getFixturesRootDir(): string {
 
 const rootPath = getFixturesRootDir();
 
-const ruleTester = new ESLintUtils.RuleTester({
-  parser: '@typescript-eslint/parser',
+const ruleTester = new TSESLint.RuleTester({
+  parser: require.resolve('@typescript-eslint/parser'),
   parserOptions: {
     sourceType: 'module',
     tsconfigRootDir: rootPath,
     project: './tsconfig.json',
   },
 });
+
+const fixtureFilename = path.join(rootPath, 'file.ts');
+
+const addFixtureFilename = (
+  code: TSESLint.ValidTestCase<unknown[]> | string,
+): TSESLint.ValidTestCase<unknown[]> => {
+  const test = typeof code === 'string' ? { code } : code;
+
+  return { filename: fixtureFilename, ...test };
+};
 
 const ConsoleClassAndVariableCode = dedent`
   class Console {
@@ -164,8 +174,8 @@ describe('error handling', () => {
   });
 
   describe('when @typescript-eslint/eslint-plugin is not available', () => {
-    const ruleTester = new ESLintUtils.RuleTester({
-      parser: '@typescript-eslint/parser',
+    const ruleTester = new TSESLint.RuleTester({
+      parser: require.resolve('@typescript-eslint/parser'),
       parserOptions: {
         sourceType: 'module',
         tsconfigRootDir: rootPath,
@@ -185,7 +195,7 @@ describe('error handling', () => {
 });
 
 ruleTester.run('unbound-method jest edition', requireRule(false), {
-  valid: validTestCases,
+  valid: validTestCases.map(addFixtureFilename),
   invalid: invalidTestCases,
 });
 
@@ -456,7 +466,7 @@ class OtherClass extends BaseClass {
 const oc = new OtherClass();
 oc.superLogThis();
     `,
-  ],
+  ].map(addFixtureFilename),
   invalid: [
     {
       code: `
--- a/src/rules/__tests__/unbound-method.test.ts
+++ b/src/rules/__tests__/unbound-method.test.ts
@@ -20,12 +20,22 @@ const ruleTester = new TSESLint.RuleTester({
 
 const fixtureFilename = path.join(rootPath, 'file.ts');
 
-const addFixtureFilename = (
-  code: TSESLint.ValidTestCase<unknown[]> | string,
-): TSESLint.ValidTestCase<unknown[]> => {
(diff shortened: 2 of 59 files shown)


SimenB:
@G-Rath what's the status here? 🙂 Would be nice to land this (and I'm happy to drop support for v5 as a breaking change)

G-Rath:
@SimenB that's basically the status - I think we just should do a new major and move on though I figured we might want to wait until ESLint v9 is out but I'm happy to do more majors smaller majors if you are.

Let me know what your preference is for how many majors and what we drop

SimenB:
We can do a quick follow-up major if needed - I don't think it's much churn for folks to update an eslint plugin (unless all the rules and their options change name or something 😅)

G-Rath:
Cool - do you just want to do v5 for now or jump straight to v6 (which'd be more useful but I think means dropping a version of Node that Jest hasn't dropped yet)

G-Rath:
awesome the part about not minding Jest being out of sync was the main part - with all the different versions around I can't actually remember which is the latest but I'll pick this back up later in the week and get some progress made.

I'll also look at landing #1506 since that should help and get `next` cleaned up, and we can go from there 🎉

bradzacher:
> for Node v14 which we still support

Question: Why do you still support Node 14?
It's almost an entire year out of support - and node 16 similarly is 6 months out of support.
Is it because you're aiming to support the same versions as `jest` itself?

dubzzz:
> 6 or jump straight to 7, I assume?

Jumping to the conversation, but I think that as a user of the set of rules, it might be helpful to have at least one version with v6 (even just a beta) then one with v7. It might simplify the bump for users of eslint-plugin-jest: they will only have to move first to v6 then to v7.

SimenB:
Yeah, agreed 👍 unless it's a bunch of extra work, but I don't think it would be?

bradzacher:
v7 may be untenable for you guys to jump straight to.
The main purpose of v7 was to clamp down the dependencies to make life easier for eslint v9.
So v7 requires eslint v8.56+ and node 18+

Because it drops old eslint versions it dropped some APIs that we exported to help with the v8/v9 interop. Idk if you guys had switched to those yet.

G-Rath:
I need to recheck my notes to remember all the details (and there might not be that much) but from memory while v6 will be an improvement for us it still leaves us with a few thorns e.g. https://github.com/typescript-eslint/typescript-eslint/pull/8485 is v7 only.

But I'm happy to do multiple majors either way 🤷

G-Rath:
> Jumping to the conversation, but I think that as a user of the set of rules, it might be helpful to have at least one version with v6 (even just a beta) then one with v7. It might simplify the bump for users of eslint-plugin-jest: they will only have to move first to v6 then to v7.

To be clear, the main impact this will have on users is dropping support for Node v14 - our dependency on `@typescript-eslint/utils` is a direct one so it doesn't impact users directly beyond the Node version; we also have a _peer_ dependency on `@typescript-eslint/eslint-plugin` which does directly impact users but that is just for importing `unbound-method` which we do a very light wrapper around so we can easily retain support for v5.

So long as you've already upgraded from Node v14, you should be fine.

(but again, I have no issue doing multiple majors - this is just often a source of confusion😄)

G-Rath:
I've done the actual Node v14 dropping in #1508 so that it'll get a dedicated changelog entry - once that is merge, this'll pass CI

SimenB:
rebase and it should be good?

G-Rath:
Yup, I'm just doing a couple of things at once while I wait for CIs in different projects to finish 😛

package.json
```diff
@@ -95,7 +95,7 @@
     ]
   },
   "dependencies": {
-    "@typescript-eslint/utils": "^5.10.0"
+    "@typescript-eslint/utils": "^6.0.0"
```
SimenB, line 98: image

as well?

src/rules/__tests__/unbound-method.test.ts
```diff
@@ -225,11 +248,14 @@ function addContainsMethodsClassInvalid(
 }
 
 ruleTester.run('unbound-method', requireRule(false), {
-  valid: [
+  valid: withFixtureFilename([
     'Promise.resolve().then(console.log);',
     "['1', '2', '3'].map(Number.parseInt);",
     '[5.2, 7.1, 3.6].map(Math.floor);',
     'const x = console.log;',
+    ...(parseInt(rawTypeScriptESLintPluginVersion.split('.')[0], 10) >= 6
```
SimenB, line 256: won't this always be true?

SimenB: (also, we should use `semver` module for this)

package.json
G-Rath: we're still supporting `@typescript-eslint/eslint-plugin` v5 and don't use a lot of the actual rules so I don't think there's a strong advantage - I'll double check that but I don't think it should block landing this right now

src/rules/__tests__/unbound-method.test.ts
G-Rath: No because we still support and test against `@typescript-eslint/eslint-plugin` v5

package.json
SimenB: I think we can drop it - no reason too keep support when bumping version (when we're landing as major regardless)

G-Rath: ok but can we do that in a PR after this so it gets a dedicated changelog entry?

SimenB: i.e. I think it makes sense to add support for further majors as a minor, but if we're doing a major release regardless, I don't think it's super valuable

SimenB: for sure!

SimenB: (sorry, I had the tab open and GH isn't super good with live updates 😅 let's blame rails!)

src/rules/__tests__/unbound-method.test.ts
G-Rath: I'll not replace this with `semver` since it'll just go away right after this PR when we drop support for v5

package.json
bradzacher: Suggestion: update this to the last v6 version.
That way you're enforcing that package managers always give you the latest version, rather than trying to resolve to some random one with fewer bug fixes.

G-Rath: I tend to favor not doing that to enable better deduplication downstream, given we consider this a very stable dependency

src/rules/prefer-lowercase-title.ts
```diff
@@ -70,11 +68,14 @@ export default createRule<
           ignore: {
             type: 'array',
             items: {
+              type: 'string',
+              // for some reason TypeScript thinks this _must_ be a read-only
```
bradzacher, line 72: We probably need to update our parameter types to be `const` parameters to avoid weirdness like this.

SimenB (approved):
👍

PR #1514, opened 2024-02-19
Fix typo in README.md

danielrentz:
Flat config: Add missing "rules" property of flat/recommended entry in "rules" object


SimenB (approved):
Sweet, thanks!

PR #1528, opened 2024-03-21
feat: remove `no-if` rule

G-Rath:
BREAKING CHANGE


SimenB:
(not against it, but why? missing motivation/issue I think 😅 )

G-Rath:
Do you mean for the changelog or that you've forgotten the history? (I don't blame you if so😅)

We replaced it with `no-conditional-in-test` which does the same but with more conditions, but we kept forgetting to actually remove this rule in our follow-up majors...

#1049 & #1027

SimenB:
haha, both! 😅 The `BREAKING CHANGE` part of the commit message should include motivation/migration path.

G-Rath:
🤦 it's been so long I forgot that

SimenB (approved):

PR #1531, opened 2024-03-23
Next major

### LINKED ISSUE #292: Snap processor makes enforcing other rules on .snap files impossible
Adding the processors to .snap files, and those processors being so explicitly about no-large-snapshots, means I cannot use max-lines, for example, on snapshots to enforce that we are not generating too many snapshots for a given suite.

I appreciate the value of limiting the rules applied to .snap files specifically, but I think it ought to be configurable.

G-Rath:
The only major changes left I can think of are changes to config and rule defaults (of which I can't think of any we actually wanted to do?) and maybe changing our snapshot processor which I'm not actually sure we need anymore? (Will discuss on #292)

Resolves #292


G-Rath:
I think this is good to go - there's a few fixes that don't require a new major but I didn't land them in `main` to avoid a rebase, but given we have to do that now anyway we could cherry-pick? I'm not fussed 🤷

SimenB (approved):

SimenB (approved):
great job!

SimenB:
Incredible work on this release @G-Rath, thank you so much!

PR #1532, opened 2024-03-23
feat: remove `snapshot` processor and `flat/snapshot` config

### LINKED ISSUE #292: Snap processor makes enforcing other rules on .snap files impossible

G-Rath:
I'm pretty sure this isn't actually needed especially in ESLint v9 - we can always add it back (or something different) in a minor version

Resolves #292

BREAKING CHANGE: removed `snapshot` processor and `flat/snapshot` config


SimenB (approved):
makes sense to me 👍 as you say, easy enough to add back if it's needed.

Include some "as it's not needed" in the deprecation message? might encourage people to say "it actually is"?

PR #1534, opened 2024-03-23
ci: run against upcoming next major ESLint version

G-Rath:
~Hopefully this'll just work...~ it did not, but now it does!

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,4 +1,5 @@
 import type { Config } from '@jest/types';
+import { version as typescriptESLintPluginVersion } from '@typescript-eslint/eslint-plugin/package.json';
 import { version as eslintVersion } from 'eslint/package.json';
 import * as semver from 'semver';
 
@@ -24,12 +25,14 @@ const config = {
         '<rootDir>/src/rules/__tests__/fixtures/*',
         '<rootDir>/src/rules/__tests__/test-utils.ts',
       ],
+      coveragePathIgnorePatterns: ['/node_modules/'],
     },
     {
       displayName: 'lint',
       runner: 'jest-runner-eslint',
       testMatch: ['<rootDir>/**/*.{js,ts}'],
       testPathIgnorePatterns: ['<rootDir>/lib/.*'],
+      coveragePathIgnorePatterns: ['/node_modules/'],
     },
   ],
 } satisfies Config.InitialOptions;
@@ -38,6 +41,18 @@ if (semver.major(eslintVersion) >= 9) {
   config.projects = config.projects.filter(
     ({ displayName }) => displayName !== 'lint',
   );
+
+  // jest/unbound-method doesn't work when using @typescript-eslint v6
+  if (semver.major(typescriptESLintPluginVersion) === 6) {
+    for (const project of config.projects) {
+      project.testPathIgnorePatterns.push(
+        '<rootDir>/src/rules/__tests__/unbound-method.test.ts',
+      );
+      project.coveragePathIgnorePatterns.push(
+        '<rootDir>/src/rules/unbound-method.ts',
+      );
+    }
+  }
 }
 
 export default config;
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,6 +1,6 @@
-import type { Config } from '@jest/types';
 import { version as typescriptESLintPluginVersion } from '@typescript-eslint/eslint-plugin/package.json';
 import { version as eslintVersion } from 'eslint/package.json';
+import type { Config } from 'jest';
 import * as semver from 'semver';
 
 const config = {
@@ -35,7 +35,7 @@ const config = {
       coveragePathIgnorePatterns: ['/node_modules/'],
     },
   ],
-} satisfies Config.InitialOptions;
+} satisfies Config;
 
 if (semver.major(eslintVersion) >= 9) {
   config.projects = config.projects.filter(
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,4 +1,6 @@
 import type { Config } from '@jest/types';
+import { version as eslintVersion } from 'eslint/package.json';
+import * as semver from 'semver';
 
 const config = {
   clearMocks: true,
@@ -31,5 +33,11 @@ const config = {
     },
   ],
 } satisfies Config.InitialOptions;
+
+if (semver.major(eslintVersion) >= 9) {
+  config.projects = config.projects.filter(
+    ({ displayName }) => displayName !== 'lint',
+  );
+}
 
 export default config;
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,0 +1,35 @@
+import type { Config } from '@jest/types';
+
+const config = {
+  clearMocks: true,
+  restoreMocks: true,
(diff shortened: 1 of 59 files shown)


G-Rath:
It seems like ESLint v9 somehow has different globals resulting in `instanceof RegExp` no longer working like in https://github.com/jestjs/jest/issues/2549 🤔

G-Rath:
Seems that `unbound-method` doesn't like being run when `eslint-plugin` v6 is present as I think it ends up getting a bad version of `@typescript-eslint/utils`?

I've confirmed this is only a development problem - everything works fine on a standard project that's using `eslint-plugin-jest` and `@typescript-eslint` v7 side by side; I'm not sure if we can easily address this in our CI though without just skipping the entire v6 test for `next` :/

package.json
```diff
@@ -104,6 +74,7 @@
     "@babel/preset-typescript": "^7.3.3",
     "@commitlint/cli": "^17.0.3",
     "@commitlint/config-conventional": "^17.0.3",
+    "@jest/types": "^29.6.3",
```
SimenB, line 77:
```suggestion
```

jest.config.ts
```diff
@@ -0,0 +1,58 @@
+import type { Config } from '@jest/types';
```
SimenB, line 1:
```suggestion
import type { Config } from 'jest';
```

https://jestjs.io/docs/configuration

jest.config.ts
```diff
@@ -0,0 +1,58 @@
...
+    },
+    {
+      displayName: 'lint',
+      runner: 'jest-runner-eslint',
+      testMatch: ['<rootDir>/**/*.{js,ts}'],
+      testPathIgnorePatterns: ['<rootDir>/lib/.*'],
+      coveragePathIgnorePatterns: ['/node_modules/'],
+    },
+  ],
+} satisfies Config.InitialOptions;
```
SimenB, line 38:
```suggestion
} satisfies Config;
```

why `satisfies` instead of assigning?

src/rules/__tests__/no-deprecated-functions.test.ts
```diff
@@ -154,6 +155,20 @@ describe('the rule', () => {
       expect(() => {
         const linter = new TSESLint.Linter();
 
+        /* istanbul ignore if */
+        if (usingFlatConfig()) {
```
SimenB, line 159: why is this needed?

src/rules/__tests__/test-utils.ts
```diff
@@ -1,6 +1,87 @@
...
+import { TSESLint } from '@typescript-eslint/utils';
+import { version as eslintVersion } from 'eslint/package.json';
+import * as semver from 'semver';
 
 const require = createRequire(__filename);
 const eslintRequire = createRequire(require.resolve('eslint'));
 
 export const espreeParser = eslintRequire.resolve('espree');
+
+export const usingFlatConfig = () => semver.major(eslintVersion) >= 9;
```
SimenB, line 11: why does this need to be a function? rather than just

```suggestion
export const usingFlatConfig = semver.major(eslintVersion) >= 9;
```

src/rules/__tests__/test-utils.ts
```diff
@@ -1,6 +1,87 @@
...
+
+export class FlatCompatRuleTester extends TSESLint.RuleTester {
+  public constructor(testerConfig?: TSESLint.RuleTesterConfig) {
+    super(FlatCompatRuleTester._flatCompat(testerConfig));
```
SimenB, line 15: is it ok to use this private thing?

src/rules/__tests__/test-utils.ts
```diff
@@ -1,6 +1,87 @@
...
+  >(config: T): T {
+    if (!config || !usingFlatConfig() || typeof config === 'string') {
+      return config;
+    }
+
+    const obj = { languageOptions: { parserOptions: {} } };
+
+    for (const [key, value] of Object.entries(config)) {
+      if (key === 'parser') {
+        // @ts-expect-error this is expected
```
SimenB, line 49: what is expected? (applies to all)

src/rules/__tests__/test-utils.ts
```diff
@@ -1,6 +1,87 @@
...
+    }
+
+    return obj as unknown as T;
+
+    // return {
+    //   languageOptions: {
+    //     parser: require(config.parser),
+    //     ...config.parserOptions,
+    //   },
+    // } as unknown as T;
```
SimenB, line 85: remove?

SimenB (changes requested):
exciting!

jest.config.ts
G-Rath: because assigning means it _is_ `Config` - so all properties marked as required are such, and optional ones are that outside of the assignment; so TypeScript won't let us do things on those properties later down without existence checks - sadly it's not perfect: we've got to define properties we want to use since the type information is not retained after the assignment, which is why we've got `coveragePathIgnorePatterns` explicitly defined with its default values.

Having said that, I've not checked out `Config.InitialOptions` so that might change things

G-Rath: Ok so `jest.Config` is just a re-export of `Config.InitialOptions` - so yeah, we could do either or; I think using `satisfies` is slightly nicer though.

To make it a bit clearer:

image

that's what happens if you use assignment, and it's because to TypeScript `config` is type `Config` rather than the actual object we created.

src/rules/__tests__/no-deprecated-functions.test.ts
G-Rath: Because we're effectively running ESLint here per usual - so when using ESLint v9 we have to use flatconfig and when using pre-v9 we have to use legacy config.

While `linter.verify` supports being passed both types of configs, since we have call `defineRule` in pre-v9 (which errors in v9) this felt the cleanest way to address it (though either way we'll always need `usingFlatConfig`.

src/rules/__tests__/test-utils.ts
G-Rath: it's our private thing so yes :)

src/rules/__tests__/test-utils.ts
G-Rath: that it errors, because it's very annoying to type especially given we're still a major version of `@typescript-eslint` behind (so the flat config types are not necessarily correct and properly supported in the utils we're using).

This is effectively a kludge to tie us over the next couple of months until v9 actually comes out, like some of the other casts and `@ts-expect-errors` we've got - I'll try to stick a general explainer comment in but every time I tried it didn't come out that subscint and I'm not expecting it to require much attention from anyone other than me 😅

src/rules/__tests__/test-utils.ts
G-Rath: oh yeah, good point!

src/rules/__tests__/test-utils.ts
G-Rath: I've refactored this with types which let me remove most of the errors - ironically the last one I could get rid of by using dedicated `if` statements but I don't think it's worth it

SimenB (approved):

PR #1541, opened 2024-03-29
refactor: use `flatMap` instead of `reduce`

G-Rath:
I realised this while doing #1540 - shaves off a few lines which is nice


SimenB:
should we enable https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-flat-map.md?

G-Rath:
I don't think it's worth it given we don't already use `unicorn` and this is literally all the ones it could flag 🤷

(+ we've been pretty stable for a while so I don't expect a lot of new code)

PR #1547, opened 2024-04-06
fix: allow ESLint 9 as peer dependency

SimenB:


G-Rath:
Beat me to it 😁

package.json
```diff
@@ -86,7 +86,7 @@
     "babel-jest": "^29.0.0",
     "babel-plugin-replace-ts-export-assignment": "^0.0.2",
     "dedent": "^1.5.0",
-    "eslint": "^7.0.0 || ^8.0.0",
+    "eslint": "^7.0.0 || ^8.0.0 || ^9.0.0",
```
G-Rath, line 89: @SimenB is there some Yarn trick I don't know about to actually select the version to use locally without a git change?

I'm finding that v9 is being installed, which completely breaks my IDE and tests - alternatively, could we remove v9 from this list? CI was running fine with installing `next` without this so I don't know quite why we need it 😅

anthony0030:
I am getting this warning:
```LOG
warning "eslint-plugin-jest > @typescript-eslint/utils@6.21.0" has incorrect peer dependency "eslint@^7.0.0 || ^8.0.0".
```
Is it related?

SimenB:
@anthony0030 https://github.com/typescript-eslint/typescript-eslint/issues/8211

(we will probably need to update to v7 whenever that is actually fixed)

Things should work fine even with the warning, tho.

anthony0030:
@SimenB thankyou for the info 😃

package.json
G-Rath: Have opened https://github.com/jest-community/eslint-plugin-jest/pull/1569

PR #1555, opened 2024-04-12
chore(deps): update typescript-eslint monorepo to v7 (major)

lluiscab:
Based on #1397 

Solves tsc errors in `src/index.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     ]
   },
   "dependencies": {
-    "@typescript-eslint/utils": "^6.0.0"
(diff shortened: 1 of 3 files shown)


G-Rath:
This is a breaking change since it drops versions of Node we still support, and there's currently no need to upgrade

lluiscab:
I was unaware of the dropped support for older NodeJS versions. I saw the pending failed PR from renovate and figured that was he only thing blocking the update.

I'd like the update mainly to finally remove `@typescript-eslint/utils@^6.0.0` from my project and use exclusively v7

PR #1563, opened 2024-04-24
feat: import formatting rules from `eslint-plugin-jest-formatting`

### LINKED ISSUE #17: Padding between it and describe blocks.
I have been looking for a rule to enforce padding around all `it` and `describe` blocks. Is this the correct place for that? Maybe `jest/padding-it-blocks` and `jest/padding-describe-blocks` rules?

```
it('My awesome test.', () => {
  expect(true).toBe(true)
})
// Enforcing this line of padding between the it blocks below this comment...

it('My other awesome test.', () => {
  expect(false).toBe(false)
})
```

### LINKED ISSUE #113: [new rule] strict formatting for describe and test blocks
We would really like to enforce some strict formatting for the various parts of jest test syntax. I'm not sure if these would be general enough to be considered useful, but they seem like a nice way to help ensure test code is nicely structured. An example of my teams desired syntax is:
```
  let varName

  beforeEach(() => {
    let varName = 'foo'
  })

  it('does something cool', () => {
    const result = callAFuncWith(varName)
    result.wow()

    expect(result).toBeAwesome()
  })
})
```
Seperate rules could enforce:
- a blank line after describe level lets
- a blank line after a beforeEach
- a blank line between an expect and test setup statements

Another option would be a new 'eslint-jest-formatting' package.

### LINKED ISSUE #1251: [new rule] prefer blank line between tests or blocks
Add a style rule to suggest adding a blank line between tests or blocks.

```ts
// INCORRECT
beforeAll(() => {
  /* ... */
}) // missing space after
describe('function a', () => {
  beforeEach(() => {
    /* ... */
  }) // missing space after
  it('returns one', () => {
    /* ... */
  }) // missing space after
  it('returns two', () => {
    /* ... */
  })
}) // missing space after
describe('function b', () => {
  it('returns three', () => {
    /* ... */
  }) // missing space after
  it('returns four', () => {
    /* ... */
  })
})

// CORRECT
beforeAll(() => {
  /* ... */
}) // space after

describe('function c', () => {
  beforeEach(() => {
    /* ... */
  }) // space after

  it('returns five', () => {
    /* ... */
  }) // space after

  it('returns six', () => {
    /* ... */
  })
}) // space after

describe('function d', () => {
  it('returns seven', () => {
    /* ... */
  }) // space after

  it('returns eight', () => {
    /* ... */
  })
})
```

I suppose the opposite could be enforced - for people who don't want spaces between blocks. Optionally, could add rules for only tests, only describe/before/after blocks, etc.

### LINKED ISSUE #1301: [new rule] Rule to enforce return after expect (code written)
```js
// BAD
expect(x).toEqual(y);

// GOOD
expect(x)
  .toEqual(y);
```

**Auto-fixable:** yes

**Code:**

https://github.com/tjw-lint/eslint-config-tjw-jest/blob/main/custom-eslint-rules/newline-before-expect-assertion.js

G-Rath:
This explores what it would be like to import the formatting rules from `eslint-plugin-jest-formatting` - I'm a big fan of these rules and have always thought it would make sense to include them as I don't think this is a particularly controversial style.

Related: https://github.com/dangreenisrael/eslint-plugin-jest-formatting/issues/122

Resolves #1301
Resolves #1251
Resolves #113
Resolves #17


G-Rath:
@SimenB I know you had opinions about including these in the past - would appreciate knowing if those have changed

G-Rath:
Note that for now I've focused on integrating the rules into our plugin with as minimal changes as possible - there's a few things I'd like to explore refactoring, but I've not found anything actually horribly broken so I think it's better to leave such refactors for a follow-up (or at least for after confirming we want to land this); likewise for adding any new configs.

SimenB (approved):
haven't looked through the code, but in general I'm happy to see this land 🙂 

(as long as the maintainers of the existing plugin are OK with it)

G-Rath:
@dangreenisrael are you happy for this to happen?

benkimpel:
> @dangreenisrael are you happy for this to happen?

@G-Rath I can't really speak for @dangreenisrael, but I'm a contributor on the repo in question and I think the best place for this plugin is probably within the Jest community org now that @SimenB has relaxed the rules around style plugins a bit.

Assuming Dan wants to donate the package, we'd want to put some kind of notice on the current repo/package to say that the plugin has been adopted by the Jest community. Are there any other changes on Dan's side that would need to happen? I think the current plugin is included in some packages that bundle several plugins together, so we might have to track those down.

SimenB:
Deprecating on npm is probably enough as people installing would get a warning. Which might be a bit heavy handed of course 😅

lasersio2:
@G-Rath was this PR abandoned?

G-Rath:
@lasersio2 nope, just waiting to hear back from the owner :)

G-Rath:
@SimenB just thinking about us doing a new major, given that @dangreenisrael seems to very quiet these days, how would you feel about us landing this without him having posted a "go ahead" on this particular PR? I feel like we're all pretty happy with this move in spirit 🤔

SimenB:
Yeah, that seems reasonable to me

G-Rath:
Ok so to recap some high level points before this lands:
  - these have been ported from https://github.com/dangreenisrael/eslint-plugin-jest-formatting
  - I've converted each rule to a dedicated file, and rewritten everything into TypeScript
  - I've updated the tests and code to get 100% code coverage
  - I've made basic refactors to address deprecations, and removed unneeded checks

The two notable things for end users are probably:
1. this sits out of band of `parseJestFnCall`, among other things meaning it technically doesn't respect the `globalAliases` or `globalPackage` settings
    - I looked at this initially but found it too hard for now because the way these rules work seems to be very different (iirc, it's a lot more token based than AST based); however, for the same reason I actually don't think this'll be as big of a deal..?
    - (either way, this'll be something I look back into at a later date)
2. I've not ported over the configs, and have no plans to at this time
    - notably because the configs use `overrides` which I don't think is appropriate for a plugin-provided config, as we have no idea what patterns developers are using for files that are appropriate for us to lint
    - and because I don't think it's much work to just configure the rules yourself

The rule configuration to use for `recommended` would be:

```
{
  plugins: ['jest'],
  rules: {
    'jest/padding-around-after-all-blocks': 'error',
    'jest/padding-around-after-each-blocks': 'error',
    'jest/padding-around-before-all-blocks': 'error',
    'jest/padding-around-before-each-blocks': 'error',
    'jest/padding-around-describe-blocks': 'error',
    'jest/padding-around-test-blocks': 'error'
  }
}
```

I'll aim to do some followup documentation updates after this is landed probably in the form of a light migration guide which'll cover this.

SchroederSteffen:
> 2. I've not ported over the configs, and have no plans to at this time
>    
>    * and because I don't think it's much work to just configure the rules yourself
> 
> The rule configuration to use for `recommended` would be:
> ...

Personally, I very much prefer extending the `recommended` (or here e.g. a `recommended-formatting`) configs than listing the rules myself. I'd like to request adding such a config for the formatting rules if possible. 🙂

SimenB:
a separate config seems fine, but without `overrides` like you mention

alfaproject:
Wouldn't that just be adding this the rule `padding-around-all`? What else would there be in the recommended config?

G-Rath:
> Wouldn't that just be adding the rule `padding-around-all`? What else would there be in the recommended config?

My understanding is that rule is more aggressive, which is why it's enabled by the `strict` config whereas the rules I posted above is what's in the `recommended`

> I very much prefer extending the recommended (or here e.g. a recommended-formatting) config over listing the rules myself

I mean, I figured it wasn't a big deal because it's only a few lines, shared config etc, and like we've got a bunch of cool rules that are not enabled in our configs that you would be missing out on if you only used configs, but I don't mind shipping a config since someone actually asked for it 😄

(the practical reason I didn't include it by default is once we have a config, it's a breaking change to do anything with it - so these days I prefer to do as little as possible and then add stuff when people ask for it)

I'll do a dedicated PR for the config so it can have its own changelog, and that'll give us a version where we can double smoketest that the rules are working fine before we officially start recommending them in our docs too - I'm thinking I'll just name the config `formatting`

SchroederSteffen:
> > I very much prefer extending the recommended (or here e.g. a recommended-formatting) config over listing the rules myself
> 
> I mean, I figured it wasn't a big deal because it's only a few lines, shared config etc, and like we've got a bunch of cool rules that are not enabled in our configs that you would be missing out on if you only used configs, but I don't mind shipping a config since someone actually asked for it 😄

Yeah, we often enable several rules in addition to shipped recommended configs. However, many times the presets are just fine and our own config is clearer and easier to maintain when using them.

Many thanks for adding it. 🙇

PR #1568, opened 2024-04-26
feat: prefer importing jest globals for specific types

tomquist:
Accessing the `jest` global in ESM must be done either through `import.meta.jest` or by importing it from `@jest/globals`. The latter is useful while migrating to ESM because the former is not accessible in non-ESM.

This adds an option to specify the types of globals for which we want to enforce the import.

### DIFF
--- a/docs/rules/prefer-importing-jest-globals.md
+++ b/docs/rules/prefer-importing-jest-globals.md
@@ -42,6 +42,42 @@ describe('foo', () => {
 });
 ```
 
+## Options
+
+This rule can be configured as follows
+
+```json
+{
+  "type": "object",
+  "properties": {
+    "types": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": ["hook", "describe", "test", "expect", "jest", "unknown"]
+      }
+    }
+  },
+  "additionalProperties": false
+}
+```
+
+#### types
+
+A list of Jest global types to enforce explicit imports for. By default, all
+Jest globals are enforced.
+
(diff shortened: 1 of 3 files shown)


SimenB (approved):
good idea 👍

G-Rath (changes requested):
cool, but I think let's drop the tuple utils stuff - I'm pretty confident we'll not be adding any new types anytime soon, so we should be safe to just directly write out the tuple

docs/rules/prefer-importing-jest-globals.md
```diff
@@ -42,6 +42,42 @@ describe('foo', () => {
 });
 ```
 
+## Options
+
+This rule can be configured as follows
+
+```json5
```
G-Rath, line 49: nit: could you make these `json` instead?

I think it's best we do that unless we're putting comments in the block, since ESLint doesn't support JSON5

src/rules/prefer-importing-jest-globals.ts
```diff
@@ -31,10 +41,29 @@ export default createRule({
...
+        additionalProperties: false,
+      },
+    ],
   },
-  defaultOptions: [],
+  defaultOptions: [
+    {
+      types: allJestFnTypes as JestFnType[],
+    },
+  ],
```
G-Rath, line 64:
```suggestion
  defaultOptions: [{ types: allJestFnTypes }],
```

G-Rath (approved):

tomquist:
Thanks for the quick review!

PR #1573, opened 2024-04-29
refactor: strict typing and deprecation fix in type definitions

y-hsgw:
I apologize for the sudden PR, but I'd like to propose some strict typing enhancements.
Please consider the following changes:

- Updated deprecated type definitions for TSESLint.Linter.Config.
- Removed unnecessary type assertions and enhanced strict typing.

Note: As this is the first PR for this repository, please feel free to provide any feedback or suggestions if there are any deficiencies.

### DIFF
--- a/src/rules/__tests__/unbound-method.test.ts
+++ b/src/rules/__tests__/unbound-method.test.ts
@@ -113,37 +113,33 @@ const invalidTestCases: Array<TSESLint.InvalidTestCase<MessageIds, Options>> = [
     ],
   },
   // toThrow matchers call the expected value (which is expected to be a function)
-  ...toThrowMatchers.map<TSESLint.InvalidTestCase<MessageIds, Options>>(
-    matcher => ({
-      code: dedent`
+  ...toThrowMatchers.map(matcher => ({
+    code: dedent`
       ${ConsoleClassAndVariableCode}
 
       expect(console.log).${matcher}();
     `,
-      errors: [
-        {
-          line: 9,
-          messageId: 'unboundWithoutThisAnnotation',
-        },
-      ],
-    }),
-  ),
+    errors: [
+      {
+        line: 9,
+        messageId: 'unboundWithoutThisAnnotation' as const,
+      },
+    ],
+  })),
   // toThrow matchers call the expected value (which is expected to be a function)
-  ...toThrowMatchers.map<TSESLint.InvalidTestCase<MessageIds, Options>>(
-    matcher => ({
-      code: dedent`
+  ...toThrowMatchers.map(matcher => ({
+    code: dedent`
       ${ConsoleClassAndVariableCode}
 
       expect(console.log).not.${matcher}();
     `,
-      errors: [
-        {
-          line: 9,
-          messageId: 'unboundWithoutThisAnnotation',
-        },
-      ],
-    }),
-  ),
+    errors: [
+      {
+        line: 9,
+        messageId: 'unboundWithoutThisAnnotation' as const,
+      },
+    ],
+  })),
 ];
 
 const requireRule = (throwWhenRequiring: boolean) => {
--- a/src/rules/__tests__/unbound-method.test.ts
+++ b/src/rules/__tests__/unbound-method.test.ts
@@ -54,7 +54,7 @@ const toThrowMatchers = [
   'toThrowError',
   'toThrowErrorMatchingSnapshot',
   'toThrowErrorMatchingInlineSnapshot',
-];
+] as const;
 
 const validTestCases: string[] = [
   ...[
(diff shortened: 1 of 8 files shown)


G-Rath:
I'll review this in more detail tomorrow when I'm at my laptop, but I don't see anything here that I'd call strictly better - can you describe what problem you're seeing that you're trying to solve?

src/rules/__tests__/unbound-method.test.ts
```diff
@@ -113,33 +113,37 @@ const invalidTestCases: Array<TSESLint.InvalidTestCase<MessageIds, Options>> = [
     ],
   },
   // toThrow matchers call the expected value (which is expected to be a function)
-  ...toThrowMatchers.map(matcher => ({
-    code: dedent`
+  ...toThrowMatchers.map<TSESLint.InvalidTestCase<MessageIds, Options>>(
```
G-Rath, line 116: We purposely don't define this as it's a complicated type so it's preferred to let typescript infer it

src/rules/__tests__/unbound-method.test.ts
```diff
@@ -113,33 +113,37 @@ const invalidTestCases: Array<TSESLint.InvalidTestCase<MessageIds, Options>> = [
...
+          line: 9,
+          messageId: 'unboundWithoutThisAnnotation',
+        },
+      ],
+    }),
+  ),
   // toThrow matchers call the expected value (which is expected to be a function)
-  ...toThrowMatchers.map(matcher => ({
-    code: dedent`
+  ...toThrowMatchers.map<TSESLint.InvalidTestCase<MessageIds, Options>>(
```
G-Rath, line 132: Ditto for this

src/rules/require-hook.ts
```diff
@@ -39,9 +39,11 @@ const shouldBeInHook = (
     case AST_NODE_TYPES.ExpressionStatement:
       return shouldBeInHook(node.expression, context, allowedFunctionCalls);
     case AST_NODE_TYPES.CallExpression:
+      const nodeName = getNodeName(node);
```
G-Rath, line 42: This is a runtime de-optimization - previously we'd only determine the node name if the other condition hasn't matched; now we'll always be getting the node name even if we don't need it

src/rules/valid-expect-in-promise.ts
```diff
@@ -110,7 +110,8 @@ const isPromiseMethodThatUsesValue = (
     }
 
     if (
-      ['Promise.resolve', 'Promise.reject'].includes(nodeName as string) &&
+      nodeName &&
```
G-Rath, line 113: This check is arguably more expensive at runtime - it's very minor, but I'd rather keep the cast

src/rules/valid-title.ts
```diff
@@ -74,13 +74,13 @@ const compileMatcherPatterns = (
...
 
-const MatcherAndMessageSchema: JSONSchema.JSONSchema4 = {
+const MatcherAndMessageSchema = {
   type: 'array',
   items: { type: 'string' },
   minItems: 1,
   maxItems: 2,
   additionalItems: false,
-} as const;
+} as const satisfies JSONSchema.JSONSchema4;
```
G-Rath, line 83: This is not strictly better than setting the type - there's no need to change this

src/rules/valid-expect-in-promise.ts
y-hsgw: Got it. I realized I missed considering performance aspects 🙇‍♂️
 I've addressed it in commit b030cc7.

src/rules/__tests__/unbound-method.test.ts
y-hsgw: Got it. I've addressed it in commit https://github.com/jest-community/eslint-plugin-jest/commit/b030cc748361ecd7692845e22114be041700e9b3.

src/rules/__tests__/unbound-method.test.ts
y-hsgw: Got it. I've addressed it in commit https://github.com/jest-community/eslint-plugin-jest/commit/b030cc748361ecd7692845e22114be041700e9b3.

y-hsgw:
> I'll review this in more detail tomorrow when I'm at my laptop, but I don't see anything here that I'd call strictly better - can you describe what problem you're seeing that you're trying to solve?


Here are the two points:

- Updated due to deprecation of TSESLint.Linter.Config.
- Removed unnecessary type assertions. I would suggest that type assertions should be avoided as much as possible from a conservatism standpoint.

y-hsgw:
Thank you for your explanation. Indeed, as you mentioned, this PR was a minor adjustment. I will close this PR now. Thank you for your assistance 🙇‍♂️

G-Rath:
No worries - as I said I appreciate the enthusiasm; feel free to have a look over some of our issues and have a go at tackling one of them if you want to help out :)

G-Rath:
@y-hsgw just catching up on my general todo list of which re-reviewing this was one as I think there are a couple of interesting possible changes highlighted (#1577 being one).

It does look like some of our `as const`'s are not needed - I'm not sure if that's because TypeScript has gotten better, the types have changed, or there's some footgun awaiting us, but if you want to do a refactoring PR that _only removes_ `as const`s that don't result in type errors, that'd be cool.

y-hsgw:
> It does look like some of our `as const`'s are not needed - I'm not sure if that's because TypeScript has gotten better, the types have changed, or there's some footgun awaiting us, but if you want to do a refactoring PR that _only removes_ `as const`s that don't result in type errors, that'd be cool.

Thanks! I've created a PR at https://github.com/jest-community/eslint-plugin-jest/pull/1578! Please review it.

Issue #1140, opened 2022-06-02
[valid-expect] include fixer for adding missing await


jayu:
How do you feel about making `jest/valid-expect` rule fixable for cases where `await` is missing ?

So adding `await` in front of `expect` in such cases:
```ts
it('should', async () => {
   expect(someAsyncThing).rejects.toBe(whatever)
})

//or 

it('should', async () => {
   expect(someAsyncThing).resolves.toBe(whatever)
})
```

Fixed code then would look like

```ts
it('should', async () => {
   await expect(someAsyncThing).rejects.toBe(whatever)
})

//or 

it('should', async () => {
   await expect(someAsyncThing).resolves.toBe(whatever)
})
```

I'm happy to give it a try if there is a chance it will be merged :)

G-Rath:
Yup sounds good - we probably should have it as a suggestion if the function isn't already `async`.

y-hsgw:
@G-Rath As for the remaining tasks for this issue, would it be appropriate to [suggest](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions) that if the target for validation is not async?


```ts
it('should', async () => {
  const someAsyncThing = () => { ... }
  expect(someAsyncThing).rejects.toBe(whatever)
})
```

G-Rath:
@y-hsgw if you mean suggesting adding `await` regardless of if `async` is present, then yeah that's probably fine; though I think we should be able to successfully add `async` to at least some patterns.

Broadly speaking the logic should be something like:
1. find the closest function expression upwards in the tree (I'm pretty sure that's what `findFirstAsyncFunction` does exactly except it also checks if its `async`; but worth double checking)
2. if it is not `async`, include making a fixer for making it `async`
3. if the line has a `return` and ... etc (all the current fixer behaviour)
4. include a fixer for adding `await`

While it might seem complicated on the face, I'm actually pretty sure we should be able to always accurately know and fix `async/await`.

I'm happy to do incremental improvement PRs too, so we can aim for addressing the simple cases and then a followup PR for handling more complex ones

PR #1574, opened 2024-05-02
feat(valid-expect): supporting automatically fixing missing `await` in some cases

y-hsgw:
When `await` is missing ,  it's adding `await`.
Relates to https://github.com/jest-community/eslint-plugin-jest/issues/1140

### DIFF
--- a/docs/rules/valid-expect.md
+++ b/docs/rules/valid-expect.md
@@ -8,7 +8,8 @@
 
 <!-- end auto-generated rule header -->
 
-Note: Test function will be fixed if it is async and does not have await in the
+> [!NOTE]
+> Test function will be fixed if it is async and does not have await in the
 async assertion.
 
 Ensure `expect()` is called with a single argument and there is an actual
--- a/docs/rules/valid-expect.md
+++ b/docs/rules/valid-expect.md
@@ -6,6 +6,9 @@
 🔧 This rule is automatically fixable by the
 [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
 
+Note: Test function will be fixed if it is async and does not have await in the
+async assertion.
+
 <!-- end auto-generated rule header -->
 
 Ensure `expect()` is called with a single argument and there is an actual
--- a/docs/rules/valid-expect.md
+++ b/docs/rules/valid-expect.md
@@ -3,6 +3,9 @@
 💼 This rule is enabled in the ✅ `recommended`
 [config](https://github.com/jest-community/eslint-plugin-jest/blob/main/README.md#shareable-configurations).
 
+🔧 This rule is automatically fixable by the
+[`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
 <!-- end auto-generated rule header -->
 
 Ensure `expect()` is called with a single argument and there is an actual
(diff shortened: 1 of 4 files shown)


docs/rules/valid-expect.md
```diff
@@ -3,6 +3,9 @@
 💼 This rule is enabled in the ✅ `recommended`
 [config](https://github.com/jest-community/eslint-plugin-jest/blob/main/README.md#shareable-configurations).
 
+🔧 This rule is automatically fixable by the
```
SimenB, line 6: can we have it say that "some patterns are fixable" or something like that? specifically now, only if the function is already async

y-hsgw: I attempted to fix it with commit 90028bf. What do you think?

G-Rath (approved):
This looks like a good start to me, though noting it is by no means complete so I've updated the title to reflect that and removed the "resolving" of #1140

I think though it's fine to land as-is since it's an improvement regardless 🎉

docs/rules/valid-expect.md
```diff
@@ -3,8 +3,14 @@
...
+[`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
 <!-- end auto-generated rule header -->
 
+Note: Test function will be fixed if it is async and does not have await in the
+async assertion.
```
G-Rath, line 12: we could use [fancy notes](https://github.com/orgs/community/discussions/16925) here

docs/rules/valid-expect.md
G-Rath: fwiw it seems currently the docs generator doesn't support customizing this in anyway, so I think what you've done is fine for now - @bmish might be something worth adding; even just the ability to explicitly name particular rules as "partially fixable" in the config which would case this string to include with "in some cases" suffix

docs/rules/valid-expect.md
y-hsgw: nice! I made the fix in commit ee54e6a.

src/rules/valid-expect.ts
```diff
@@ -48,6 +50,18 @@ const findPromiseCallExpressionNode = (node: TSESTree.Node) =>
...
+  parent,
+}: TSESTree.Node): TSESTree.Node | null => {
+  if (!parent) {
+    return null;
+  }
+
+  return isFunction(parent) && parent.async
+    ? parent
+    : findFirstAsyncFunction(parent);
+};
```
FloEdelmann, line 63: Sorry for the late response; I only saw this after the new release.

Doesn't this cause false-positives for non-async functions contained in async functions? It's probably a really rare situation in test files, but I think it would be more correct like this:

```ts
  if (!parent) {
    return null;
  }

  if (isFunction(parent)) {
    return parent.async ? parent : null;
  }

  return findFirstAsyncFunction(parent);
```

FloEdelmann: Ah, I see the logic is changed in https://github.com/jest-community/eslint-plugin-jest/pull/1579 anyway.

G-Rath: Do you have a test case that's failing? If so please share either way so we can ensure that going forward it doesn't fail

FloEdelmann: No failing test case, just noticed this while reading the code.

PR #1579, opened 2024-05-03
feat(valid-expect): supporting automatically fixing adding async in some cases

### LINKED ISSUE #1140: [valid-expect] include fixer for adding missing await

y-hsgw:
Resolves #1140

### DIFF
--- a/src/rules/valid-expect.ts
+++ b/src/rules/valid-expect.ts
@@ -63,7 +63,10 @@ const findFirstFunctionExpression = ({
 
 const getNormalizeFunctionExpression = (
   functionExpression: FunctionExpression,
-): TSESTree.Node => {
+):
+  | TSESTree.PropertyComputedName
+  | TSESTree.PropertyNonComputedName
+  | FunctionExpression => {
   if (
     functionExpression.parent.type === AST_NODE_TYPES.Property &&
     functionExpression.type === AST_NODE_TYPES.FunctionExpression
@@ -377,10 +380,13 @@ export default createRule<[Options], MessageIds>({
                 const targetFunction =
                   getNormalizeFunctionExpression(functionExpression);
 
-                return fixer.insertTextBefore(targetFunction, 'async ');
+                return [
+                  fixer.insertTextBefore(targetFunction, 'async '),
+                  fixer.insertTextBefore(finalNode, 'await '),
+                ];
               }
               const returnStatement =
-                finalNode.parent?.type === AST_NODE_TYPES.ReturnStatement
+                finalNode.parent.type === AST_NODE_TYPES.ReturnStatement
                   ? finalNode.parent
                   : null;
 
--- a/src/rules/valid-expect.ts
+++ b/src/rules/valid-expect.ts
@@ -4,6 +4,7 @@
  */
 
 import { AST_NODE_TYPES, type TSESTree } from '@typescript-eslint/utils';
+import type { RuleFix } from '@typescript-eslint/utils/ts-eslint';
 import {
   type FunctionExpression,
   ModifierName,
@@ -361,25 +362,6 @@ export default createRule<[Options], MessageIds>({
           // if we didn't warn user already
           !promiseArrayExceptionExists(finalNode.loc)
         ) {
-          const functionExpression = findFirstFunctionExpression(finalNode);
-
-          if (functionExpression && !functionExpression.async) {
-            context.report({
-              loc: functionExpression.loc,
-              data: { orReturned },
-              messageId:
-                finalNode === targetNode
-                  ? 'asyncMustBeAwaited'
-                  : 'promisesWithAsyncAssertionsMustBeAwaited',
-              node,
-              fix(fixer) {
-                const targetFunction =
-                  getNormalizeFunctionExpression(functionExpression);
-
-                return fixer.insertTextBefore(targetFunction, 'async ');
-              },
-            });
-          }
           context.report({
             loc: finalNode.loc,
             data: { orReturned },
@@ -389,10 +371,20 @@ export default createRule<[Options], MessageIds>({
                 : 'promisesWithAsyncAssertionsMustBeAwaited',
             node,
             fix(fixer) {
+              const functionExpression = findFirstFunctionExpression(finalNode);
+
               if (!functionExpression) {
-                return null;
+                return [];
               }
 
+              const fixes: RuleFix[] = [];
+
+              if (!functionExpression.async) {
+                const targetFunction =
+                  getNormalizeFunctionExpression(functionExpression);
+
+                fixes.push(fixer.insertTextBefore(targetFunction, 'async '));
+              }
               const returnStatement =
                 finalNode.parent.type === AST_NODE_TYPES.ReturnStatement
                   ? finalNode.parent
@@ -403,10 +395,13 @@ export default createRule<[Options], MessageIds>({
                   getSourceCode(context).getText(returnStatement);
                 const replacedText = sourceCodeText.replace('return', 'await');
 
-                return fixer.replaceText(returnStatement, replacedText);
+                return [
+                  ...fixes,
+                  fixer.replaceText(returnStatement, replacedText),
+                ];
               }
 
-              return fixer.insertTextBefore(finalNode, 'await ');
+              return [...fixes, fixer.insertTextBefore(finalNode, 'await ')];
             },
           });
 
--- a/src/rules/valid-expect.ts
+++ b/src/rules/valid-expect.ts
@@ -4,6 +4,7 @@
  */
 
 import { AST_NODE_TYPES, type TSESTree } from '@typescript-eslint/utils';
+import type { RuleFix } from '@typescript-eslint/utils/ts-eslint';
 import {
   type FunctionExpression,
   ModifierName,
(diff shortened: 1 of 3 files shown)


G-Rath (changes requested):
looks like a good start, but this isn't quite the right approach to being incremental 😅 - a fix should address the full issue.

What I meant by incremental is landing a fixer for e.g. `expect.extend` or functions without any nesting or a single expect, then digging into advanced cases like "what if you have a function within your test function?"; but also you don't have to do it like that either - if you want to tackle more in a single PR that's great! based on what you've done here for example, I think you might just need to switch to returning an array of fixes and then this might be landable

y-hsgw:
I misunderstood the concept of incremental improvement. 
Instead of partially addressing the issue, I should have aimed to not only add 'async' but also include 'await' where needed. 
As you suggested, it seems feasible to address the entire problem within this PR, so I'll give it a shot!

y-hsgw:
I'm encountering a situation where the logic to add 'await' to all expects isn't functioning correctly when there are multiple asynchronous expectations.
If there's a suitable logic you can suggest, I'd appreciate your guidance🙇‍♂️

G-Rath:
@y-hsgw push up your code and I can take a look. That also sounds like a good example of where we might want to be interactive: if you're having trouble with the fixing, consider focusing on having it just bail out instead for now.

y-hsgw:
I might have misunderstood.
it seems [my changes](https://github.com/jest-community/eslint-plugin-jest/pull/1579/files#diff-a63b56f41f9d7b260c021d0db0c224fde76950dea886f2d6b97869ca47dd2070R1002-R1009) to the test could be incorrect. The [official documentation](https://eslint.org/docs/latest/integrate/nodejs-api#testing-fixes:~:text=ESLint%20makes%20its,the%20first%20fix.) states, 'RuleTester applies only the first fix.' However, I attempted to validate multiple cases simultaneously.

G-Rath:
@y-hsgw I think you might be misunderstanding - that relates to when there are conflicts:

> When there is a conflict between two fixes (because they apply to the same section of code) RuleTester applies only the first fix.

The fixer works by taking (via the return) an _array_ of fix _operations_ to apply which make up a _single_ fix - if there is only one operation, then you can just return that but you can think about it as being an array of operations at all times.

So when those docs talk about "applying all fixes" and "applies only the first fix" they're referring to the high-level fix aka "a collection of fix operations", which is what you're building here - the issue is right now you're returning two seperate fix operations depending on conditions, when you want to be returning a single array with _both_ operations added conditionally depending on if they're needed for the particular code under lint.

y-hsgw:
@G-Rath 
Thank you for the clarification. I had misunderstood and thought there might be conflicts. However, it seems that even with the current logic, async and await are added as shown in the attached video. Would it still be appropriate to return a single array?


https://github.com/jest-community/eslint-plugin-jest/assets/49516827/9934cc2d-8bf6-4622-b5b8-1d36a1132a29

G-Rath:
Yes because as the tests show via the `output`, you're not actually fixing the issue in one pass - it works in your editor because ESLint re-runs rules up to 10 times in a pass to handle overlap between rule fixes; e.g. if we have a fix that outputs single quotes and prettier with double quotes is setup, then the _final_ output of a single run of `eslint --fix` will produce the right code, because both rules actually got run multiple times.

It should be very trivial to return a single array too :)

y-hsgw:
@G-Rath 
I've refactored to return an array, but the tests fail when there are multiple expects. Do you have any insights into what might be causing this?
[58c4046](https://github.com/jest-community/eslint-plugin-jest/pull/1579/commits/58c4046bde1f786e948d56f06b4541a03c6771a3)

G-Rath:
@y-hsgw my guess is that it's because you're adding the `async` in the first fixer, which ends up changing the logic in the second fixer - if I add the `async` to the function, then both `await`s get added correctly.

I would recommend focusing on a single test (you can add `only: true` to a test case for this) and stick `console.log`s around every condition and check, then compare what you get with the `async` present vs when its not.

src/rules/valid-expect.ts
```diff
@@ -362,6 +361,25 @@ export default createRule<[Options], MessageIds>({
...
+                  : 'promisesWithAsyncAssertionsMustBeAwaited',
+              node,
+              fix(fixer) {
+                const targetFunction =
+                  getNormalizeFunctionExpression(functionExpression);
+
+                return fixer.insertTextBefore(targetFunction, 'async ');
+              },
+            });
+          }
```
y-hsgw, line 382: @G-Rath 
I've attempted to modify the logic, but is my understanding correct?

G-Rath: No this isn't correct, since we're now reporting an error for the function which is 1. not the actual issue and 2. reported per expect (which itself already reports an error).

y-hsgw: I see. Do I need to add both `async` and `await` within a single `context.report`?

G-Rath: Yes

y-hsgw: Thanks.
>  if I add the async to the function, then both awaits get added correctly.

Is my understanding of adding async in the following code correct?

```ts
fixer.insertTextBefore(targetFunction, 'async ');
```

G-Rath: Yup that looks like it's probably right, though it's not the only way.

y-hsgw: > my guess is that it's because you're adding the async in the first fixer, which ends up changing the logic in the second fixer - if I add the async to the function, then both awaits get added correctly.

Could you please confirm my understanding? Are you pointing out the order of the fixer's array to return? Is adding async first, as shown below, the issue?

```ts
const fixes: RuleFix[] = [];

if (!functionExpression.async) {
  const targetFunction =
    getNormalizeFunctionExpression(functionExpression);

  fixes.push(fixer.insertTextBefore(targetFunction, 'async '));
}
return [...fixes, fixer.insertTextBefore(finalNode, 'await ')];
 ```

Comment (G-Rath) [reply src/rules/valid-expect.ts:382]: I don't know what the issue is, otherwise I'd have already fixed this 😅

You'll either need to dig into it or leave this with me to dig into when I next have the bandwidth.

If you're finding it overwhelming, keep in mind you can technically "do what you want" locally - sometimes I delete all the tests for a rule or create a brand new one where I can focus on a particular thing and explore what ESLint is doing with console log etc.

https://astexplorer.net/ is also very useful - I often use the "transform" option to debug particular fixers:

image

Also make sure to have a look at other rules to see how they're doing things and just consider stuff from another angle - in this case, since you're having trouble fixing when there are multiple expects, you might decide to focus on autofixing only when there is one expect; that might turn out to be harder to detect within the `CallExpression` since you can't "unreport" an error, but in that case you could push the results into a global array and then do the reporting in `Program:exit` (that's something a few of our other rules do) - it's not ideal but if it works its an improvement.

File: src/rules/valid-expect.ts [hunk 2]
```diff
@@ -346,39 +368,63 @@ export default createRule<[Options], MessageIds>({
...
+              targetNode === finalNode
                 ? 'asyncMustBeAwaited'
                 : 'promisesWithAsyncAssertionsMustBeAwaited',
+          });
+        }
+        if (isParentArrayExpression) {
+          pushPromiseArrayException(finalNode.loc);
+        }
+      },
+      'Program:exit'() {
```
y-hsgw, line 383: @G-Rath 

>  in that case you could push the results into a global array and then do the reporting in `Program:exit`

I tried making adjustments based on the advice you provided. How does it look?

src/rules/valid-expect.ts
```diff
@@ -4,7 +4,9 @@
  */
 
 import { AST_NODE_TYPES, type TSESTree } from '@typescript-eslint/utils';
+import type { RuleFix } from '@typescript-eslint/utils/ts-eslint';
```
G-Rath, line 7: can you pull this type from `TSESTree` (or it might be `TSESLint` - either way, you can access it through the `@typescript-eslint/utils` import)

y-hsgw: Thanks! Fixed in 3b27d09.

G-Rath (approved):
awesome stuff! 🎉

G-Rath:
hmm

```
Error: Command failed with exit code 1: git push --tags https://x-access-token:[secure]@github.com/jest-community/eslint-plugin-jest.git HEAD:main
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: error: 6 of 6 required status checks are expected.        
To https://github.com/jest-community/eslint-plugin-jest.git
 ! [remote rejected] HEAD -> main (protected branch hook declined)
```

For some reason the `main` branch decided to skip our `docs` job meaning that releasing then failed due to it being a required status check 🤔

PR #1592, opened 2024-05-21
revert: "chore(deps): lock file maintenance"

G-Rath:
Reverts jest-community/eslint-plugin-jest#1590


G-Rath:
Addressed through #1593

PR #1593, opened 2024-05-21
test: only run linting when using ESLint v8

G-Rath:
I've not dug into the actual code yet, but I'm pretty sure this is why #1590 is failing CI - more specifically, the latest version of `eslint-plugin-n` (which does not support ESLint v7) is now doing something v7 doesn't support


SimenB (approved):
:old-man-yells-at-code:

PR #1594, opened 2024-05-23
docs: refer to eslint 9 migration guide

froesecom:
The differences between `.eslintrc` config formats and `eslint.config.js` config formats can be confusing. Eslint has a migration guide that is helpful for understanding the differences and may be helpful for other users of `eslint-plugin-jest`.
Thanks for your hard work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ yarn add --dev eslint eslint-plugin-jest
 > Refer to the
 > [ESLint documentation on the new configuration file format](https://eslint.org/docs/latest/use/configure/configuration-files-new)
 > for more.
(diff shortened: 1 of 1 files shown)


G-Rath:
I've got mixed feelings about this because we're already linking to the main doc which has the migration guide on the sidebar, and ideally I'd like to leave this up to ESLint as much as possible (e.g. if they decide to move the migration doc, that's their call).

I would say this PR would be better made to ESLint themselves - I think it would be useful for the main configuration files page we're already linking to to include a link to the migration guide at the top

froesecom:
Yes, fair enough. I only put this PR up because the `eslint-plugin-jest` docs don't provide examples for the `eslint.config.js` file and I found it pretty confusing to translate between `.eslintrc` and `eslint.config.js`. Anyway, I'll close this. If you'd like me to help with providing examples for `eslint.config.js` I'm happy to do so.

PR #1599, opened 2024-05-30
feat(prefer-jest-mocked): add new rule

### LINKED ISSUE #1470: [new-rule] Force to use `jest.mocked(fn)` instead of `fn as jest.Mock`
**Requirements:**
- ts only
- Jest version 28+ only
- auto-fixable 
- should cover `as jest.MockedFunction` type casting
- should cover `as jest.Mock` type casting
- should fix all the next examples

**Examples:**

```tsx
// bad
const mockedUseFocused = useFocused as jest.MockedFunction<typeof useFocused>;
(AccessibilityInfo.isAccessibilityServiceEnabled as jest.Mock).mockResolvedValue(false);
const filter = (MessageService.getMessage as jest.Mock).mock.calls[0][0];
((getIsUserProfileFeatureSwitchEnabled as unknown) as jest.Mock).mockReturnValue(false);
((obj.obg2.obg3.obgN as jest.Mock).mockReturnValue(1);
((obj.myfn().myfn2 as jest.Mock).mockReturnValue('str');


// better
const mockedUseFocused = jest.mocked(useFocused)
jest.mocked(AccessibilityInfo.isAccessibilityServiceEnabled).mockResolvedValue(false);
const filter = jest.mocked(MessageService.getMessage).mock.calls[0][0];
jest.mocked(getIsUserProfileFeatureSwitchEnabled).mockReturnValue(false);
jest.mocked(obj.obg2.obg3.obgN).mockReturnValue(1);
jest.mocked(obj.myfn().myfn2).mockReturnValue('str');
```

**Draft implementation**

```js
function unpackMemberExpression(node) {
  if (node.type === 'Identifier') {
    return node.name;
  }
  return `${unpackMemberExpression(node.object)}.${node.property.name}`;
}

function getFnName(node) {
  if (node.type === 'MemberExpression') {
    // case `Obj.myFn as jest.Mock`
    return unpackMemberExpression(node);
  }
  if (node.type === 'TSAsExpression') {
    // case: `myFn as unknown as jest.Mock`
    return getFnName(node.expression);
  }

  // case `myFn as jest.Mock`
  return node.name;
}

const mockTypes = ['MockedFunction', 'Mock'];

// Bad:
//   (myFn as jest.Mock)
//   myFn as jest.MockedFunction<typeof myFn>
// Good:
//   jest.mocked(myFn).mockReturnValue(...)
const noAsJestExpression = {
  meta: {
    type: 'suggestion',
    fixable: 'code',
    messages: {
      noAsJestExpression:
        'Please use jest api `jest.mocked({{fnName}})` instead of `{{fnName}} as jest.Mock`',
    },
  },
  create(context) {
    return {
      TSAsExpression(node) {
        const isJestMock =
          node?.typeAnnotation?.typeName?.left?.name === 'jest' &&
          mockTypes.includes(node?.typeAnnotation?.typeName?.right?.name);

        if (!isJestMock) {
          return;
        }

        const fnName = getFnName(node.expression);

        if (!fnName) {
          console.log(
            "[no-as-jest-expression]: Can't extracted fn name from expression:\n" +
              context.getSourceCode().getText(node)
          );
          return;
        }

        context.report({
          node,
          messageId: 'noAsJestExpression',
          data: {
            fnName,
          },
          fix(fixer) {
            return fixer.replaceText(node, `jest.mocked(${fnName})`);
          },
        });
      },
    };
  },
};

module.exports = noAsJestExpression;
```

MillerSvt:
Resolves #1470


src/rules/__tests__/prefer-mocked.test.ts
```diff
@@ -0,0 +1,321 @@
...
+const rootPath = getFixturesRootDir();
+
+const ruleTester = new RuleTester({
+  parser: require.resolve('@typescript-eslint/parser'),
+  parserOptions: {
+    sourceType: 'module',
+    tsconfigRootDir: rootPath,
+    project: './tsconfig.json',
+  },
+});
```
G-Rath, line 19: you shouldn't need `project` and co since we're not using type information

```suggestion
const ruleTester = new RuleTester({
  parser: require.resolve('@typescript-eslint/parser'),
});
```

src/rules/prefer-mocked.ts
```diff
@@ -0,0 +1,56 @@
...
+      useJestMocked: 'Prefer jest.mocked({{ replacement }})',
+    },
+    schema: [],
+    type: 'suggestion',
+    fixable: 'code',
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      'TSAsExpression:has(TSTypeReference > TSQualifiedName:has(Identifier.left[name="jest"]):has(Identifier.right[name="Mock"],Identifier.right[name="MockedFunction"]))'(
```
G-Rath, line 38: Please don't use complex expressions like this - they're brittle and harder to maintain (in part because TypeScript has no idea what they're doing)

src/rules/prefer-mocked.ts
```diff
@@ -0,0 +1,56 @@
...
+        node: ValidatedTsAsExpression,
+      ) {
+        const fnName = getFnName(node.expression, context.getSourceCode().text);
```
G-Rath, line 41: Can you see if we can reuse our [util here?](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/utils/followTypeAssertionChain.ts) - I expect you probably can by checking if the `parent` is an assertion node

G-Rath (changes requested):
thanks, looks like a good start

in addition to my specific comments:
  - some of your tests have whitespace on their "empty" lines; would be great if you could remove that
  - we should have a couple of cases that cover chained assertions and with `jest` in the middle (e.g. `as unknown as string as unknown as jest.MockedFunction` and `as unknown as jest.MockedFunction as string as jest.MockedFunction` - it's silly but legal)
  - we should support angle bracket assertions too

src/rules/__tests__/prefer-mocked.test.ts
```diff
@@ -0,0 +1,407 @@
+import dedent from 'dedent';
+import rule from '../prefer-mocked';
+import { FlatCompatRuleTester as RuleTester } from './test-utils';
+
+const ruleTester = new RuleTester({
+  parser: require.resolve('@typescript-eslint/parser'),
+});
+
+ruleTester.run('prefer-mocked', rule, {
```
G-Rath, line 9: nit: generally I'd prefer to not use `dedent` for single line calls, and not have a newline between cases

src/rules/prefer-mocked.ts
```diff
@@ -0,0 +1,74 @@
+import { AST_NODE_TYPES, type TSESTree } from '@typescript-eslint/utils';
+import { createRule, followTypeAssertionChain } from './utils';
+
+const mockTypes = ['Mock', 'MockedFunction', 'MockedClass', 'MockedObject'];
+
+export default createRule({
+  name: __filename,
+  meta: {
+    docs: {
+      description: 'Prefer jest.mocked() over (fn as jest.Mock)',
```
G-Rath, line 10:
```suggestion
      description: 'Prefer `jest.mocked()` over `fn as jest.Mock`',
```

src/rules/prefer-mocked.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    messages: {
+      useJestMocked: 'Prefer jest.mocked({{ replacement }})',
```
G-Rath, line 13:
```suggestion
      useJestMocked: 'Prefer `jest.mocked({{ replacement }})`',
```

src/rules/__tests__/prefer-mocked.test.ts
```diff
@@ -0,0 +1,407 @@
...
+      type MockType = jest.Mock;
+      const mockFn = jest.fn();
+      (mockFn as MockType).mockReturnValue(1);
+    `,
+
+    dedent`
+      (foo as Mock.jest).mockReturnValue(1);
+    `,
+  ],
+  invalid: [
```
G-Rath, line 61: can you add the expected `data` values for the message in each error?

src/rules/prefer-mocked.ts
```diff
@@ -0,0 +1,74 @@
...
+      ) {
+        return;
+      }
+
+      const fnName = context
+        .getSourceCode()
+        .text.slice(...followTypeAssertionChain(node.expression).range);
+
+      context.report({
+        node: node,
```
G-Rath, line 50:
```suggestion
        node,
```

src/rules/prefer-mocked.ts
```diff
@@ -0,0 +1,74 @@
...
+        messageId: 'useJestMocked',
+        data: {
+          replacement: '',
```
G-Rath, line 53: either we shouldn't have this or it should have a proper value (this'll get picked up by adding it to the tests)

src/rules/prefer-mocked.ts
```diff
@@ -0,0 +1,74 @@
...
+        left.type !== AST_NODE_TYPES.Identifier ||
+        right.type !== AST_NODE_TYPES.Identifier ||
+        left.name !== 'jest' ||
+        !mockTypes.includes(right.name)
+      ) {
+        return;
+      }
+
+      const fnName = context
+        .getSourceCode()
```
G-Rath, line 46: please use our `getSourceCode` helper here, as `getSourceCode` is removed in ESLint v9

src/rules/prefer-mocked.ts
G-Rath, line 1: nit: I wonder if it might be slightly better to call this `prefer-jest-mocked`? 🤔

G-Rath (changes requested):
Looks good, just got a few last tweaks - could you also add a test case or two covering multiline shenanigans?

i.e.

```
(
  new Array(100)
    .fill(undefined)
    .map(x => x.value)
    .filter(v => !!v).myProperty as jest.MockedFunction<{
    method: () => void;
  }>
).mockReturnValue(1);
```

Also could you check in a test that uses square brackets, just for coverage? i.e. `(Obj['foo'] as jest.MockedFunction).mockReturnValue(1);`

src/rules/prefer-mocked.ts
MillerSvt: I think it will be better to remove this parameter, because multi-line code will affect the output of errors

G-Rath (approved):
thanks!

PR #1626, opened 2024-07-06
chore: skip lib checks for now

G-Rath:
Currently our typechecking is failing due to a conflict between the latest TypeScript and `@typescript-eslint/typescript-estree` v6 - this unblocks us without needing to immediately drop support for that version

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,8 @@
     "strict": true,
     "resolveJsonModule": true,
     "isolatedModules": true,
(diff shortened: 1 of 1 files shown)


SimenB:
can't we just ditch v7, then release a new major with v6?

G-Rath:
yup I'm happy to do that, and was thinking about that - I've landed this in the meantime so we're immediately unblocked but I'm happy to do the major this/next week if you're keen

PR #1639, opened 2024-08-06
fix(prefer-importing-jest-globals): support typescript-eslint parser

### LINKED ISSUE #1634: [prefer-importing-jest-globals] autofix is broken
The plugin is relying on the old version of the eslint config to check whether the code is using esmodules or commonjs.

```
const isModule = context.parserOptions.sourceType === 'module';
```
should be something like
```
const isModule =  context.parserOptions.sourceType === 'module' || context.languageOptions.sourceType === 'module';
```
There doesn't seem to be an easy workaround for this, as the old structure is invalid in the new version of eslint. (ie I can't just add `parserOptions.sourceType` to my config. eslint won't run if I do).

There also seems to be an issue with imports being inserted inline. If I change the plugin code to use the new version of the config, so it uses esmodule imports, I end up with a few cases similar to this:
```
<Form onSubmit={import { describe, expect, it, jest } from '@jest/globals';jest.fn()}>{({formProps}) => <form {...formProps}>{children}</form>}</Form>
```
where the import is inserted directly where it's used, rather than at the top of the file.

ejzimmer:
ESLint's new config structure moves `sourceType` from `parserOptions` to `languageOptions` . This change supports that.

(I couldn't find any other references to `sourceType` in the code. It exists in tests, but they're all using the new structure)

Resolves #1634


G-Rath (changes requested):
would you mind exploring add a test for this? it could be hard with legacy vs flat config, so feel free bail if you can't get it working, and I can take a look later this week

ejzimmer:
> would you mind exploring add a test for this? it could be hard with legacy vs flat config, so feel free bail if you can't get it working, and I can take a look later this week

I've added what I think a test case should look like, but it fails due to the type checking (both TypeScript, and the run-time check of the config). I'm not sure why though, because it looks like the test utils are converting the flat config to the old config?

G-Rath:
good point, though actually the flat compat tester is meant to convert legacy config to flat config when testing on v9 and our suite is passing with full coverage across all ESLint versions meaning this should already be covered.

Looking into this manually, I'm not able to actually produce a bug as ESLint internally maps `sourceType` to `context.parserOptions.sourceType` - I think it would be best if you could provide a min. reproduction of the issue(s) you're running into on the issue, and we can go from there.

ejzimmer:
Minimal reproduction is https://github.com/ejzimmer/prefer-importing-jest-globals-test

It looks like the typescript parser might be the issue. It works fine if I remove `parser: tsParser` from the config, but breaks with it in.

G-Rath:
ok yup this is a thing with `@typescript-eslint/parser`, and that's why you're having trouble with a test 🙂 

<details>
<summary>Here's a "correct" set of tests that fails without this patch</summary>

```
new RuleTester({
}).run('prefer-importing-jest-globals: typescript edition', rule, {
  valid: [],
  invalid: [
    {
      code: dedent`
        import describe from '@jest/globals';
        describe("suite", () => {
          test("foo");
          expect(true).toBeDefined();
        })
      `,
      output: dedent`
        import { describe, expect, test } from '@jest/globals';
        describe("suite", () => {
          test("foo");
          expect(true).toBeDefined();
        })
      `,
      parserOptions: { sourceType: 'module' },
      errors: [
        {
          endColumn: 7,
          column: 3,
          line: 3,
          messageId: 'preferImportingJestGlobal',
        },
      ],
    },
    {
      code: dedent`
        const {describe} = require('@jest/globals');
        describe("suite", () => {
          test("foo");
          expect(true).toBeDefined();
        })
      `,
      output: dedent`
        const { describe, expect, test } = require('@jest/globals');
        describe("suite", () => {
          test("foo");
          expect(true).toBeDefined();
        })
      `,
      parserOptions: { sourceType: 'script' },
      errors: [
        {
          endColumn: 7,
          column: 3,
          line: 3,
          messageId: 'preferImportingJestGlobal',
        },
      ],
    },
  ],
});
```
</details>

ejzimmer:
Oh, fantastic, thanks. I don't have any more time this week, but I'll have a look at this first thing next week

ejzimmer:
> ok yup this is a thing with `@typescript-eslint/parser`, and that's why you're having trouble with a test 🙂
> Here's a "correct" set of tests that fails without this patch

I've updated the tests, and everything looks ok, but the test coverage is failing. I assume this is because the newer version of eslint never hit the second branch of the ternary that I added? I'm not sure we can do anything about that...

src/rules/__tests__/prefer-importing-jest-globals.test.ts
```diff
@@ -1,16 +1,10 @@
 import dedent from 'dedent';
 import rule from '../prefer-importing-jest-globals';
-import { FlatCompatRuleTester as RuleTester, espreeParser } from './test-utils';
+import { FlatCompatRuleTester as RuleTester } from './test-utils';
 
-const ruleTester = new RuleTester({
```
G-Rath, line 5: I'm pretty sure when I was working on this all of these tests were passing regardless of this change - overall, we should be retaining the existing test structure, and just adding a new typescript parser based one.

This is a good example of what I mean:

https://github.com/jest-community/eslint-plugin-jest/blob/f969f92a0a3b72df5f87d898d9e3249474fcf8e5/src/rules/__tests__/prefer-to-contain.test.ts#L214-L228

G-Rath (changes requested):

src/rules/__tests__/prefer-importing-jest-globals.test.ts
ejzimmer: Ah, right, I get it now I think!

G-Rath (approved):
perfect, thanks!

ejzimmer:
> perfect, thanks!

Thanks for all your help on this one 💜

PR #1646, opened 2024-09-01
perf(prefer-importing-jest-globals): stop collecting import specifiers for no reason

eryue0220:
This pr is to remove unused `ImportDeclaration` check.

### DIFF
--- a/src/rules/prefer-importing-jest-globals.ts
+++ b/src/rules/prefer-importing-jest-globals.ts
@@ -60,19 +60,10 @@ export default createRule({
   defaultOptions: [{ types: allJestFnTypes }],
   create(context) {
     const { types = allJestFnTypes } = context.options[0] || {};
-    const importedFunctionsWithSource: Record<string, string> = {};
     const functionsToImport = new Set<string>();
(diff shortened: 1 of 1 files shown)


G-Rath:
Can you explain why this isn't needed? It must have been being covered by at least one test so it can't have been unused, though the fact that our suite still passes would suggest it's not needed but it would be good to understand why

eryue0220:
> Can you explain why this isn't needed? It must have been being covered by at least one test so it can't have been unused, though the fact that our suite still passes would suggest it's not needed but it would be good to understand why

I have two thoughts about that, the variables `importedFunctionsWithSource` only collect import elements from import statement, but what it really check is [here](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-importing-jest-globals.ts#L84), another is performance, if we don't use that, but it still iterates the AST nodes to collect something, it's a little worthless.

G-Rath:
Oh I see, we're only ever writing to `importedFunctionsWithSource` - we never actually read from it; nice catch!

G-Rath (approved):

Issue #1660, opened 2024-10-23
[Feature]: Compatibility with the latest version of eslint


nayounsang:
Hello.
The contents of the docs are not compatible with the latest eslint version. I think it would be good to write about the latest eslint version for confusion.

nayounsang:
This is temporary solution for my case.
```js
// eslint.config.mjs
import jestPlugin from "eslint-plugin-jest";

// ...

  {
    files: ["**/*.spec.{js,ts}"],
    plugins: {
      jest: jestPlugin,
    },
    languageOptions: {
      globals: {
        describe: "readonly",
        it: "readonly",
        test: "readonly",
        expect: "readonly",
        beforeEach: "readonly",
        afterEach: "readonly",
        beforeAll: "readonly",
        afterAll: "readonly",
        jest: "readonly",
      },
    },
    rules: {
      ...jestPlugin.configs["flat/recommended"].rules,
    },
  },

// ...

```

G-Rath:
While the start of our readme only mentions `.eslintrc`, we do have documentation for flat config, including a full example under the ["Running rules only on test-related files" section of the readme](https://github.com/jest-community/eslint-plugin-jest?tab=readme-ov-file#running-rules-only-on-test-related-files), along with more examples in the ["Shareable configurations" section](https://github.com/jest-community/eslint-plugin-jest?tab=readme-ov-file#shareable-configurations).

From what I can tell just about all of our documentation especially at the start is pretty agnostic to the config format being used - probably the main one that isn't that might be worth updating is the first part of the "Usage" section, but I think the [ESLint docs on configuring plugins](https://eslint.org/docs/latest/use/configure/plugins) should make it pretty obvious what you need to do for that in flat config.

mayronH:
@nayounsang I think the most basic config is just:

```js
// eslint.config.mjs

import pluginJs from '@eslint/js'
import pluginJest from 'eslint-plugin-jest'
import globals from 'globals'

export default [
  { files: ['**/*.js'], languageOptions: { sourceType: 'commonjs' } },
  { languageOptions: { globals: globals.browser } },
  pluginJs.configs.recommended,

  pluginJest.configs['flat/recommended'],
]

```

PR #1661, opened 2024-10-23
Es lint 9 docs

nayounsang:
#1660

# What I do
- I write for basically config in es-lint 9 version.
- I make file about global variables json file for es-lint 9 version.

# Describe
- I write docs with `If you are using `eslint` version 9 or higher,` about env & plugin part.
- In the latest es-lint, value of global variable must be `readonly` or `writeable`. So I make the json file.

# Additional
- I don't know where to write the part about setting up rules. The related API already exists.
- Currently I have a very basic `jest es-lint` setup. Are there any other places where something goes wrong? I don't know because I haven't experienced it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ doing:
 ```
 
 If you are using `eslint` version 9 or higher,
+
 ```js
 import jestPlugin from "eslint-plugin-jest";
 
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@ can omit the `eslint-plugin-` prefix:
 If you are using `eslint` version 9 or higher,
 
 ```js
+import jestPlugin from "eslint-plugin-jest";
+
+// ...
+
 plugins: {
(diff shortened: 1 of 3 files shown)


G-Rath (changes requested):
blocking until we've [figured out what changes should actually be made, if any](https://github.com/jest-community/eslint-plugin-jest/issues/1660#issuecomment-2433041678)

nayounsang:
> blocking until we've [figured out what changes should actually be made, if any](https://github.com/jest-community/eslint-plugin-jest/issues/1660#issuecomment-2433041678)

Ah, then I guess we can close this PR. However, currently eslint docs are set at the top in [eslint.config.js format](https://eslint.org/docs/latest/use/configure/configuration-files) not `.eslintrc`.
So, It would be good to expose the example introduced in the comments of the issue in the `README` in a more accessible place.
What do you think of this?

Issue #1469, opened 2023-11-16
Feature: Generate typings for rules etc.


pauliesnug:
Currently we package with Babel to pure JS, but it would be awesome if we could include some `.d.ts` files with the official build for programmatic, auto-completion, and extension use. This shouldn't be too difficult considering it's already written in TypeScript.

G-Rath:
Can you provide some real-world examples of where the types would actually be used? (beyond that someone could "one day" want to extend a rule, since I don't think anyone is doing that currently with our rules) My understanding is that ESLint tooling is still not in a good position to properly utilize typings for rules and that that doesn't change with the new flat config format.

The downsides with shipping types is that they add a bit to the package size and increase the surface area of our build infrastructure since Babel doesn't support generating types - neither are particularly costly but relative to not having an actual common real world usecase they're pretty expensive...

pauliesnug:
> Can you provide some real-world examples of where the types would actually be used? (beyond that someone could "one day" want to extend a rule, since I don't think anyone is doing that currently with our rules) My understanding is that ESLint tooling is still not in a good position to properly utilize typings for rules and that that doesn't change with the new flat config format.

@G-Rath Sure! For this issue to be sparked, I was trying to implement a way to use `eslint-plugin-jest` with [`eslint-define-config`](https://github.com/antfu/eslint-define-config) which is a really awesome strongly-typed way to configure multiple ESLint plugins in a modular way. All of this is to say that without TypeScript types it's impossible to use the features of TypeScript and in some cases impossible to even use the plugin in the first place. Even just an interface with the structure of:
```ts
export interface RuleOptions {
       /**
        * this is an example of a description for the rule.
        * 
        * @see [rule-name](https://docs.example.com/rule-name.md)
        */
      'jest/rule-name': string;
}
```
would make all the difference. This would produce a much lesser impact. Alternatively, types could be provided in a separate package if bundle size is a concern.

SimenB:
I think it would be fine to ship types. We already run `tsc`, so it'd just be a quick config change to have it emit definition files. Are we certain the generated types helps with `eslint-define-config`? Seems to me like we need custom code to work with them (https://github.com/eslint-types/eslint-define-config/blob/main/README.md#want-to-support-your-own-plugin)

I don't really care about bundle size, but we could also bundle up the types into a single file like we do in Jest (via https://www.npmjs.com/package/@microsoft/api-extractor) in order to minimise the impact

[...]

PR #1667, opened 2024-11-04
feat: add types to package

mskelton:
Fixes https://github.com/jest-community/eslint-plugin-jest/issues/1469

I opted to generate the types manually with a hard coded `index.d.ts` file given that the types are very straightforward and the result works much better with typescript-eslint in my experience vs simply relying on `tsc` generated types.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,28 @@
+import type { Linter, Rule } from 'eslint';
+
+declare const plugin: {
+  meta: {
+    name: string;
+    version: string;
+  };
+  environments: {
+    globals: {
+      globals: {
+        [key: string]: boolean;
+      };
+    };
+  };
+  configs: {
+    all: Linter.LegacyConfig;
+    recommended: Linter.LegacyConfig;
+    style: Linter.LegacyConfig;
+    'flat/all': Linter.FlatConfig;
+    'flat/recommended': Linter.FlatConfig;
+    'flat/style': Linter.FlatConfig;
+  };
+  rules: {
+    [key: string]: Rule.RuleModule;
+  };
+};
+
+export = plugin;
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -133,5 +133,11 @@ module.exports = {
         'import/no-commonjs': 'off',
       },
     },
+    {
+      files: ['*.d.ts'],
+      rules: {
+        strict: 'off',
+      },
+    },
   ],
 };
--- a/package.json
+++ b/package.json
@@ -15,9 +15,11 @@
     "url": "jkimbo.com"
   },
   "main": "lib/index.js",
+  "types": "index.d.ts",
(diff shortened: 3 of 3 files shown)


index.d.ts
```diff
@@ -0,0 +1,28 @@
...
+    globals: {
+      globals: {
+        [key: string]: boolean;
+      };
+    };
+  };
+  configs: {
+    all: Linter.Config;
+    recommended: Linter.Config;
+    style: Linter.Config;
```
G-Rath, line 18: issue: we should use this type to be compatible with `@types/eslint@9`

```suggestion
    all: Linter.LegacyConfig;
    recommended: Linter.LegacyConfig;
    style: Linter.LegacyConfig;
```

index.d.ts
```diff
@@ -0,0 +1,28 @@
...
+    'flat/all': Linter.FlatConfig;
+    'flat/recommended': Linter.FlatConfig;
+    'flat/style': Linter.FlatConfig;
+  };
+  rules: {
+    [key: string]: Rule.RuleModule;
+  };
+};
+
+export default plugin;
```
G-Rath, line 28: issue: this is the correct export type for how the package is published:

```suggestion
export = plugin;
```

index.d.ts
```diff
@@ -0,0 +1,28 @@
...
+
+declare const plugin: {
+  meta: {
+    name: string;
+    version: string;
+  };
+  environments: {
+    globals: {
+      globals: {
+        [key: string]: boolean;
```
G-Rath, line 11: note: ideally this should be `Linter.GlobalConf`, but that was introduced in `@types/eslint@9`

.eslintignore
```diff
@@ -2,3 +2,4 @@ coverage/
 lib/
 !.eslintrc.js
 src/rules/__tests__/fixtures/
+index.d.ts
```
G-Rath, line 5: issue: we shouldn't be ignoring this file from formatting

G-Rath (changes requested):
thanks, a great start that just needs a few tweaks

mskelton:
@G-Rath Updated per your suggestions

index.d.ts
```diff
@@ -0,0 +1,29 @@
+// eslint-disable-next-line strict
```
G-Rath, line 1: suggestion: can you disable this via our ESLint config?

mskelton: Fixed

G-Rath (approved):
awesome, thanks!

G-Rath:
Apparently there's a recent issue with `semantic-release` that means it cannot find pull requests when looking to post its "this was released in ..." message, which should hopefully be fixed in the next few days (https://github.com/semantic-release/github/pull/960) 😅 

The release was still successful, so this has gone out in v28.9.0

PR #1675, opened 2024-11-20
feat: cleanup main types

MichaelDeBoey:
Follow-up of #1667

CC/ @G-Rath @mskelton

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 import type { TSESLint } from '@typescript-eslint/utils';
 
-type SupportedConfigs = 'all' | 'recommended' | 'style';
+type JestPluginConfigName = 'all' | 'recommended' | 'style';
 
 declare const plugin: {
   meta: {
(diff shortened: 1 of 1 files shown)


G-Rath:
I don't really see this as strictly better - the types as they are pretty straightforward and maintainable as they are imo, and I in particular like having the configs written out individually to make it clear what is available without having to mentally unroll some type logic.

At the very least, this is not a "feat" or a "fix"

mskelton:
I would agree with @G-Rath on this not necessarily improving the types greatly.

MichaelDeBoey:
@G-Rath @mskelton We're also using TSESLint's types here, which is a dependency we already have

If we use ESLint's dependency, we would need to add `@types/eslint` to the `dependencies` in `package.json` to work correctly

G-Rath:
@MichaelDeBoey you'll already need `@types/eslint` as a dependency if you're type-checking `eslint.config.js` on v8, and ESLint v9 ships with its own types that are compatible with these ones, so that shouldn't be an issue

G-Rath:
I'm going to close this due to age, and since it's not clear that this is actually an improvement

PR #1679, opened 2024-12-03
fix!: remove `no-alias-methods` from `style` config

### LINKED ISSUE #1678: `no-alias-methods` error in recommended, warn in style
`jest/no-alias-methods` is [set to `error`](https://github.com/jest-community/eslint-plugin-jest/blob/483cfb4a051290ee9c07ef7a4d4b37e94d8a0ad9/src/index.ts#L35) in the `recommended` config (since #1221, August 2022), but [set to `warn`](https://github.com/jest-community/eslint-plugin-jest/blob/483cfb4a051290ee9c07ef7a4d4b37e94d8a0ad9/src/index.ts#L56) in the `style` config (since https://github.com/jest-community/eslint-plugin-jest/commit/14cb81579c23a3be34e5e0a989f516a398ec33cb, October 2019).

I think this doesn't make sense, since the `style` config is supposed to be applied on top of the `recommended` config, but doing that decreases the rule severity.

Likely it was just missed in #1221 to remove `jest/no-alias-methods` from the `style` config.

FloEdelmann:
Fixes #1678.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ Manually fixable by
 | [expect-expect](docs/rules/expect-expect.md)                                         | Enforce assertion to be made in a test body                               |     | ✅  |     |     |
 | [max-expects](docs/rules/max-expects.md)                                             | Enforces a maximum number assertion calls in a test body                  |     |     |     |     |
 | [max-nested-describe](docs/rules/max-nested-describe.md)                             | Enforces a maximum depth to nested describe calls                         |     |     |     |     |
-| [no-alias-methods](docs/rules/no-alias-methods.md)                                   | Disallow alias methods                                                    | ✅  | 🎨  | 🔧  |     |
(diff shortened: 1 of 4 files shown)


G-Rath:
While technically a fix, changing configs is usually considered a breaking change for eslint plugins and we're expecting to do a new major "soon" so I don't think we can land this for now

FloEdelmann:
> changing configs is usually considered a breaking change

Even removing rules from configs? It will technically result in less lint errors, so that shouldn't be breaking. But well, most people probably extend `recommended`+`style`, so that does increase the rule severity from `warn` to `error`, so it might indeed break some builds.

I'll change the commit message to include a `!` and we can just leave this PR open until you're comfortable with merging it.

G-Rath:
This has been addressed in v29 which was just released - I had intended to attribute the commit to you but the release got mucked up so sorry about that

PR #1681, opened 2024-12-04
feat(unbound-methods): ignore functions passed to `jest.mocked`

### LINKED ISSUE #1596: [unbound-method]: don't report unbound methods passed to `jest.mocked`
```ts
import foo from './foo'

jest.mock('./foo')

jest.mocked(foo.bar).mockResolvedValue(baz)
//          ^^^^^^^ jest/unbound-method reports an error here

expect(foo.bar).toHaveBeenCalled() // no error here
```

Since `jest.mocked` will be called with mock functions, it should not make a difference whether it is bound or not, right? So I'd suggest to not report an error for unbound methods passed to `jest.mocked`, just like already happens when passing it to `expect`.

FloEdelmann:
Fixes #1596.

### DIFF
--- a/src/rules/unbound-method.ts
+++ b/src/rules/unbound-method.ts
@@ -79,6 +79,16 @@ export default createRule<Options, MessageIds>({
       ...baseSelectors,
       MemberExpression(node: TSESTree.MemberExpression): void {
         if (node.parent?.type === AST_NODE_TYPES.CallExpression) {
+          if (
+            node.parent.callee.type === AST_NODE_TYPES.MemberExpression &&
+            node.parent.callee.object.type === AST_NODE_TYPES.Identifier &&
+            node.parent.callee.object.name === 'jest' &&
+            node.parent.callee.property.type === AST_NODE_TYPES.Identifier &&
+            node.parent.callee.property.name === 'mocked'
+          ) {
+            return;
+          }
+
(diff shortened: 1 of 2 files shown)


src/rules/unbound-method.ts
```diff
@@ -79,6 +79,16 @@ export default createRule<Options, MessageIds>({
...
         if (node.parent?.type === AST_NODE_TYPES.CallExpression) {
+          if (
+            node.parent.callee.type === AST_NODE_TYPES.MemberExpression &&
+            node.parent.callee.object.type === AST_NODE_TYPES.Identifier &&
+            node.parent.callee.object.name === 'jest' &&
+            node.parent.callee.property.type === AST_NODE_TYPES.Identifier &&
+            node.parent.callee.property.name === 'mocked'
+          ) {
+            return;
+          }
```
G-Rath, line 90: you leverage `parseJestFnCall` for this - you can see an example of how to use the result for this in `no-confusing-set-timeout`:

```
function isJestSetTimeout(jestFnCall: ParsedJestFnCall) {
  return (
    jestFnCall.type === 'jest' &&
    isIdentifier(jestFnCall.members[0], 'setTimeout')
  );
}
```

G-Rath (approved):
thanks!

PR #1688, opened 2024-12-20
feat(valid-expect): allow calling `expect` with no arguments

sonnyp:
This is particularly useful for jest-extended

* https://jest-extended.jestcommunity.dev/docs/matchers/pass
* https://jest-extended.jestcommunity.dev/docs/matchers/fail


G-Rath:
thanks, I'm happy with this but will probably wait until after the holidays to land this to give the link checker a chance to come right (seems to be unhappy with https://eslint.org/docs/latest/use/core-concepts#rule-suggestions which looks bonked anyway, and is generated by `eslint-doc-generator` so I'm hopeful it'll get resolved upstream)

G-Rath (approved):

PR #1691, opened 2024-12-30
fix(padding-around-test-blocks): update description

y-hsgw:
I have fixed the `padding-around-test-blocks` rule documentation.

### DIFF
--- a/src/rules/padding-around-test-blocks.ts
+++ b/src/rules/padding-around-test-blocks.ts
@@ -27,6 +27,6 @@ export const config = [
 
 export default createPaddingRule(
   __filename,
-  'Enforce padding around `test` blocks',
+  'Enforce padding around `test` and `it` blocks',
   config,
 );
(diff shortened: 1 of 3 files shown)


src/rules/padding-around-test-blocks.ts
```diff
@@ -27,6 +27,6 @@ export const config = [
 
 export default createPaddingRule(
   __filename,
-  'Enforce padding around afterAll blocks',
+  'Enforce padding around `test` blocks',
```
G-Rath, line 30:
```suggestion
  'Enforce padding around `test` and `it` blocks',
```

G-Rath (changes requested):
thanks - would you mind updating this to include `it` as well?

src/rules/padding-around-test-blocks.ts
y-hsgw: I fixed in [5f7c7b0](https://github.com/jest-community/eslint-plugin-jest/pull/1691/commits/5f7c7b02267d738dcd047a4793b0645447a25aec).

G-Rath (approved):

PR #1697, opened 2025-01-14
chore: downgrade `markdown-link-check` to known working version and make it louder

### LINKED ISSUE #1696: Markdown link check is failing
https://github.com/jest-community/eslint-plugin-jest/actions/workflows/markdown-link-check.yml

@G-Rath do you have any idea what's going on here? 😅  We might wanna use (and possibly migrate to the option recommended in its readme?)

Image

No context for _why_ it fails... 🤔 

Running it locally hangs forever (or, at least 5 minutes, at which point I killed it)

G-Rath:
[It seems that something is broken with the latest version(s)](https://github.com/tcort/markdown-link-check/issues/369) when using a config file; in the long-run it might be best to migrate to another tool but for now this should get our CI passing.

I've also removed the `-q(uiet)` flag in CI since logs are free and that could make it easier to debug in future

Resolves #1696

### DIFF
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -35,4 +35,3 @@ jobs:
         run:
           find . -name "*.md" | grep -v node_modules | grep -v CHANGELOG.md |
           xargs -n 1 yarn markdown-link-check -c markdown_link_check_config.json
-          -q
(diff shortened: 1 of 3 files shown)


SimenB (approved):
ah, great stuff! thank you! 😀

PR #1710, opened 2025-02-25
chore: add missing `'use strict'` directive

Fdawgs:
Added a missing `'use strict'` directive to the only CJS file in this repo without one.

Test failures unrelated, occurring on the main branch.


G-Rath:
Thanks, it looks like this file wasn't being linted - I've opened #1712 fixing that; thankfully this was the only actionable violation (which isn't surprising)

G-Rath (approved):

PR #1711, opened 2025-02-25
chore: ensure types are available for `semver`

G-Rath:
Apparently we were depending on this implicitly (which surprises me), and #1708 updated `conventional-changelog-writer` to a version that no longer has this as a dependency, causing CI to fail in some cases because only some versions of `@typescript-eslint` depend on them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "@types/eslint": "^8.4.6",
     "@types/jest": "^29.0.0",
     "@types/node": "^16.0.0",
+    "@types/semver": "^7.5.8",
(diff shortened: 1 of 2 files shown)


SimenB:
Why did it automerge with failing CI 🤔

G-Rath:
I don't know - it's done that before; my first guess would be that the branching rules are not setup to block it's merging (sorry 😅)

SimenB:
I tried enabling merge queue now, with mandatory status checks. let's see if that works 😀

SimenB:
#1715 enabled automerge and merge queue stuff, so 👍

PR #1736, opened 2025-05-25
fix(no-large-snapshots): use a far better message for when an unexpected snapshot is found

ulrichstark:
The incorrect message was introduced by #278

### DIFF
--- a/src/rules/no-large-snapshots.ts
+++ b/src/rules/no-large-snapshots.ts
@@ -80,7 +80,7 @@ export default createRule<[RuleOptions], MessageId>({
     },
     messages: {
       noSnapshot:
-        'Expected to not encounter a Jest snapshot but was found with {{ lineCount }} lines long',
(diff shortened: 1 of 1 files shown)


src/rules/no-large-snapshots.ts
```diff
@@ -79,7 +79,8 @@ export default createRule<[RuleOptions], MessageId>({
       description: 'Disallow large snapshots',
     },
     messages: {
-      noSnapshot: '`{{ lineCount }}`s should begin with lowercase',
+      noSnapshot:
+        'Expected to not encounter a Jest snapshot but was found with {{ lineCount }} lines long',
```
G-Rath, line 83:
```suggestion
        'Expected to not encounter a Jest snapshot but one was found that is {{ lineCount }} lines long',
```

G-Rath (approved):
wow, I can't believe that has been around for so long 😅

PR #1764, opened 2025-06-15
ci: have `renovate` skip updating `@types/node`

G-Rath:
We want to use `@types/node` for the _oldest_ version of Node that we support


G-Rath:
According to the docs Renovate should have validated this change since it was on `renovate/reconfigure` but it doesn't look to have worked 🤷

PR #1771, opened 2025-06-18
feat: next major

### LINKED ISSUE #1743: Next major
With Jest v30 around the corner and dropping support for Node v16, I think it's time we look at doing our next major.

So far I've got:
  - drop Node v16 support
  - remove `no-alias-methods` from `style` config since it's already in `recommended`
  - drop support for `@typescript-eslint/utils` v6 and v7
  - drop support for `eslint` v7 (I don't think we should drop v8 support yet)

G-Rath:
Resolves #1743

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -73,15 +73,8 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [18.x, 20.x, 22.x, 23.x, 24.x]
-        eslint-version: [7, 8, 9]
+        eslint-version: [8, 9]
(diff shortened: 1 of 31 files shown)


G-Rath:
um ok well that seems like a bug

G-Rath:
I would assume this has happened because of #1585 since that's the last thing that touched this - I think to be on the safe side I'm going to just land this as-is and mark v29.0.0 as a not-prerelease in GH, and then maybe land another PR to ensure the `npm` version published is no longer marked as `next` since I don't think that'll happen when I merge this due to the "skip ci" tag

(I was planning to wait a week or so, but oh well)

PR #1790, opened 2025-07-01
ci: get smoketest workflow passing again

G-Rath:
Currently `npm` errors as we've technically got peer dependency violations due to [`jest-runner-eslint` not officially supporting Jest v30](https://github.com/jest-community/jest-runner-eslint/pull/291) (or ESLint v9 for that matter), so I've finally switched us to using Yarn to setup this job.

This has also revealed that we might not yet be able to use v2 of the repositories package so I've downgraded that for now but will revisit once the job is working again

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -30,6 +30,7 @@ jobs:
           yarn install
           yarn build
           yarn link
+          ln -s ../ node_modules/eslint-plugin-jest
       - uses: AriPerkkio/eslint-remote-tester-run-action@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -22,12 +22,12 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-node@v4
         with:
-          node-version: 16
+          node-version: 20
```
SimenB, line 25:
```suggestion
          node-version: lts/*
```

PR #1842, opened 2025-11-03
docs(valid-title): add documentation for `ignoreTypeOfTestName` option

edg2s:


G-Rath (approved):
thanks, this is good as-is but if you have the time it'd be good to have this also document `ignoreTypeOfDescribeName`.

Happy if you just want to say no to that though, as I can do it in a follow up 🙂

edg2s:
> thanks, this is good as-is but if you have the time it'd be good to have this also document `ignoreTypeOfDescribeName`.
> 
> Happy if you just want to say no to that though, as I can do it in a follow up 🙂

That is already documented.

G-Rath (approved):
thanks!

PR #1845, opened 2025-11-08
feat: create new `valid-mock-module-path` rule

### LINKED ISSUE #1584: New rule to validate path given to `jest.mock`
It would be really useful to have a lint rule that tells you when the path supplied to `jest.mock` does not exist. I would envision the rule working like this:

# Disallow mocking of module paths that do not exist

When using `jest.mock` and `jest.doMock`, the path supplied must be an module or local file that exists.

## Rule details

This rule triggers an error if the path supplied to `jest.mock` or `jest.doMock` does not exist.

The following patterns are considered errors:

```ts
// Module that cannot be found
jest.mock('some-module-not-in-package-json');

// Local file that cannot be found
jest.mock('../../this/path/does/not/exist');
```

The following patterns not considered errors:

```ts
// Module that can be found
jest.mock('some-module-in-package-json');

// Local file that can be found
jest.mock('../../this/path/really/does/exist');
```

hainenber:
Resolves #1584 

Add new rule to verify whether `jest.mock` in 1st argument, intended for a module and/or a local file, to exists or not. 

For local files, I've limited file checks for `.ts.` and `.js` files only as I don't want to increase the complexity further with combination of `.cjs`, `.mts` and alike.

New unit tests added.

### DIFF
--- a/src/rules/__tests__/valid-mock-module-path.test.ts
+++ b/src/rules/__tests__/valid-mock-module-path.test.ts
@@ -157,20 +157,18 @@ describe('valid-mock-module-path', () => {
 
       /* istanbul ignore if */
       if (usingFlatConfig) {
-        linter.verify(
-          'jest.mock("./fixtures/module")',
-          [
-            {
-              files: [__filename],
-              plugins: {
-                jest: {
-                  rules: { 'valid-mock-module-path': mockUnexpectedError() },
-                },
+        linter.verify('jest.mock("./fixtures/module")', [
+          {
+            plugins: {
+              jest: {
+                rules: { 'valid-mock-module-path': mockUnexpectedError() },
               },
             },
-          ],
-          __filename,
-        );
+            rules: {
+              'jest/valid-mock-module-path': 'error',
+            },
+          },
+        ]);
 
         return;
       }


src/rules/valid-mocked-module-path.ts
```diff
@@ -0,0 +1,99 @@
...
+          !(
+            isSupportedAccessor(property) &&
+            ['mock', 'doMock'].includes(getAccessorValue(property))
+          )
+        ) {
+          return;
+        }
+
+        const [nameNode] = node.arguments;
+        const moduleName = findModuleName(nameNode);
```
G-Rath: nit: likewise, this doesn't need its own variable

```suggestion
        const moduleName = findModuleName(node.arguments[0]);
```

src/rules/valid-mocked-module-path.ts
```diff
@@ -0,0 +1,99 @@
...
+                } catch {
+                  return false;
+                }
+              });
+
+            if (!hasPossiblyModulePaths) {
+              throw { code: 'MODULE_NOT_FOUND' };
+            }
+          } else {
+            require.resolve(moduleName.value);
```
G-Rath: we can do an early return to avoid this `else`, and if you invert the condition we can deindent the other (larger) code path

G-Rath (changes requested):
looks good, just a few bits to cleanup - I'm also on the fence about the name: I think "mock" might be slightly better than "mocked" because we're dealing about something that is going to be mocked, rather something that has already been mocked, but then I'm not sure if it should be `valid-mock-module-path` or `valid-module-mock-path` 😅 

also please revert the changes re link checking - I've landed #1846 which addresses the issue, and we don't need to update the package

hainenber:
I think I'll go with `valid-mock-module-path`. It kinda resonates well with me :D

hainenber:
hi @G-Rath, I've addressed your review comments. Do lmk if there are further concerns :D

src/rules/valid-mock-module-path.ts
```diff
@@ -0,0 +1,121 @@
...
+    ],
+  },
+  defaultOptions: [
+    {
+      moduleFileExtensions: ['.js', '.ts', '.tsx', '.jsx', '.json'],
+    },
+  ],
+  create(
+    context,
+    [{ moduleFileExtensions = ['.js', '.ts', '.tsx', '.jsx', '.json'] }],
```
G-Rath: can we add a few more tests for this option - we should have one for `.json`, and it would be good to have one that adds e.g. `.css` which'll serve to show both arbitrary extensions work and that it replaces rather than extends the default

(I know you've already got at least one test covering the latter, but its still nice to have another)

docs/rules/valid-mock-module-path.md
```diff
@@ -0,0 +1,73 @@
...
+      "moduleFileExtensions": [".tsx", ".ts"]
+    }
+  ]
+}
+```
+
+### `moduleFileExtensions`
+
+This array option controls which file extensions the plugin checks for
+existence. Valid values are:
```
G-Rath:
```suggestion
existence. The default extensions are:
```

docs/rules/valid-mock-module-path.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Options
+
+```json
+{
+  "jest/valid-mock-module-path": [
+    "error",
+    {
+      "moduleFileExtensions": [".tsx", ".ts"]
```
G-Rath:
```suggestion
      "moduleFileExtensions": [".js", ".ts", ".jsx", ".tsx", ".json"]
```

hainenber:
hi @G-Rath, I've refactored per the 2nd round of review, including these items:

- I deleted away `ENOENT` err check since `ENOENT` errors are captured in inner `try-catch` of the `hasPossiblyModulePaths` chained function. 
- There are 2 newly added test suites to cover raised concerns.
- Re-casted caugh error to `unknown`
- Refactored to move `context.report` outside of `try-catch` block.

G-Rath (approved):
looks good to me! it'd be great if you could add a test using mocking for the error path, but not critical so let me know if you don't have time or run into trouble with it

PR #1848, opened 2025-11-12
feat: create new `pair-to-have-been-called-assertions` rule

### LINKED ISSUE #1820: Rule proposal: `jest/prefer-strict-mock-assertions`
### Problem
Developers often use `toHaveBeenCalledWith()` without verifying the call count, leading to tests that pass when a mock is called multiple times unexpectedly.

### Solution
Require `toHaveBeenCalledTimes()` when using `toHaveBeenCalledWith()` to ensure complete mock validation.

### Examples

```javascript
// ❌ Potentially flaky - passes if called 1 + times
expect(mockFn).toHaveBeenCalledWith(expectedArg)

// ✅ Explicit and safe - ensures exact call count
expect(mockFn).toHaveBeenCalledTimes(1)
expect(mockFn).toHaveBeenCalledWith(expectedArg)
```

### Benefits
- Prevents false positives in tests
- Makes test intentions explicit
- Improves test reliability and maintainability
- Follows testing best practices

Are you willing to implement this rule? Yes

nicolasgillot:
## Summary

This PR introduces a new ESLint rule `pair-to-have-been-called-assertions` that enforces using `toHaveBeenCalledTimes()` or `toBeCalledTimes()` whenever using matchers like `toHaveBeenCalledWith()`, `toBeCalledWith()`, `toHaveBeenNthCalledWith()`, `toBeNthCalledWith()`, `toHaveBeenLastCalledWith()`, or `toBeLastCalledWith()`.

### Why?

When testing mock functions, developers often use `toHaveBeenCalledWith()` to verify that a function was called with specific arguments. However, without also checking the call count, the test can pass even if the mock was called more times than expected, potentially masking bugs.

### What does this rule do?

- Requires pairing `toHaveBeenCalledWith()` (and similar matchers) with `toHaveBeenCalledTimes()`
- Provides auto-fix that adds the missing `toHaveBeenCalledTimes(1)` assertion
- Detects contradictory assertions (e.g., `toHaveBeenCalledTimes(0)` with `toHaveBeenCalledWith()`)
- Works with all call-checking matchers: `toHaveBeenCalledWith`, `toBeCalledWith`, `toHaveBeenNthCalledWith`, `toBeNthCalledWith`, `toHaveBeenLastCalledWith`, `toBeLastCalledWith`

### Examples

❌ Incorrect:
```js
expect(mockFn).toHaveBeenCalledWith('arg');
```

✅ Correct:
```js
expect(mockFn).toHaveBeenCalledTimes(1);
expect(mockFn).toHaveBeenCalledWith('arg');
```

Resolves #1820


docs/rules/pair-to-have-been-called-assertions.md
```diff
@@ -0,0 +1,142 @@
...
+expect(mockFn).toHaveBeenLastCalledWith('last');
+
+// Using toBeNthCalledWith without call count
+expect(mockFn).toBeNthCalledWith(2, 'second');
+
+// Using toBeLastCalledWith without call count
+expect(mockFn).toBeLastCalledWith('last');
+```
+
+### Correct
```
G-Rath:
```suggestion
Examples of **correct** code with this rule:
```

docs/rules/pair-to-have-been-called-assertions.md
```diff
@@ -0,0 +1,142 @@
...
+// Using 'not' modifier doesn't require call count
+expect(mockFn).not.toHaveBeenCalledWith('arg');
+
+// Only checking call count is fine
+expect(mockFn).toHaveBeenCalledTimes(1);
+```
+
+## Additional Checks
+
+### Contradictory Assertions
```
G-Rath: At this point I think this is out of scope for this rule - can we remove it for now?

docs/rules/pair-to-have-been-called-assertions.md
```diff
@@ -0,0 +1,142 @@
...
+
+## When Not To Use It
+
+If you have a specific testing strategy where checking call counts is not
+necessary or you're only interested in verifying that a function was called with
+specific arguments at least once, you can disable this rule.
+
+However, it's generally recommended to keep this rule enabled as it helps catch
+bugs where functions are called more times than expected, leading to more robust
+and reliable tests.
```
G-Rath: I don't think this is particularly useful to have in this section, especially as this a rule that is enabled by default

```suggestion

```

G-Rath (changes requested):
looks like a good start - could you remove the `contradictoryCallTimes` logic and do any resulting simplifications, and review the `istanbul` comments as I'm pretty sure most should be coverable

G-Rath:
Closing due to age - happy to have this PR revived if someone wants to try build off it

PR #1851, opened 2025-11-13
Use ESLint `extends`

joshkel:
Since [March of this year](https://eslint.org/blog/2025/03/flat-config-extends-define-config-global-ignores/#bringing-back-extends), ESLint supports and [recommends](https://eslint.org/docs/latest/use/configure/configuration-files#extending-configurations) using `extends` for flat configurations (instead of having to manually spread the pieces of the configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,10 @@ With
 just import the plugin and away you go:
 
 ```js
+const { defineConfig } = require('@eslint/config');
 const pluginJest = require('eslint-plugin-jest');
 
-module.exports = [
+module.exports = defineConfig([
   {
     // update this to match your test files
     files: ['**/*.spec.js', '**/*.test.js'],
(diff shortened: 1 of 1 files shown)


G-Rath:
It looks like that is a feature of the `defineConfig` helper, not actually ESLint itself - personally I think I'd prefer to stick with what we have for now since the helpers are still evolving and it keeps this example consistent with others

joshkel:
Thanks for the correction, @G-Rath. I failed to realize that `extends` requires `defineConfig`.

`defineConfig` is recommended throughout the ESLint documentation; from what I've been able to follow of ESLint 9.x's development, it's quite stable. I've updated the remaining examples, but, of course, it's up to you all whether you want to update this plugin's docs to match.

G-Rath:
closing due to age

PR #1852, opened 2025-11-15
ci: pull Danger from pre-built docker images in GHCR

yoshi-taka:
CI speed improvement

https://github.com/jest-community/eslint-plugin-jest/actions/workflows/lint.yml

- GHCR images (ghcr.io/danger/danger-js) are prebuilt and cached on GitHub-hosted runners
- Completely removed Dockerfile-based builds


G-Rath:
Could you explain this change a bit more, ideally with some linkage to official documentation? My expectation is that GH already uses caching when appropriate for docker images, and doesn't build them every time.

I'm also not sure if this is supported by Renovate

yoshi-taka:
As you can see below, this PR runs noticeably faster, dropping from about one minute to twenty seconds compared with many other PRs.
image

If we look at what the previous action was doing, the below screenshot shows that it was performing a full Docker build.

image
e.g. https://github.com/jest-community/eslint-plugin-jest/actions/runs/19342384594/job/55333811424#step:2:1


GitHub Actions has a behavior that many people are not aware of:
when you reference a repository with `uses:`, Actions builds a Docker image from that repository’s Dockerfile.

yoshi-taka:
Renovate is able to update pinned Docker digests, as documented here: https://docs.renovatebot.com/docker/#digest-pinning

G-Rath:
I understand it aims to speed up CI, but it's not clear to me if this is official, stable, trust worthy, etc which is why ideally I'd like some docs or a blog post

yoshi-taka:
danger-js has been publishing images to GHCR for quite some time.
https://github.com/danger/danger-js/pkgs/container/danger-js

The code that handles the publishing is here:
https://github.com/danger/danger-js/blob/main/.github/workflows/publish_package.yml

yoshi-taka:
This is the closest official reference I could find: https://github.com/danger/danger/blob/master/CHANGELOG.md#932

SimenB:
happy to take a docker image over referring to the action 👍

SimenB:
I do see this in the logs, tho:

>Could not add a commit status, the GitHub token for Danger does not have access rights.

Miiiight be due to PR from someone without push access - I've never really wrapped my head around permissions for GHA

yoshi-taka:
@SimenB 
I think the more likely cause is token permission restrictions on fork PRs, rather than the switch to GHCR itself.
As far as I can tell, this PR only changes how Danger is referenced, from
`danger/danger-js@...` to `docker://ghcr.io/danger/danger-js@sha256:...`,
and does not change the workflow's permission model.

`lint.yml` still runs on `pull_request`, and Danger still relies only on
`GITHUB_TOKEN` for `issues: write`, `pull-requests: write`, and `statuses: write`.

So if the failure is around setting a status, posting a comment, or adding labels,
it seems more likely to be caused by token restrictions on fork PRs.

If useful, it might be worth trying the same change from a branch in this repository
to confirm whether the issue is specific to fork PR permissions.

G-Rath (approved):

PR #1862, opened 2025-11-25
fix(unbound-method): ignore inline use of `jest.mocked(...)`

### LINKED ISSUE #1800: [unbound-method]: should ignore accessing `jest.mocked().mock`
I noticed that the `unbound-method` rule reports the following code:
```ts
const parameter = jest.mocked(service.method).mock.calls[0][0];
```

However, the following variant is not reported:
```ts
const mockedMethod = jest.mocked(service.method);
const parameter = mockedMethod.mock.calls[0][0];
```

Using any `.mockXY()` methods is also fine (as expected since #1596):
```ts
jest.mocked(service.method).mockReturnValue();
```

Image

Reproduction: https://github.com/SchroederSteffen/jest-eslint-unbound-method-repro/blob/main/src/service.spec.ts

I think the mentioned case should NOT report an error.

eryue0220:
Fix: #1800


G-Rath (commented):
at this point I feel like this is an overly complicated solution - have you looked into why our existing logic for handling `jest.mocked` isn't catching this case?

eryue0220:
> at this point I feel like this is an overly complicated solution - have you looked into why our existing logic for handling `jest.mocked` isn't catching this case?

Sorry for the response, in this case the `jest.mocked` will be early returned in this [if statement](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/utils/parseJestFnCall.ts#L340). A more easier way is to add condition that is `lastLink !== 'mocked'`. But I'm not sure this change is safe enough, so I tried a new way.

src/rules/unbound-method.ts
```diff
@@ -73,9 +74,58 @@ export default createRule<Options, MessageIds>({
...
+      if (node.parent?.type !== AST_NODE_TYPES.CallExpression) {
+        return false;
+      }
+
+      const parentCall = node.parent;
+
+      // Check if the call is jest.mocked() by examining the callee
+      if (
+        parentCall.callee.type === AST_NODE_TYPES.MemberExpression &&
+        isSupportedAccessor(parentCall.callee.object) &&
```
G-Rath: `isSupportedAccessor` takes an optional `value` as its second param for just these situations - so rather than having to do the work twice, you can just do `isSupportedAccessor(..., 'jest')`

that in turn means you should be able to refactor this all to be a `return`

src/rules/unbound-method.ts
```diff
@@ -73,9 +74,58 @@ export default createRule<Options, MessageIds>({
...
+    /**
+     * Checks if a MemberExpression is an argument to a `jest.mocked()` call.
+     * This handles cases like `jest.mocked(service.method)` where `service.method`
+     * should not be flagged as an unbound method.
+     */
+    const isArgumentToJestMocked = (
+      node: TSESTree.MemberExpression,
+    ): boolean => {
+      // Check if the immediate parent is a CallExpression
+      if (node.parent?.type !== AST_NODE_TYPES.CallExpression) {
```
G-Rath: this is the opposite of the check we do later on, so ideally we can merge the two but that might be messy due to the optional chaining and require an `else` which I try to avoid so not too fussed to keep it like this for now

G-Rath (changes requested):
I still would expect we can identify the call with `parseJestFn`, but what you've got it a lot simpler than what you had before and finding the right node to parse can be a pain at times, so I'm happy to go with this once you've simplified it a bit more with the `isSupportedAccessor` change

eryue0220:
As mentioned before, the `parseJestFn` failed due to this [condition check](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/utils/parseJestFnCall.ts#L341], `jest.mocked(service.method).mock.calls[0][0]` was parsed as `MemberExpression`, then it failed.

src/rules/unbound-method.ts
eryue0220: Updated.

G-Rath (approved):

PR #1863, opened 2025-11-30
feat: create new `no-unneeded-async-expect-function` rule

### LINKED ISSUE #1722: Proposal: warn on unnecessary async wrapper in expect(...).rejects or expect(...).resolves
### Problem

It is common to write assertions for rejected promises using `.rejects`. However, sometimes developers write it in a more verbose and unnecessary way like this:

```ts
await expect(async () => {
  await somePromise();
}).rejects.toThrow('Error');
```

This can be simplified to:

```ts
await expect(somePromise()).rejects.toThrow('Error');
```

The async function wrapper adds no value and makes the test harder to read, especially when the only purpose of the function is to wrap a single `await` call.

### Proposal

I'd like to propose a lint rule that detects and warns on unnecessary async function wrappers passed to expect(...).rejects or expect(...).resolves.

There are two possible implementation paths:

#### Option 1: Extend `prefer-expect-resolves`

This rule already encourages usage of `.resolves` / `.rejects` over `try/catch` or other patterns.  
It may be natural to include this redundant async wrapper pattern as part of that rule’s responsibility.

#### Option 2: Introduce a new rule

Create a new rule such as `no-unnecessary-async-wrapper-in-expect` that focuses specifically on this anti-pattern.  
It would help keep test code concise and idiomatic.

### Example Code

**Invalid:**

```ts
await expect(async () => {
  await doSomethingAsync();
}).rejects.toThrow();
```

**Valid:**

```ts
await expect(doSomethingAsync()).rejects.toThrow();
```

### Discussion

I'm unsure whether this case is better suited as an extension of prefer-expect-resolves, or whether it's better introduced as a standalone rule. I’d love to hear your thoughts on that — or if there's another preferable approach, I'd really appreciate your guidance.

Thanks for your time!

hainenber:
Closes #1722 

I've implemented said rule to lint for unnecessary `async` wrapper for single `await` call inspired (or blatantly copied? :D) from @G-Rath's ESQuery selector in linked issue.

This is also put into `recommended` option as suggested. PTAL when you have time, folks!


G-Rath (changes requested):
We can't change our configs without doing a new major, so please remove this from recommended 

(On mobile so can't currently do a full review, but figured I'd this in in case you get to it before me 🙂)

hainenber:
I've done removing this rule from `recommended` config. Aside from a crashed Jest worker in MacOS test setup, this is ready for review :D

src/rules/__tests__/no-async-wrapper-for-expected-promise.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+ruleTester.run('no-async-wrapper-for-expected-promise', rule, {
+  valid: [
+    'expect.hasAssertions()',
+    dedent`
+        it('pass', async () => {
+            expect();
+        })
+        `,
+    dedent`
```
G-Rath: could you clean up the indentation of the code samples? the code should be indented one level from the indent level of the `dedent`, and the end backtick should be at the same indent as the `dedent`

i.e.


```suggestion
    dedent`
      it('pass', async () => {
        expect();
      })
    `,
    dedent`
```

src/rules/no-async-wrapper-for-expected-promise.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        if (parent?.type !== AST_NODE_TYPES.CallExpression) {
+          return;
+        }
+
+        const [awaitNode] = parent.arguments;
+
+        if (
+          (awaitNode?.type !== AST_NODE_TYPES.ArrowFunctionExpression &&
+            awaitNode?.type !== AST_NODE_TYPES.FunctionExpression) ||
```
G-Rath: we've got a utility for this

```suggestion
          !awaitNode ||
          !isFunction(awaitNode) ||
```

G-Rath (changes requested):

src/rules/no-async-wrapper-for-expected-promise.ts
hainenber: TIL. I'll amend in next commit

hainenber:
All amended in newest commit :D

src/rules/__tests__/no-async-wrapper-for-expected-promise.test.ts
```diff
@@ -0,0 +1,149 @@
...
+          const a = 1;
+          await doSomethingSync(a);
+        }).rejects.toThrow();
+      })
+    `,
+  ],
+  invalid: [
+    {
+      code: dedent`
+        it('should be fix', async () => {
```
G-Rath: nit: these would be more correct as "should be fixed" 😅

hainenber:
I'm cool with that proposed name actually. Sounds more comprehensible!

src/rules/__tests__/no-async-wrapper-for-expected-promise.test.ts
hainenber: Resolved in commit https://github.com/jest-community/eslint-plugin-jest/pull/1863/commits/6c023de8367642384b4d6992f1a7adedc4b5c5bc :D

src/rules/__tests__/no-unneeded-async-expect-function.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    dedent`
+      it('pass', async () => {
+        await expect(async () => {
+          const a = 1;
+          await doSomethingSync(a);
+        }).rejects.toThrow();
+      })
+    `,
+  ],
+  invalid: [
```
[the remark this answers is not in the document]
hainenber: This is done in HEAD commit

G-Rath (approved):
🥳

PR #1901, opened 2025-12-28
chore: use `eslint` v9 by default, and update `eslint-remote-tester` too

G-Rath:
We might as well do this since we're close to v10 coming out and flat config is the future


G-Rath:
Note that with this we're now technically violating peer dependency requirements relating to `jest-runner-eslint` as while it does support Jest v30 and ESLint v9, [it uses a version of `create-jest-runner` that does not](https://github.com/jest-community/jest-runner-eslint/issues/294#issuecomment-3695179194)

PR #1946, opened 2026-03-22
fix(valid-mock-module-path): don't report virtual mocks

### LINKED ISSUE #1945: [valid-mock-module-path] false positive when using virtual mocks
Jest supports [virtual mocks](https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options), or mocking modules/paths that do not exist anywhere in the system.

```ts
jest.mock(
  '../moduleName',
  () => {
    /*
     * Custom implementation of a module that doesn't exist in JS,
     * like a generated module or a native module in react-native.
     */
  },
  { virtual: true },
);
```

However, this rule falsely identifies these mocks as invalid.

Xunnamius:
Fixes https://github.com/jest-community/eslint-plugin-jest/issues/1945

### DIFF
--- a/docs/rules/valid-mock-module-path.md
+++ b/docs/rules/valid-mock-module-path.md
@@ -24,6 +24,9 @@ jest.mock('../../this/module/does/not/exist');
 
 // Local file that cannot be found
 jest.mock('../../this/path/does/not/exist.js');
+
+// Local file that cannot be found and is NOT virtual
+jest.mock('../../this/path/does/not/exist.js', undefined, { virtual: false });
 ```
 
 The following patterns are **not** considered errors:
@@ -38,6 +41,13 @@ jest.mock('../../this/module/really/does/exist');
 
 // Local file that cannot be found
 jest.mock('../../this/path/really/does/exist.js');
+
+// Module(s) that cannot be found but are configured as virtual
+jest.mock('@org/some-module-not-in-package-json', undefined, { virtual: true });
+jest.mock('some-module-not-in-package-json', undefined, { virtual: true });
+
+// Local file that cannot be found but is configured as virtual
+jest.mock('../../this/path/does/not/exist.js', undefined, { virtual: true });
 ```
 
 ## Options
--- a/src/rules/__tests__/valid-mock-module-path.test.ts
+++ b/src/rules/__tests__/valid-mock-module-path.test.ts
@@ -10,7 +10,7 @@ import {
 const ruleTester = new RuleTester({
   parser: espreeParser,
   parserOptions: {
-    ecmaVersion: 2015,
+    ecmaVersion: 2018,
   },
 });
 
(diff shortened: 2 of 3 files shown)


G-Rath (changes requested):
you're missing a bunch of cases - what if...

* an array or function is passed in as the param
* a variable is passed as the param
* an object is passed, without any properties
* an object is passed, which has a spread
* an object is passed, with a property that is not `virtual`
* an object is passed, with a computed property
  * ...whose key comes from a variable
  * ...whose key is a string literal...
    * ...which is `virtual`
    * ...which is not `virtual`
  * ...whose key is a template string literal...
    * ...which is `virtual`
    * ...which is not `virtual`
    * ...which has interpolation
* an object is passed, with the `virtual` property but whose value...
  * ...is a variable
  * ...is a variable, and is shorthanded
  * ...is an array or function
  * ...is a number

Within these cases, you should find your missing coverage 🙂

Xunnamius:
> you're missing a bunch of cases - what if...

I assume this is a response to my "a test case that can trigger the else path" question? If so, I'll see if one of these triggers the else path. Many of these I've already considered (computed property, passing function, object spread, etc) 🙂

G-Rath:
Yes and no - these are all tests that you should have, because most of them have different AST shapes, and I know they will cover your `else` branches (specifically, spread is the test you're missing for coverage as `properties` is type `ObjectLiteralElement` which is an alias for `Property | SpreadElement`, but we still want the majority of the situations I've mentioned explicitly tested)

Xunnamius:
Spread was literally the first thing I tried, haha. However, I'm getting a syntax error from your test pipeline when I include it. Is there something specific I should add there to enable that syntax?

As for the other test cases, looking at the existing tests for this rule, I tried to match their level of exhaustiveness. Unfortunately I'm approaching the end of my timebox for this, so I can't go much further beyond that this week

Xunnamius:
> However, I'm getting a syntax error from your test pipeline when I include it.

Looks like this:

```
valid-mock-module-path › invalid › jest.mock('../module/does/not/exist', undefined, { ...{ v: true } })

    assert(received)

    Expected value to be equal to:
      true
    Received:
      false
    
    Message:
      A fatal parsing error occurred: Parsing error: Unexpected token ...

      at runRuleForItem (node_modules/eslint/lib/rule-tester/rule-tester.js:894:4)
```

(I'm coming from babel land, where it's usually an incantation/config property/package to get syntaxes working)

G-Rath:
> I'm getting a syntax error from your test pipeline when I include it.

You want to adjust the configuration being passed to `RuleTester`, specifically the `ecmaVersion` - I believe you want 2015:

https://github.com/jest-community/eslint-plugin-jest/blob/d50f50c81af62caa4c73888939e20f4b874fdbe9/src/rules/__tests__/prefer-equality-matcher.test.ts#L5-L10

G-Rath (approved):

PR #1976, opened 2026-06-23
fix(no-export): don't report on assignment to locals named `module`

eryue0220:
Fixes false positives in `jest/no-export` when detecting CommonJS export assignments in test files.

1. Require the exact `exports` property name instead of matching `export` typos via `/^exports?$/`.
2. Respect variable shadowing: ignore assignments when `module` or `exports` refers to a local binding.

Follow-up PRs will cover missed detection patterns (`module["exports"]`, nested chains) and treating `describe` blocks as test files.


G-Rath (changes requested):
I'm heading to bed right now but can review tomorrow.

Could you pull some of these changes out into dedicated PRs so they can have their own changelog entries? That'll also hopefully help with getting coverage sorted

eryue0220:
> I'm heading to bed right now but can review tomorrow.
> 
> Could you pull some of these changes out into dedicated PRs so they can have their own changelog entries? That'll also hopefully help with getting coverage sorted

Hi @G-Rath,

I've refactored the pr into three, another two are #1977 and #1978.

And now you can review them when you're available.

G-Rath (changes requested):
if I've understood right, this is mainly about ensuring we're checking the global `module` and `exports`? if so I'd be keen to land that specifically next as it's a clear bug

eryue0220:
> if I've understood right, this is mainly about ensuring we're checking the global `module` and `exports`? if so I'd be keen to land that specifically next as it's a clear bug

Yes, that's right. This PR is now scoped to the global CommonJS `module` / `exports` handling: avoiding false positives for shadowed bindings and requiring the actual `exports` accessor.

G-Rath:
cool in that case can you not drop support for `module.export`, and remove the support for `module[exports] = function() {};`? as they're unrelated to checking the scope of `module`

eryue0220:
> cool in that case can you not drop support for `module.export`, and remove the support for `module[exports] = function() {};`? as they're unrelated to checking the scope of `module`

Okay, I've updated.

src/rules/__tests__/no-export.test.ts
```diff
@@ -19,6 +19,24 @@ ruleTester.run('no-export', rule, {
     'export default function () {}',
     'module.exports = function(){}',
     'module.exports.myThing = "valid";',
+    dedent`
```
G-Rath, line 22: this is the main test I expect to see in your diff for "don't report on assignment to locals named `module`" - so long as the other tests and the related code required to support them is present, you have changes to make

G-Rath (changes requested):

src/rules/__tests__/no-export.test.ts
eryue0220: > this is the main test I expect to see in your diff for "don't report on assignment to locals named `module`" - so long as the other tests and the related code required to support them is present, you have changes to make

Updated.

G-Rath (approved):

PR #1977, opened 2026-06-24
fix(no-export): detect more CommonJS export assignment patterns

eryue0220:
Based on #1976, fixes false negatives in jest/no-export for CommonJS export assignments in test files.

* Previously, the rule only inspected a single member level (AssignmentExpression > MemberExpression), so several valid export patterns were missed:
* module["exports"] = … — computed property access with a string literal
Nested assignments such as module.exports.foo.bar = … — only the outermost member was checked


src/rules/no-export.ts
```diff
@@ -1,5 +1,93 @@
...
+} from '@typescript-eslint/utils';
+import {
+  createRule,
+  isStringNode,
+  isSupportedAccessor,
+  isTypeOfJestFnCall,
+  resolveScope,
+} from './utils';
+
+const isGlobalModuleIdentifier = (
```
G-Rath, line 8: I think it would make more sense to have a single `isGlobalIdentifier` function that you can pass the identifier name into rather than two slightly different functions

G-Rath (changes requested):

src/rules/no-export.ts
eryue0220: Updated.

src/rules/__tests__/no-export.test.ts
```diff
@@ -79,8 +106,27 @@ ruleTester.run('no-export', rule, {
       errors: [{ endColumn: 15, column: 1, messageId: 'unexpectedExport' }],
     },
     {
-      code: 'module.export.invalid = function() {}; ;  test("a test", () => { expect(1).toBe(1);});',
-      errors: [{ endColumn: 22, column: 1, messageId: 'unexpectedExport' }],
+      code: 'module[exports] = function() {}; test("a test", () => { expect(1).toBe(1);});',
```
G-Rath: we cannot assume that a variable has a specific value, and frankly I don't think there's any valid reason for someone to do this so it's fine to not support it. 

on the other hand, there's really no reason to do any other assignment to `module` so it might be simpler still to just check for an assignment to _anything_ on global `module` - that would simplify our logic to:

1. is this a global?
2. is this either `module` or `exports`?
3. is this on the left hand side of an assignment?

G-Rath (changes requested):

src/rules/__tests__/no-export.test.ts
eryue0220: Done — I've simplified the logic along those lines and dropped the `module[exports]` test case.

The check now boils down to:

1. **Is this a global?** — `isGlobalIdentifier` uses `resolveScope(...) === null`, so locally shadowed `module`/`exports` bindings are ignored.
2. **Is this either `module` or `exports`?** — for a direct assignment we check the LHS identifier; for member expressions we walk to the root identifier and check whether it's global `module` or `exports` (so any assignment on global `module` is caught, not just `module.exports`).
3. **Is this on the left-hand side of an assignment?** — we only handle this in the `AssignmentExpression` visitor, inspecting `node.left`.

No longer special-case computed properties like `module[exports]` — I agree there's no good reason to support that pattern, and the simplified root-identifier check covers the cases we care about without assuming anything about variable values.

G-Rath (changes requested):
can we now pull out adding support for `exports`? that feels like it'll take a huge chunk out of this diff.

also can we not drop support for `export` for now? I know it's technically not valid but it's been there from the start and for now it'll also keep the diff down

eryue0220:
> can we now pull out adding support for `exports`? that feels like it'll take a huge chunk out of this diff.
> 
> also can we not drop support for `export` for now? I know it's technically not valid but it's been there from the start and for now it'll also keep the diff down

Okay updated.

G-Rath:
> Okay updated.

No you have not - I'm wanting a PR that _just_ adds support for `exports` like what we currently have for `modules`, not adding anything further at this point

eryue0220:
> > Okay updated.
> 
> No you have not - I'm wanting a PR that _just_ adds support for `exports` like what we currently have for `modules`, not adding anything further at this point

another pr about `exports`: https://github.com/jest-community/eslint-plugin-jest/pull/1986

PR #1978, opened 2026-06-24
fix(no-export): treat describe blocks as test files

eryue0220:
# Summary
This pr based on #1977 and it includes: 

-  Treat files containing `describe` blocks as test files when deciding whether to report exports
- Aligns rule behavior with the [documentation](docs/rules/no-export.md), which states the rule applies to files containing tests


G-Rath (changes requested):
~I would expect the only thing required to support `describe` would be the change on line 36/124~

right I see you've got #1977 in this too - can you drop all the other commits in this aside from 557d7ea7bed0c42ff777411972e381561cad63a5 ?

eryue0220:
> ~I would expect the only thing required to support `describe` would be the change on line 36/124~
> 
> right I see you've got #1977 in this too - can you drop all the other commits in this aside from [557d7ea](https://github.com/jest-community/eslint-plugin-jest/commit/557d7ea7bed0c42ff777411972e381561cad63a5) ?

Ok, updated.

G-Rath (approved):

PR #1986, opened 2026-07-02
fix(no-export): check for global `exports` too

eryue0220:
# Summary

This pr is based on  #1977. 

Walk the full member chain so nested module.exports.foo.bar assignments are reported, and support module["exports"] string literal access while ignoring computed properties that use local bindings.

### DIFF
--- a/src/rules/__tests__/no-export.test.ts
+++ b/src/rules/__tests__/no-export.test.ts
@@ -14,7 +14,6 @@ ruleTester.run('no-export', rule, {
   valid: [
     'describe("a test", () => { expect(1).toBe(1); })',
     'window.location = "valid"',
-    'module.somethingElse = "foo";',
     'export const myThing = "valid"',
     'export default function () {}',
     'module.exports = function(){}',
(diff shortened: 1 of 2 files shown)


G-Rath (changes requested):
`module.somethingElse = "foo"` does not create an export, so it is not something we should report on (and by extension that applies to anything else you've tried to support of that nature)

fwiw my expectation w/ this PR is that it _just_ adds support for `exports` in the same way we support `module.exports`, without any other changes, even if there are other bugs

eryue0220:
> `module.somethingElse = "foo"` does not create an export, so it is not something we should report on (and by extension that applies to anything else you've tried to support of that nature)
> 
> fwiw my expectation w/ this PR is that it _just_ adds support for `exports` in the same way we support `module.exports`, without any other changes, even if there are other bugs

Okay, updated.

G-Rath:
@eryue0220 I appreciate you're wanting to contribute but it is getting tiring having to go back and forward on these - would you be ok if I was to do PRs for the changes with you as a co-author on the commits so you'll still get the contribution credit?

eryue0220:
> @eryue0220 I appreciate you're wanting to contribute but it is getting tiring having to go back and forward on these - would you be ok if I was to do PRs for the changes with you as a co-author on the commits so you'll still get the contribution credit? 

Sure, It's okay if you would like

PR #1992, opened 2026-07-15
fix: allow TypeScript v7 as a peer dependency

G-Rath:
How time flies


PutziSan:
@G-Rath thx!
Can this be merged and released? Would be cool to install TS 7 without hacking some overrides into the package json :D

PR #1996, opened 2026-07-22
feat(valid-expect-with-promise): add `checkThenables` option for custom promises

christopher-buss:
I am using a `Promise` not from the default lib therefore the rule does not fire. This PR removes the requirement for a Promise to be from the default library and instead falls back to checking if it's any thenable type.

### DIFF
--- a/docs/rules/valid-expect-with-promise.md
+++ b/docs/rules/valid-expect-with-promise.md
@@ -41,3 +41,65 @@ expect('hello world').toBe('hello sunshine');
 
 expect(new Promise(r => r(0))).rejects.toThrow('oh noes!');
 ```
+
+## Options
+
+```json
+{
+  "jest/valid-expect-with-promise": [
+    "error",
+    {
+      "checkThenables": false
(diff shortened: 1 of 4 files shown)


G-Rath (changes requested):
I think this should be an option like with [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises/#checkthenables) (which is the rule I used as a reference when implementing this), since not everything with a `then` is a promise.

For example, Cypress has a [`then` function](https://docs.cypress.io/api/commands/then) which is not a promise

G-Rath:
I would also suggest checking out [their implementation](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-floating-promises.ts#L459-L477), as I assume it's a good default in terms of what to check for

christopher-buss:
don't think the errors here are related to me from what I can tell

G-Rath (approved):
thanks!

PR #2022, opened 2026-09-21
chore: use explicit false for Codecov comments

noritaka1166:
## Summary

- Replace `comment: off` with `comment: false` in `.codecov.yml`
- Use Codecov’s documented boolean value to explicitly disable PR comments

## Reference

- [Codecov: Pull Request Comments — Disable comment](https://docs.codecov.com/docs/pull-request-comments)


G-Rath:
I don't see anywhere in the docs that say "off" is incorrect or not supported?

PR #2024, opened 2026-09-21
chore: migrate Babel config to TypeScript

noritaka1166:
## Summary

- Migrate the Babel configuration from `babel.config.js` to `babel.config.cts`
- Use `.cts` to retain Node.js 20 support while enabling TypeScript configuration
- Update ESLint and TypeScript configuration references
- Remove the obsolete TODO now that Babel supports TypeScript config files ([babel/babel#8529](https://github.com/babel/babel/issues/8529))


G-Rath (changes requested):
the point of doing this would be to use typescript rather than jsdoc for typing

noritaka1166:
> the point of doing this would be to use typescript rather than jsdoc for typing

@G-Rath 
Updated to use a TypeScript type annotation instead of JSDoc.

